PR #394, opened 2015-05-11
Cmake build 

eugenwintersberger:
Hi folks,
  from what I see now the code builds fine under Linux. Before I continue and do the fine tuning I would like to ask you to check out the cmake_build branch and try to run the build on your systems. From what I see the build installs everything required except for the validation tool. 

Concerning the validation tool: as it is broken I would like to remove it from the code repository. Are there any objections with this. 

One final note: I tried a first build on Windows during the weekend and despite some errors it seems to work already surprisingly well (far from production state but it produces for instance the HTML documentation).


prjemian:
Linux cmake build: works for me on linux mint 17.1
delete the current validation tool in the code branch: I agree
Windows cmake build: fails for me due to lack of tool chain

Installed a cmake binary from http://www.cmake.org/download/
In the past, nothing else was needed.
What else is necessary in the toolchain now?
Why is a C compiler needed?

Here's the cmd.exe session:
Microsoft Windows [Version 6.3.9600]
(c) 2013 Microsoft Corporation. All rights reserved.

C:\Users\Pete\Documents\eclipse\NeXus\definitions>cd build
cd build

C:\Users\Pete\Documents\eclipse\NeXus\definitions\build>cmake ..
cmake ..
CMake Warning at CMakeLists.txt:39 (project):
   To use the NMake generator, cmake must be run from a shell that can 
use the
   compiler cl from the command line.  This environment does not contain
   INCLUDE, LIB, or LIBPATH, and these must be set for the cl compiler to
   work.


-- The C compiler identification is unknown
-- The CXX compiler identification is unknown
CMake Error at CMakeLists.txt:39 (project):
   The CMAKE_C_COMPILER:

     cl

   is not a full path and was not found in the PATH.

   Tell CMake where to find the compiler by setting either the environment
   variable "CC" or the CMake cache entry CMAKE_C_COMPILER to the full 
path to
   the compiler, or to the compiler name if it is in the PATH.


CMake Error at CMakeLists.txt:39 (project):
   The CMAKE_CXX_COMPILER:

     cl

   is not a full path and was not found in the PATH.

   Tell CMake where to find the compiler by setting either the environment
   variable "CXX" or the CMake cache entry CMAKE_CXX_COMPILER to the 
full path
   to the compiler, or to the compiler name if it is in the PATH.


-- Configuring incomplete, errors occurred!
See also 
"C:/Users/Pete/Documents/eclipse/NeXus/definitions/build/CMakeFiles/CMakeOutput.log".
See also 
"C:/Users/Pete/Documents/eclipse/NeXus/definitions/build/CMakeFiles/CMakeError.log".

C:\Users\Pete\Documents\eclipse\NeXus\definitions\build>

eugenwintersberger:
For Windows you currently need xsltproc. There is a very nice binary available which does the job. The C Compiler is a strange thing - I have not figured out yet why it is required. Anyhow, we should first focus on the Linux build. 
There are two things which concern me:
-> am I installing everything required
-> are the schemas working ;)

eugenwintersberger:
I forgot to mention: I found an xsltproc binary in the source tree of the documentation. However, I would like to remove this too as I think it is a bad idea to ship binaries within a source tree. 
My idea is rather to provide an MSI package for xsltproc rather than shipping it with the source distribution. I am currently learning how to do MSI packages and this would be a good example for a beginner ;)

prjemian:
AFAIK: <definitions>/utils/win32/xsltproc.exe was included (with LICENSE 
file) to *ensure* successful builds on Windows.  You want to break that?

prjemian:
Regarding just the Linux build:

On 5/11/2015 3:12 PM, Eugen Wintersberger wrote:
> There are two things which concern me:
> -> am I installing everything required
> -> are the schemas working ;)

The requirements for installation are not clearly stated.  Here's what I 
have long thought the definitions repository provides:
* NXDL files in several directories
* the manual
* nxdl.xsd and nxdlTypes.xsd: the rules for NXDL files

That's it.

The files BASE.xsd and NeXus.xsd are used by the validation process that 
has been questioned.  They are not needed for anything else.

The build process creates some .xsd and .sch files.  These are the 
schema files Eugen is asking about.  They, too, are only used for the 
validation process that has been questioned.

-rw-r--r-- 1 user_account   213424 May 11 13:11 all_base.nxdl.xml
-rw-r--r-- 1 user_account   307752 May 11 13:11 all_base.sch
-rw-r--r-- 1 user_account   447332 May 11 13:11 all.nxdl.xml
-rw-r--r-- 1 user_account   803865 May 11 13:11 all.sch
-rw-r--r-- 1 user_account  1037805 May 11 13:11 all_unsorted.xsd
-rw-r--r-- 1 user_account   347788 May 11 13:11 all.xml

There are no rules governing the content in the .xml files.  They 
contain various sets of the NXDL files, some with the comment lines 
replaced with blank lines.  The .sch files are schematron schema files 
and are used for validation.   For what it is worth, all these files are 
"well-formed XML" which means they do not violate the XML syntax rules.

Are any of these files used by anyone else?
Perhaps nexpy?

I suggest that, if we are to remove the existing validation code, then 
we should also remove BASE.xsd, NeXus.xsd, and stop building the items 
in the build directory "schema".

eugenwintersberger:
Concerning the xsltproc.exe: to build the definitions the user already has to install a whole bunch of software: CMake, Python, Sphinx, maybe LaTeX, etc. Thus I believe providing a link to the tarball with xsltproc.exe makes it easy enough to install the program. 

I totally agree with your installation targets. This means that I can even omit the schema files in the ALL target which makes things far more easier. I will put this on the agenda for the next telecon. I am on holiday next week and thus would have time to finish the package right after the telecon (if we find an agreement on what to install ;)).

eugenwintersberger:
I am so far done with the cmake build of the Nexus definitions. The pull request will remain until Wednesday May 27th, 2015 and will be merged into master on Thursday May 28th, 2015.

PR #408, opened 2015-05-29
Issue 341

prjemian:
now sidebar appears and right margin is constant throughout, problem was in _static/blockquote.css with an unnecessary import


prjemian:
* confirmed successful build on jenkins 
* confirmed successful access through web browsers on windows, linux, and android
* confirmed that sidebar is visible on all test systems
* confirmed that right margin on NXcanSAS page remains constant (no indentation hourglass effect)

Issue #399, opened 2015-05-27
NXtransformations' field default name is not useful for automatic code generation


PeterC-DLS:
It is currently defined as "anonymous__NEEDS_XSD_CHANGE__", perhaps "transformation" is better.

zjttoefs:
The idea is that this group has a suggested, but not required name. The current XSD cannot handle that to my knowledge. Hence this "marker" for now. 
Good to have a ticket to keep an eye on it though.

prjemian:
There may be other NXDL files in which the notion of an "anonymous" field (anonymous: name to be defined when the file is written) is used.  

This needs a general solution.

PeterC-DLS:
See possible solution in pull request #411

zjttoefs:
Since that has been merged, I guess we're done.

PR #411, opened 2015-06-10
Add nameType tag attribute

PeterC-DLS:
This allows generating code to handle multiple arbitrarily named fields, and also, handle associated fields where their names follow as simple substitution pattern. It is a possible solution for #399


zjttoefs:
@PeterC-DLS any problems with adding the default="specified"? Otherwise I'd add that and accept the pull request.

prjemian:
agreed to accept at today's teleconference

Issue #418, opened 2015-08-27
clarify that the application definitions are allowed to define new fields


zjttoefs:
This is required for a number of reasons but the documentation is a misleading:

http://download.nexusformat.org/doc/html/classes/base_classes/index.html

prjemian:
Is this the part which is misleading?

NeXus base class definitions define the complete set of terms that might be used in an instance of that class.

Are there other parts of the documentation that mislead about fields declared in base classes vs. fields in application definitions?

zjttoefs:
I've edited that place and one other instance in that pull request referenced here. 
No other wording I came across in a quick browse was as strong.

PR #419, opened 2015-08-27
address most obvious cases claiming base class completeness

zjttoefs:
The main sections actually concerning base classes and application definitions
were not worded as strongly. So we are toning it down here.

Refs #418


prjemian:
fixes #418

Issue #437, opened 2016-01-22
update verysimple example to use NIAC2014 signal/axes


prjemian:
This example uses NeXpy to write a very simple NeXus HDF5 Data file.
It needs to be updated to use current NeXpy code
and checked to see that it uses the signal and axes *group* attributes.

prjemian:
follows completion of #434

zjttoefs:
Since the _indices are missing, I'll reopen that for the time being.

Issue #436, opened 2016-01-22
suggest updates of NXDL classes for NXdata group signal & axes attributes


prjemian:
These classes use the signal and/or axes attribute on the data field (and should be urged to change to the NIAC2014 way):

* NXdetector
* NXguide
* NXlauetof
* NXmonitor
* NXxbase

refs #434

prjemian:
opened branch _issue_436_ for this work

https://github.com/nexusformat/definitions/tree/issue_436

prjemian:
NXlauetof and NXxbase are part of #447 (PR #448)

prjemian:
mark axis and primary attributes as deprecated rather than delete them

prjemian:
Let's review this at next telco _before_ making a PR, those TODO items in NXdetector should be resolved.

prjemian:
part of #443 and #447

prjemian:
merged in #454

Issue #435, opened 2016-01-22
need to be able to mark attributes as deprecated


prjemian:
fields and groups can be so marked
needs to extend to attributes as well

prjemian:
will include with issue_434 branch

PR #439, opened 2016-01-23
Issue 434

prjemian:


prjemian:
@zjttoefs @rayosborn Can you look this over to see if it resolves all your concerns (except issues #436 & #437)?  If it is acceptable (make comment here), then I will merge it.

The only controversial item is that I marked the old signal/axes/axis attributes as deprecated (#435).  Might be better to invent yet another new attribute called "discouraged".  What do you think?

zjttoefs:
I would have preferred a pull request that doesn't combine the build system change and editorial edits.

From what I can see briefly some of the examples are missing the axis_indices attributes. Those should be added. And the "axes" group attribute should be a string array, even in the simple 1D case. 

Otherwise I think there is no need to distinguish between "deprecated" and "discouraged". There are too many non-native speakers involved that don't get these nuances. Reluctant software people will interpret either as "still works, though".

I'm not sure about the apologetic tone that datarules.rst has on not using HDF5 dimension scales. They are not as powerful as our axes group attributes. At least last time I checked. It's nothing you introduced with these changes I believe but might we worth addressing while we're making edits in that area.

prjemian:
No changes to the build system with this PR.  Those changes were 
complete in a previous PR.

Rather, one of the changes in this PR was to remove files marked with 
".obsolete" suffix.  That commit could be removed if it is out of place.

rayosborn:
The nexusformat API cannot read the new axis_indices and they are not included in 99.99% of NeXus files created so far, so we had better proceed cautiously with the new system or it will just cause more confusion. I still think mandating the new system is a recipe for delaying NeXus adoption.

prjemian:
The wording says the new method of signal/axes is "recommended" while 
the previous are marked "deprecated" with these changes.  This is the 
caution we are using, rather than mandating these changes.

prjemian:
So, is this OK to merge?

prjemian:
Sensing no objections, I'll merge tomorrow (~ 24h from now)

zjttoefs:
Sorry, I was planning to merge that Tuesday but got interrupted. 
Merged now. It does not address all open problems in a final way but takes us in the right direction and there are tickets for the remaining items.

zjttoefs:
Merged. (I thought Github might sense this.)

PR #460, opened 2016-03-11
Copyright dates updated by new Python script

prjemian:


prjemian:
That last action should be added to the update code.  It won't fit the default pattern.  OR, we could just code the last year into the conf.py directly.  That's easier by far.

Issue #469, opened 2016-05-13
consider one-line summaries for application definitions and contributed ones


zjttoefs:
After the success with #270

prjemian:
These commits were made to the master branch and need to be moved to a new branch to become part of a pull request before being merged into the master branch.  Need to revert the master branch history for it this action.

Others who fetched since 364d958 should do a ``git fetch`` before continuing work on the master.

Following this help (http://stackoverflow.com/questions/18082830/reverting-a-series-of-pushed-merges-and-commits-in-git-without-rewriting-histor), used these commands:

	git clone https://github.com/nexusformat/definitions
	cd definitions/
	git branch issue_469
	git checkout issue_469
	git branch -f master 5df54d1
	git push -f origin master

zjttoefs:
I see no ill effects and it's much nicer!

PR #472, opened 2016-05-24
NXcontainer Contributed Definition

mtwharmby:
This PR adds a new NXcontainer contributed class definition. This is to be used in work being done on the XPDF beamline at Diamond Light Source and its structure has been discussed extensively with @markbasham. Also included are two diagrams used as part of the documentation to explain the intent of the class, using a relatively complicated example.


contributed_definitions/NXcontainer.nxdl.xml
```diff
@@ -0,0 +1,172 @@
...
+	name="NXcontainer"
+	category="contributed"
+	version="1.0"
+	type="group"
+	extends="NXobject"
+	>
+
+	<doc>
+		Template of the state of a container holding the sample under 
+		investigation.
```
prjemian: With #469, the first *single* line is a summary.  This should be one line only.  With recent commit 039c61f, the summaries were further shortened.  This one might be changed to read
"State of a container holding the sample under investigation."

contributed_definitions/NXcontainer.nxdl.xml
```diff
@@ -0,0 +1,172 @@
...
+			<dim index="1" value="n_comp"/>
+		</dimensions>
+	</field>
+	<field name="relative_molecular_mass" type="NX_FLOAT" units="NX_MASS">
+		<doc>Relative molecular mass of container.</doc>
+		<dimensions rank="1">
+			<dim index="1" value="n_comp"/>
+		</dimensions>
+	</field>
+	<field name="packing_fraction" type="NX_FLOAT" units="NX_UNITLESS">
```
prjemian: This field is a duplicate of the same content above.  Remove.

contributed_definitions/NXcontainer.nxdl.xml
```diff
@@ -0,0 +1,172 @@
...
+	</group>
+	<group name="orientation" type="NXtransformations">
+		<doc>
+			The angle the container makes to the beam and how it may change 
+			during the experiment.In combination with shape this should allow 
+			the beampath through the container to be modelled to allow the 
+			adsorption of the container to be calculated.
+		</doc>
+	</group>
+	<group name="empty_measurement" type="NXentry">
```
prjemian: Use the NXDL <link> element here instead of <group>, as states the <doc>.  Refer to http://download.nexusformat.org/doc/html/nxdl_desc.html#link

mtwharmby: Sounds like the right idea. But I don't know what the target attribute of the link would be. For example, if there are several containers surrounding the sample, I would expect multiple NXentry elements (one for each container) and each wwith a unique name. It's not possible to know those without knowing the experimental set up (which I don't want to prescribe!). Not sure how to resolve this (I have also raised this with @markbasham offline).

prjemian: suggest something like this, replacing the &lt;group> element with:

    <link name="empty_measurement" target="/NXentry"><doc> as you have written </doc></link>

prjemian: The documentation for the &lt;link> element could be improved as it is possible to link a group OR a field element.  
This is the *improved* text:

http://download.nexusformat.org/doc/html/nxdl_desc.html#nxdl-data-type-linktype


In this case, link the entire NXentry group.  Since the HDF5 name of the NXentry group is not known by the NXDL, it can't be specified here.  But, you can specify the NeXus class path (the path to the target field or group as specified by the chain of NX_class attribute values rather than the HDF5 addresses).

prjemian: ALSO, the name "empty_measurement" is but one case of such a reference.  It is conceivable to use a non-empty container as the reference measurement, such as for a solution concentration series, where a zero concentration sample in the container would be used.

Suggest the name: "reference_measurement" with documentation adjusted to match.  Or keep the name but still adjust the docs.

contributed_definitions/NXcontainer.nxdl.xml
mtwharmby: This has been fixed in the file but hasn't been updated in these notes.

contributed_definitions/NXcontainer.nxdl.xml
```diff
@@ -151,12 +151,14 @@
 			adsorption of the container to be calculated.
 		</doc>
 	</group>
-	<group name="empty_measurement" type="NXentry">
+	<group name="reference_measurement" type="NXentry">
```
prjemian: The revised structure here places a NXentry group (via a link) inside a NXentry group.  That needs to be changed.

replace

	<group name="reference_measurement" type="NXentry">
		<link name="reference_measurement" target="/NXentry"></link>
		<doc>
			A link to a full data collection which contains the actual 
			measured data for this container within the experimental set up 
			(with no sample or inner container(s)). This data set will also 
			include the wavelength/energy, measurement time and intensity for 
			which these data are valid.
		</doc>
	</group>

with

	<link name="reference_measurement" target="/NXentry">
		<doc>
			A link to a full data collection which contains the actual 
			measured data for this container within the experimental set up 
			(with no sample or inner container(s)). This data set will also 
			include the wavelength/energy, measurement time and intensity for 
			which these data are valid.
		</doc>
	</link>

mtwharmby: Corrected as requested. I was confused by this bit of the docs:
http://download.nexusformat.org/doc/html/nxdl_desc.html#link
Which states: "A link element can only be a child of a field or group element"

prjemian: And this instance has it as a child of a definition element.  I'll adjust the documentation.  Thanks.

contributed_definitions/NXcontainer.nxdl.xml
```diff
@@ -0,0 +1,170 @@
...
+		   scattered by the sample with scattered beams passing through the 
+		   upstream wall of the capillary, then windows 4 and 5. As part of the
+		   corrections for a PDF experiment it is necessary to subtract the PDF
+		   of the empty container (i.e. each of the windows and the capillary).
+		   To calculate the PDF of the empty container it is necessary to have 
+		   the measured scattering data and to know the nature (e.g. density, 
+		   elemental composition, etc.) of the portion of the container which 
+		   the beam passed through.
+		   
+		   figure:: container/ComplexContainerBeampath.png
```
prjemian: this line does not generate the figure yet
prefix with two dots at the correct indentation, like line 44 above, then the documentation page will be built properly by Sphinx

prjemian:
With the last change, my local copy now builds documentation properly and validates against the nxdl.xsd XML Schema with no errors.
I believe the proposed NXcontainer NXDL is ready to consider accepting as a contributed definition.  Let's discuss this at the next telco.

PR #475, opened 2016-06-04
add NXlog as child group of NXentry

prjemian:
This is a proposition.

some example data files (such as scan101.nxs from NeXpy) have this

Looks like the general intent of NXlog could convince someone to place it as a child of NXentry, such as when the NXlog contains the result of data reduction or analysis

### DIFF
--- a/base_classes/NXentry.nxdl.xml
+++ b/base_classes/NXentry.nxdl.xml
@@ -187,5 +187,6 @@
 	<group type="NXparameters" />
 	<group type="NXprocess" />
 	<group type="NXsubentry" />
+	<group type="NXlog" />
 </definition>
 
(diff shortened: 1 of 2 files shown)


prjemian:
as discussed (http://wiki.nexusformat.org/Telco_20160614), this PR will be withdrawn.

resolved by the general agreement on being permissive with base classes. That is while base classes aim to cover most use cases, they don't list all acceptable groups and fields. In order to be able to progress the standard additional data is acceptable. Validation tools should not treat such additional items not covered by a base class as errors, but notes or warnings. Users should encouraged to find the best defined location for their information, but we understand there is not a defined place for all possible data.

This may require clarification in the manual.

Issue #492, opened 2016-10-14
move NXcanSAS to applications


prjemian:
refs #420: NXcanSAS was ratified at 2016NIAC

prjemian:
@zjttoefs : this **branch** is ready for a PR - should we have canSAS look it over first?

As I look at the 2012 specification document, it shows the SASinstrument as optional.

If canSAS review comes first, we'll need to build the docs separately so they can view the current specification exactly as it exists. Otherwise, the revised docs will be built automatically by the NeXus build server and we can refer there.

prjemian:
structure of current NXDL
[NXcanSAS-xture.txt](https://github.com/nexusformat/definitions/files/540344/NXcanSAS-xture.txt)

PDF of current HTML documentation
[NXcanSAS.pdf](https://github.com/nexusformat/definitions/files/540345/NXcanSAS.pdf)

prjemian:
pull request will wait until the canSAS data formats working group has a chance to review

prjemian:
Review checklist:

* [ ] Accept that SASinstrument group is optional (which seemed to be our intent all along)
* [ ] Accept current scope of NXcanSAS

A discussion arose during the NIAC2016 review based on this question:
Is this to be a strict minimum information necessary for reduced SAS data or an extension of the 1-D format (which included specifications of other items)?
Should NXcanSAS define names for all those terms that describe raw data information?  This specifies the name *if* the value is reported.

* 1-D XML format: http://www.cansas.org/formats/canSAS1d/1.1/doc/
* "Absolute minimum requirement for analysis of SAS data" (as described in first figure here: http://cansas-org.github.io/canSAS2012/framework.html)

We can always revise this NXcanSAS standard and update its version number.

smk78:
Review checklist:
~~~~
Accept that SASinstrument group is optional (which seemed to be our intent all along)
~~~~
YES

~~~~
Accept that we want to define SAStransmission_spectrum (which is optional)
~~~~
YES
Transmissions might be a bit of a 'neutron thing', but they are valuable information that does not normally get captured for posterity anywhere else. They are also a useful sanity check and diagnostic tool, however: for example, there have been many occasions when a look at the transmissions of some samples has enabled me to show a user that the reason there analysis was not as they expected was because they were fitting the wrong dataset to the wrong contrast!

[...]

Issue #397, opened 2015-05-20
clarify the geometry documentation as agreed upon at 2014 NIAC


prjemian:
The specification of geometry and axis coordinates needs to be clarified in the manual, as revisions were agreed upon at the 2014 NIAC meeting.

Use this as the starting point
http://www.bernstein-plus-sons.com/software/CBF/doc/cif_img_1.7.8.html#axis

yayahjb:
See also the attached Dectris document, and the history of the discussion at
https://sites.google.com/site/nexuscbf -- HJB

prjemian:
see #195

zjttoefs:
Will add some information and links here to aid to updating of the manual.

zjttoefs:
The code to convert CBF to NeXus is deeply bundled in CBFlib. You'll find evidence it exists here:

http://sourceforge.net/p/cbflib/code-0/HEAD/tree/trunk/CBFlib/src/cbf_hdf5.c

Some documentation is here:

http://cbflib.sourceforge.net/doc/CBFlib.html#2.7

The short bit we already have in the manual is here: http://download.nexusformat.org/sphinx/design.html#coordinate-transformations

zjttoefs:
This poster is probably the most useful to visually experience the CBF-NeXus mapping:

https://sites.google.com/site/nexuscbf/home/presentation-foils/Integration_Poster_10May14_w_overlays.pdf

zjttoefs:
I don't think there is a concrete problem outstanding. There is always room for improvement.

PR #516, opened 2016-12-01
(issue 492) NXcanSAS v1.0 release

prjemian:
Reviewers: check your box after you have reviewed the [NXcanSAS NXDL code](https://github.com/nexusformat/definitions/blob/NXcansas-492/applications/NXcanSAS.nxdl.xml)

* This is the final review before the v1.0 release.  
* Technical corrections only, please.  
* Keep all comments on *this* issue page (not on the canSAS-DFWG Google group discussion).

* [x] Andrew Jackson (https://github.com/nexusformat/definitions/pull/516#issuecomment-264491064)
* [x] Paul Butler (https://github.com/nexusformat/definitions/pull/516#issuecomment-267596234)
* [x] Steve King (https://github.com/nexusformat/definitions/pull/516#issuecomment-267654821)
* [x] Brian Pauw (https://github.com/nexusformat/definitions/pull/516#issuecomment-264547888)
* [x] Adrian Rennie (https://github.com/nexusformat/definitions/pull/516#issuecomment-267626260)

Copy of HTML documentation: 
[NXcanSAS.pdf](https://github.com/nexusformat/definitions/files/623702/NXcanSAS.pdf)


prjemian:
Sent emails to each reviewer (@ajj, @butlerpd, @smk78, @tokduj, and Adrian Rennie) requesting their review.

toqduj:
Hi. Just a few small items based on the HTML documentation:

In the section on Q uncertainties:
"Generally, this is the estimate of the uncertainty of each Q. Typically the estimated standard deviation." 
should be changed to:
"This is the estimate of the uncertainty of each Q. By default, this will be interpreted to be the estimated standard deviation. In special cases, when a standard deviation cannot possibly be used [note: do such cases exist?], its value can specify another measure of distribution width."
Note that a standard deviation always exists for a population of values, so I do not know if that last clause should be in there or not. 

In the section on resolution:
Defining the mean value µ and the standard deviation s, the span described by µ-s to µ+s contains 68% of the population (the standard deviation is, in effect, a half-width of the distribution). That means that the sentence introducing the distribution types in the resolution, which now states: 
"Square : note that the width of the square would be ~1.4 times the standard deviation specified in the vector",
should state:
"Square : note that the full width of the square would be ~2.9 times the standard deviation specified in the vector"

On the intensity vector I: 
Do we want to stick with typical units in centimetres, or give them proper SI units of metres? That would make the typical units 1/m/sr and m^2. Since both 1/cm and 1/m are commonly applied units, I think we should go for the more correct one. Don't worry, I'm not going to argue (anymore) for using 1/m for Q. That one "should" be 1/nm to stick with common use (ugh). 

I/@uncertainties:
"For Poisson statistics, use 1/√I."
--> 
"For Poisson statistics, can not be less than √I."

On dQw/dQl:
w and l specify directions that are ill-defined in 2D systems. Can we modify this to correspond to x and y?

On Qmean:
Is there a possibility for both Q and Qmean to be specified? If so, perhaps an example in the text?

General remark: are the directions x, y specified and/or in accordance with the NeXus coordinate system definitions?

Other than that, I think it looks very good!

ajj:
Docs for @uncertainties on Q : agree with @toqduj on rewording. I would prefer people to use the natural definition for each distribution e.g. Full Width would be the natural one for "Square" rather than having to convert to/from standard deviation. However, that gets thorny quickly definition wise.

Do we allow similar syntax for @uncertainties - since we might have components and their sources. Tend to think this should always be standard deviation if it is to be interpreted as an "error bar" in the normal usage.

Docs for @resolution: Says "This may be used to describe the slit-length at each datum" after the Qdev example. This is a bit confusing as Qdev is not the slit-length. I suggest removing that sentence.

Otherwise looks OK to me.

ajj:
@toqduj -> typical units in SANS are 1/Å for Q and 1/cm for I(q). At least 1/m would annoy everyone equally ;)

@toqduj -> Q & Qmean -> Yes, always in fact. Shouldn't give Qmean without Q (derived from NIST usage where Qmean is centre for resolution distribution which might be different from nominal Q value of bin).

toqduj:
Regarding @ajj in [here](https://github.com/nexusformat/definitions/pull/516#issuecomment-264435108) 
in case you want to use "full width" instead of standard deviation for a square distribution, you will need to specify *for every distribution shape* what the uncertainties values indicate. This, as you say, gets thorny quickly in the interpreting code. I'd recommend sticking to standard deviations, nicely related to the normal use of the error bar. 

As for Q in units of 1/m, there is something nice about not having to worry about the meaning of your reciprocal length variable to figure out which units it should be in. I now use meters and reciprocal meters throughout the McSAS code, which avoids having to pepper the code with conversion factors (actually, we have a nice "units" library, which converts all input to SI units for internal use). These days, an exponent of 10^9 (or its square) is no problem in storing or programming anymore, at least in modern programming languages. 

As an added bonus, all your intensity-vs-q plots are now in 1/m vs 1/m units, which nicely cancels out :).

smk78:
toqdui said: _'I now use meters and reciprocal meters throughout the McSAS code, which avoids having to pepper the code with conversion factors (actually, we have a nice "units" library, which converts all input to SI units for internal use).'_

I think we need to remember that what we are putting the finishing touches to here is a file format, and not an attempt to dictate how the global SAS community ought/should/could represent their data. To mandate that NXcanSAS only accept data that is 1/m vs 1/m will mean that either every program that writes/reads data will need modifying, or that no one - except maybe toqdui :-) - ever uses NXcanSAS, in which case we are wasting our time. Sorry if this sounds harsh, but we have to be pragmatic. A file format is a data wrapper, it should accept as wide a variety of data (of the specified type) as possible provided that data can be described accordingly.

smk78:
To prjemian: You’ve previously also circulated a structure file. Can you post the latest current example, please? Thanks.

prjemian:
Here is the structure of the current revision:

    NXcanSAS
      entry : NXentry
        @canSAS_class : NX_CHAR = SASentry
        @default : NX_CHAR
        @version : NX_CHAR = 1.0
        definition = NXcanSAS
        run
        title
        collection : NXcollection
          @canSAS_class : NX_CHAR = SASnote
        data : NXdata
          @I_axes : NX_CHAR
          @Mask_indices : NX_CHAR
          @Q_indices : NX_INT
          @canSAS_class : NX_CHAR = SASdata
          @signal : NX_CHAR = I
          I : NX_NUMBER
          Idev : NX_NUMBER
          Q : NX_NUMBER
          Qdev : NX_NUMBER
          Qmean : NX_NUMBER
          ShadowFactor
          dQl : NX_NUMBER
          dQw : NX_NUMBER
        data : NXdata
          @T_axes : NX_CHAR = T
          @canSAS_class : NX_CHAR = SAStransmission_spectrum
          @name : NX_CHAR
          @signal : NX_CHAR = T
          @timestamp : NX_DATE_TIME
          T : NX_NUMBER
          Tdev : NX_NUMBER
          lambda : NX_NUMBER
        instrument : NXinstrument
          @canSAS_class : NX_CHAR = SASinstrument
          collimator : NXcollimator
            @canSAS_class : NX_CHAR = SAScollimation
            distance : NX_NUMBER
            length : NX_NUMBER
            aperture : NXaperture
              shape
              x_gap : NX_NUMBER
              y_gap : NX_NUMBER
          detector : NXdetector
            @canSAS_class : NX_CHAR = SASdetector
            SDD : NX_NUMBER
            beam_center_x : NX_FLOAT
            beam_center_y : NX_FLOAT
            name
            pitch
            roll
            slit_length : NX_NUMBER
            x_pixel_size : NX_FLOAT
            x_position
            y_pixel_size : NX_FLOAT
            y_position
            yaw
          source : NXsource
            @canSAS_class : NX_CHAR = SASsource
            beam_shape
            beam_size_x : NX_NUMBER
            beam_size_y : NX_NUMBER
            incident_wavelength : NX_NUMBER
            incident_wavelength_spread : NX_NUMBER
            radiation = Spallation Neutron Source 
                      | Pulsed Reactor Neutron Source 
                      | Reactor Neutron Source 
                      | Synchrotron X-ray Source 
                      | Pulsed Muon Source 
                      | Rotating Anode X-ray 
                      | Fixed Tube X-ray 
                      | UV Laser 
                      | Free-Electron Laser 
                      | Optical Laser 
                      | Ion Source 
                      | UV Plasma Source 
                      | neutron 
                      | x-ray 
                      | muon 
                      | electron 
                      | ultraviolet 
                      | visible light 
                      | positron 
                      | proton
            wavelength_max : NX_NUMBER
            wavelength_min : NX_NUMBER
        process : NXprocess
          @canSAS_class : NX_CHAR = SASprocess
          date : NX_DATE_TIME
          description
          name
          term
          collection : NXcollection
            @canSAS_class : NX_CHAR = SASprocessnote
          note : NXnote
        sample : NXsample
          @canSAS_class : NX_CHAR = SASsample
          details
          name
          pitch
          roll
          temperature : NX_NUMBER
          thickness : NX_FLOAT
          transmission : NX_NUMBER
          x_position
          y_position
          yaw

toqduj:
@smk78 The units are not fixed in any case (that's what the @units attribute is for). However, there are recommendations for units in the document. If we don't want to recommend any particular form, that's fine too. 

Note, however, that any flexibility in the format specification, requires additional programming effort on the ingestion side to deal with. Therefore, specifying something is not so bad either, to avoid having to write for an infinity number of possible cases.

prjemian:
read this section in the NeXus manual regarding units.
http://download.nexusformat.org/doc/html/datarules.html#nexus-data-units

ajj:
@smk78 there was definitely no intention on my part to impose unit usage. It is true that our examples use the customary units from the neutron world (1/cm and 1/A) rather than those from the x-ray world (1/mm or 1/m and 1/nm), but I think that it is clear in the docs that the writers and readers of the data need to ensure that they are properly handling the units as per NeXus standards.

I was being a bit tongue in cheek in my comments to @toqduj about the 1/m for Q - that is one that is not worth fighting. The nice thing about 1/cm or 1/A give values that are range nicely from around 1 and 3 orders up for I and from 1 and 3 orders down for Q, which makes them more "human scale". There is also the calculation precision question and pre-cancelling your exponents.

We need to ensure that the barrier to entry is sufficiently low - if people have to do unit conversions from their existing data workflows they will just ignore this.

I'm happy as is aside from fixing the resolution slit-length part of the docs (would have ticked box, but that doesn't work!). @prjemian - fix the slit-length comment and then count me as a tick!

toqduj:
Likewise, I'm happy with the unit description in the NeXus docs as linked to by @prjemian. Count me ticked off as well, with the minor changes touched on in my first post.

toqduj:
as for @ajj : I'd agree that for display units, those values make sense. For storage and internal use I would stick to SI, however. Anyway, we now each can do what we want and grumble in our respective beards at one another :).

prjemian:
@toqduj asked:

    On dQw/dQl:
    w and l specify directions that are ill-defined in 2D systems. Can we modify this to correspond to x and y?

These terms are the standard terms in the literature for slit-smeared systems (where dQl >> dQw).  Switching to x & y would be inconsistent with that decades-old body of work.  There are instruments for which dQl is aligned with X and others in which it is aligned with Y.  The most common difference is between a lab source and a synchrotron source.

For the 2D systems in question, does that refer to slit-smeared collimation and a 2D detector or anisotropic scattering from the sample?  For the latter, this is a factor in the smearing model.  For the former, are there any instruments with 2D detectors for which dQl >> dQw?  Arguably, dQw and dQl are not used in such cases.  Rather, use a 2D Qdev array that defines the resolution at each pixel.

prjemian:
@toqduj asked:

    General remark: are the directions x, y specified and/or in accordance with the NeXus coordinate system definitions?

Yes.  See the NeXus documentation: http://download.nexusformat.org/doc/html/design.html#nexus-coordinate-systems

prjemian:
Here's the documentation revisions from  3636857:
[NXcanSAS.pdf](https://github.com/nexusformat/definitions/files/629898/NXcanSAS.pdf)

No structure revisions.

This ticks off @ajj and @toqduj,  What about @smk78, @adrianrennie, and @butlerpd?

adrianrennie:
I have a comment about I/@uncertainties (re @toqduj) who suggested:  "For Poisson statistics, use 1/√I." --> "For Poisson statistics, can not be less than √I.".  For treated data these types of expression are not generally correct as that simple calculation does not allow for the scaling of counts to intensity units in reciprocal length.  I think that a simple square root works only when the data are described in particle counts.  This definition should probably avoid specification of the procedure to estimate this uncertainty.

prjemian:
The 1/√I comment derived from a suggestion about the overall process, not as specific directions to be followed.  I will remove this based on your advice (@adrianrennie), since the comment creates an unnecessary distraction.  It is better to write less, than more.

smk78:
Just reviewed the pdf doc & structure file. Generally good. But I have a number of comments/questions:

PDF Documentation

 1) Suggest including the NeXus coordinate system diagram to facilitate interpretation of x,y,z
 2) Regarding the discussion above about uncertainty vs uncertainties, I note we use axes and indicies already, so uncertainties seems reasonable
 3) Under NXdata : I: three forms for intensity are elaborated. The way the typical units for the first two are represented are inconsistent (either you use the caret symbol or not I say). And doesn't the second example need /atom (or atom^-1) added to it? Is using barns/sr/atom (or barns sr^-1 atom^-1), as the amorphous/total scatterers software (eg, GUDRUN, etc) tend to do, allowed? Finally, the third example could just be counts, right?
 4) Under NXdetector : beam_center_x/y : can we state that these can be negative, please? They are on SANS2D...
 5) Can we expand on the meanings of NXcollimator : aperture and NXsource : beam_shape, please, to clarify how they are different?
 6) Under NXsource : incident_wavelength_spread: how would this be used to describe the Maxwellian from a TOF source where I don't think a FWHM would be appropriate? Also see 'Structure File' 2) below.
 7) Under NXsample : transmission: I suggest a sentence to note that the ability to store a transmission spectrum, instead of a single value, is provided elsewhere in the structure?
 8) Under NXsample : x/y_position: I was a bit confused by this; if you were actually collecting data on a sample, wouldn't the bit you are collecting data on actually be (x,y,z) = (0,0,z)? Or, perhaps if scanning/rastering was what was in mind, don't you need to specify your coordinate reference point on the sample?
 9) Where should one put the name of the data processing/analysis program: Under NXprocess : name, or under NXprocess : description?

Structure File

 1) There are several items without a NX_type (string, integer, etc)
 2) SASentry has 'run' and 'title', but in the pdf it has 'run', 'title' and 'name'
 3) Why does SAStransmission_spectrum have @timestamp, but SASdata does not?
 4) What about including NXsource : wavelength_peak? Also see 'PDF Doc' 6) above. 
 5) Why are @units not in the structure file?
 6) Why are @uncertainties not in the structure file?

prjemian:
@smk78 wrote:

    Why are ``@units`` and ``@uncertainties`` not listed in the structure file?

Seems that the software generating this view only shows attributes on groups and not fields (could be changed, shows a lot more and tends to be distracting from a structural view).  The NX_class group attribute is also removed from specific listing, as it is presented on the line where the group is declared.

prjemian:
The default type is NX_CHAR (a string) unless otherwise stated.

prjemian:
@smk78 wrote:

    SASentry has 'run' and 'title', but in the pdf it has 'run', 'title' and 'name'

In the PDF, ``title`` and ``run`` are fields.  ``name`` is an attribute of the ``run`` field.  The PDF is consistent with the NXDL file.

prjemian:
About the *typical units* for different ways to report intensity, that wording has been replaced with "such as".

Just how specific must we become when giving examples of units?  These use inconsistent notation: 1/cm/sr, cm^2, but they are easily understood.  It is tedious to read through cm^-1 sr^-1 (or variations).  If this becomes a barrier, let's drop the example text, like the Poisson statistics above.

prjemian:
@smk78 asked:

    Where should one put the name of the data processing/analysis program: Under NXprocess : name, or under NXprocess : description?

The NXprocess group has a field ``program`` : Name of the program used.
http://download.nexusformat.org/doc/html/classes/base_classes/NXprocess.html

prjemian:
@smk78 wrote:

    Can we expand on the meanings of NXcollimator : aperture and NXsource : beam_shape, please, to clarify how they are different?

Can you offer a suggestion?

prjemian:
revised PDF: 
[NXcanSAS.pdf](https://github.com/nexusformat/definitions/files/632128/NXcanSAS.pdf)

revised structure (includes field attributes):

	NXcanSAS (application definition, version 1.0)
	  (overlays NXentry)
	  NXentry
		@default
		@canSAS_class
		@version
		definition:NX_CHAR
		run:NX_CHAR
		  @name
		title:NX_CHAR
		NXcollection
		  @canSAS_class
		NXdata
		  @canSAS_class
		  @signal
		  @I_axes
		  @Q_indices
		  @Mask_indices
		  dQl:NX_NUMBER
		  dQw:NX_NUMBER
		  I:NX_NUMBER
		    @uncertainties
		    @scaling_factor
		  Idev:NX_NUMBER
		  Q:NX_NUMBER
		    @uncertainties
		    @resolution
		    @resolution_description
		  Qdev:NX_NUMBER
		  Qmean:NX_NUMBER
		  ShadowFactor:NX_CHAR
		NXdata
		  @canSAS_class
		  @signal
		  @T_axes
		  @name
		  @timestamp
		  lambda:NX_NUMBER
		  T:NX_NUMBER
		    @uncertainties
		  Tdev:NX_NUMBER
		NXinstrument
		  @canSAS_class
		  NXcollimator
		    @canSAS_class
		    distance:NX_NUMBER
		    length:NX_NUMBER
		    aperture:NXaperture
		      shape:NX_CHAR
		      x_gap:NX_NUMBER
		      y_gap:NX_NUMBER
		  NXdetector
		    @canSAS_class
		    beam_center_x:NX_FLOAT
		    beam_center_y:NX_FLOAT
		    name:NX_CHAR
		    pitch:NX_NUMBER
		    roll:NX_NUMBER
		    SDD:NX_NUMBER
		    slit_length:NX_NUMBER
		    x_pixel_size:NX_FLOAT
		    x_position:NX_NUMBER
		    yaw:NX_NUMBER
		    y_pixel_size:NX_FLOAT
		    y_position:NX_NUMBER
		  NXsource
		    @canSAS_class
		    beam_shape:NX_CHAR
		    beam_size_x:NX_NUMBER
		    beam_size_y:NX_NUMBER
		    incident_wavelength:NX_NUMBER
		    incident_wavelength_spread:NX_NUMBER
		    radiation:NX_CHAR
		    wavelength_max:NX_NUMBER
		    wavelength_min:NX_NUMBER
		NXprocess
		  @canSAS_class
		  date:NX_DATE_TIME
		  description:NX_CHAR
		  name:NX_CHAR
		  term:NX_CHAR
		  NXcollection
		    @canSAS_class
		  NXnote
		NXsample
		  @canSAS_class
		  details:NX_CHAR
		  name:NX_CHAR
		  pitch:NX_NUMBER
		  roll:NX_NUMBER
		  temperature:NX_NUMBER
		  thickness:NX_FLOAT
		  transmission:NX_NUMBER
		  x_position:NX_NUMBER
		  yaw:NX_NUMBER
		  y_position:NX_NUMBER

prjemian:
This recent version addresses many points raised by @smk78 above.  Other points will need consensus before the NXDL is revised.

prjemian:
@smk78 wrote above

    Under NXsample : x/y_position: ...

the two fields, ``x_position`` and ``y_position`` are intended to document the setting of the stage motors that hold the sample.  How this information is useful for or participates in any small-angle scattering analysis process is beyond me.  Seems more like documentation about the sample name.  Such as "name" was moved to position (x,y) for this I(Q).  We've done that type of investigation using very small beam sizes and picking off different layers from a layered sample.

For the use you describe, the same question applies.  How will this information be useful to analysis?

As for adding a ``z_position``, that is OK to add to any data file.  Add it to NXcanSAS?

smk78:
Sorry @prjemian . I missed NXprocess : program field!

smk78:
@prjemian asked for a suggestion about clarifying the difference between NXcollimator : aperture and NXsource : beam_shape, etc. My attempt:

NXsource : beam_shape
Text description of the shape of the beam actually incident on the sample. This will normally describe the final pinhole/aperture/slit right before the sample, unless there is an absorbing mask on the sample itself.
NXsource : beam_size_x / NXsource : beam_size_y
Size of the incident beam along the x/y axis as defined by NXsource : beam_shape.

NXcollimator : aperture
Describes any collimating pinhole/aperture/slit in the primary collimation of the instrument, such as might be used to vary the flux into guides/polarisers/focussing elements. Most instruments will have more than one of these. Not to be confused with NXsource : beam_shape.
NXcollimator : aperture : shape
Text description of the type and shape of aperture (eg, pinhole, 4-blade slit, Soller slit)
NXcollimator : aperture : x_gap / NXcollimator : aperture : y_gap
Size of beam along the x/y axis as defined by NXcollimator : aperture : shape.

smk78:
Responding to @prjemian regarding NXsample : x_position/y_position:

If these are indeed intended to be stage motor readings then the value of capturing them in the reduced data file does indeed seem thin, as without a lot more metadata they will be meaningless!

As I previously alluded to, there are occasions (eg, rheometers, microfludic cells, stop-flow, chunks of metal/rock, etc) where the 'sample' is physically the same but the experiment is examining different bits of it (ie, scans/rasters).

If someone put multiple NXdata blocks, representing data collections on different bits of the sample, in a NXcanSAS file, how would they identify what the data in those blocks was from?

prjemian:
In https://github.com/nexusformat/definitions/pull/516#issuecomment-265139422, @smk78 wrote about multiple NXdata blocks with different SAS data from same sample:

The best example of such (not exactly your case, but a variation for sure) is presented in the canSAS 1-D data examples, the contrast variation SANS data from a steel alloy with various heat treatments.  Each heat treatment is a separate SASentry.  For each SASentry, there are two SASdata representing horizontal and vertical sections from the 2-D detector data.  The horizontal section is parallel to the applied saturation (5T) magnetic field and represents nuclear scattering only.  The vertical section is perpendicular to the applied field and contains both nuclear *and* magnetic scattering.

See the structure file here:
https://github.com/canSAS-org/NXcanSAS_examples/blob/master/1d_standard/structure/cs_af1410.h5.txt
The data file has 10 heat treatments: 10 SASentry groups.  Look at any one of them.  The SASdata groups are coupled with ``run`` fields that provide explanatory text.  These were translated directly from canSAS1d/1.1.  If they were written fresh in NXcanSAS, I would have chosen to put that explanatory text inside each SASdata group.

smk78:
So, if I summarise:

Multiple NXdata blocks in a NXentry should be avoided _except_ for storing I(Q) and the corresponding T(lambda) data. If you want to store multiple sample measurements in the same NXcanSAS file use multiple NXentry's. If we do this then we can do away with NXsample : x_position / y_position / roll / pitch / yaw and allow any explanatory text of what part of the sample was actually measured or what conditions the measurement was made under to be put in the NXsample : details free text field.

prjemian:
In https://github.com/nexusformat/definitions/pull/516#issuecomment-265135227, @smk78 presented explanations for NXsource and NXaperture.

## NXsource v. NXbeam

The description given for NXsource exactly fits the use of NXbeam.  NXsource describes *The neutron or x-ray storage ring/facility* while NXbeam describes *Properties of the neutron or X-ray beam at a given location.*

An NXbeam group can be added to NXsample to describe the beam incident on the sample.  We do not need to add this to NXcanSAS, but we *can* add an explanation about what might be done.

## NXaperture, NXcollimator, *et al.*

Apertures and collimators are two different types of device in the same broader category of devices that modifies the collimation and shape of the beam.  An aperture is 2-D (both orthogonal to direction of the beam)and modifies the shape of the beam.  An aperture must work with some other device to collimate the beam.  A collimator is 3-D and modifies the angular dispersion of the beam.

The NXcanSAS standard takes its use of NXaperture from canSAS1d in which these terms were combined.  Now, within NeXus, I propose we separate our use into the appropriate device.  We have this structure now:

    ...
    NXinstrument
      ...
      NXcollimator
        ...
        NXaperture

Does not look right to me now.  Both NXaperture and NXcollimator should be under NXinstrument, as in:

    ...
    NXinstrument
      ...
      NXaperture (or NXpinhole or NXslit)
      NXcollimator


## Links to relevant NeXus base classes:

* http://download.nexusformat.org/doc/html/classes/base_classes/NXaperture.html
* http://download.nexusformat.org/doc/html/classes/base_classes/NXbeam.html
* http://download.nexusformat.org/doc/html/classes/base_classes/NXcollimator.html
* http://download.nexusformat.org/doc/html/classes/base_classes/NXsource.html

## Notes

NeXus is starting to move away from NXgeometry towards a description more consistent with CIF (see issue #397).  This complicates its use in NXcanSAS at this time.  In my opinion, we should avoid NXgeometry (and NXshape) now, and provide terms within NXcanSAS that describe what we have in mind.

There is also interest in NeXus to move towards an object model, so that NXslit and NXpinhole would both derive common fields from NXaperture.  This has not happened yet and the discussions are not progressing due to maintaining consistency with existing data files.

prjemian:
re https://github.com/nexusformat/definitions/pull/516#issuecomment-265190140:

I tend to agree, NXsample : x_position / y_position / roll / pitch / yaw are metadata and not useful for the analysis of I(Q).  Is it useful to standardise (!) these names (and their spelling), should one choose to include them in a data file?

smk78:
Re: @prjemian  comment:

I was merely seeking to clarify _our_ usage of NXsource, NXcollimator, NXaperture given our _current_ structure! I would be content with a change to the structure.

prjemian:
You helped realize the current inconsistency.

toqduj:
Perhaps someone can offer a quick explanation about the current discussion.
I thought NeXus can fully specify the entire experimental configuration for
a measurement, from source definitions down to the pinholes and detector.
Why are we doing that all over again for the NXcanSAS format?

prjemian:
see: https://github.com/nexusformat/definitions/issues/492#issuecomment-256432708 and discussion that followed

summary: NXcanSAS is an extension of the canSAS1d standard to include multi-dimensional data, expressed in the NeXus data format.  In NXcanSAS, we are standardizing the spelling and usage of the names and structures, should one choose to use them.

Please refer to the second bullet point above (https://github.com/nexusformat/definitions/pull/516#issue-192754283).

smk78:
Re @toqduj comment:

We have accepted that certain elements of the experimental configuration and measurement conditions constitute useful metadata for interpreting reduced data. We are now just trying to ensure that we aren't including any unnecessary/pointless elements in NXcanSAS, and that our usage of what is left is consistent and understandable and sensibly structured.

prjemian:
revised PDF and structure files:

* [NXcanSAS.pdf](https://github.com/nexusformat/definitions/files/645016/NXcanSAS.pdf)
* [NXcanSAS-xture.txt](https://github.com/nexusformat/definitions/files/645018/NXcanSAS-xture.txt)

smk78:
In NXcanSAS-xture I note that Idev, Q and Qdev all have @units = NX_PER_LENGTH but that I has no specified @units. I can sort of understand why this is (eg, SAXS data that's just counts) but then surely Idev would not have any units, so either way there seems to be an inconsistency?

smk78:
Re my comments 7 days ago: why does the data block for SAStransmission_spectrum have @timestamp but the data block for SASdata does not?

smk78:
Wouldn't @units for Tdev be NX_DIMENSIONLESS?

smk78:
In NXcanSAS.pdf I notice we still have the direction to use 1/SQRT(I) for Poisson statistics. @adrianrennie cautioned against this above.

smk78:
Actually @prjemian , before I get carried away here, is the pdf at https://github.com/nexusformat/definitions/files/645016/NXcanSAS.pdf (your link 9 hours ago) really the revised version? Only I notice it is timestamped 10/19/2016 04:19 PM !!!

prjemian:
re https://github.com/nexusformat/definitions/pull/516#issuecomment-266407654: confirmed that "Poisson" text is removed from the current version.  PDF file is stale.

prjemian:
to *ensure* that HTML is completely rebuilt, use this command from root-level directory:

    make {rm,make}builddir

prjemian:
revised PDF and structure files:

* [NXcanSAS.pdf](https://github.com/nexusformat/definitions/files/646587/NXcanSAS.pdf)
* [NXcanSAS-xture.txt](https://github.com/nexusformat/definitions/files/646588/NXcanSAS-xture.txt)

butlerpd:
Have lost track of where we are with this.  Is it released? or still waiting for check marks from me, Steve and Adrian?  If the latter I would check but as mentioned somewhere else I apparently have no such privilege ...  The only thing I see is that resolution should be plural as there are many sources to the resolution not just one ..... JOKE .. no really ... that was a joke :-)

Seriously feel free to check me off -- As  Steve states "we are putting the finishing touches to here is a file format, and not an attempt to dictate how the global SAS community ought/should/could represent their data" -- I think the current state of the spec, as I've said before, represents a giant step forward and now is sufficiently agnostic to issues which may still be controversial to allow future versions to improve on it if and when consensus is reached, and further does not overly prescribe requirements that could stifle innovation.  Are there more issues that need addressing before releasing this version?

prjemian:
With a4a0477, butlerpd will be "ticked off".

@smk78 & @adrianrennie, are you both satisfied?  I'd like to merge this over the weekend.

prjemian:
revised PDF and structure files:

* [NXcanSAS.pdf](https://github.com/nexusformat/definitions/files/657489/NXcanSAS.pdf)
* [NXcanSAS-xture.txt](https://github.com/nexusformat/definitions/files/657491/NXcanSAS-xture.txt)

adrianrennie:
A tick can be placed by my name.

smk78:
Go ahead, Pete. I won't be able to review any more before Wed.

Steve

Sent from my Windows 8.1 Phone
________________________________

butlerpd:
Right ... teach me to make a joke --- won't happen again.
Thanks Pete for the hard work and perserverence!

prjemian:
Ok, we're all in.  Once I figure out the tagging procedure that NeXus has agreed upon, I'll merge.

Thanks, all, for the hard work.  And the giggles, too.

Issue #546, opened 2017-03-09
NXmx uses minOccurs where it is not allowed


prjemian:
The [travis-ci](https://travis-ci.org/nexusformat/definitions/jobs/209217121) build server has identified an error in the definition of [NXmx](https://github.com/nexusformat/definitions/blob/master/applications/NXmx.nxdl.xml#L176):

    <dim index="4" value="k" minOccurs="0"/>

Here is the error report:

	======================================================================
	ERROR: test__2__applications__NXmx (test_nxdl.Individual_NXDL_Tests)
	----------------------------------------------------------------------
	Traceback (most recent call last):
	  File "/home/travis/build/nexusformat/definitions/utils/test_nxdl.py", line 90, in test_wrap
	    validate_xml(nxdl_file_name)
	  File "/home/travis/build/nexusformat/definitions/utils/test_nxdl.py", line 61, in validate_xml
	    raise NXDL_Invalid(msg)
	NXDL_Invalid: applications/NXmx.nxdl.xml : Element '{http://definition.nexusformat.org/nxdl/3.1}dim', attribute 'minOccurs': The attribute 'minOccurs' is not allowed., line 176

prjemian:
@yayahjb Does the use of ``minOccurs`` mean that the ``dim`` with ``index=4`` is optional?  That is, ``dataRank`` could be either 3 or 4?

The ``nxdl.xsd`` [rules](https://github.com/nexusformat/definitions/blob/master/nxdl.xsd#L1158) do not have a way to express that.  Yet.  (This would be given as one of the ``<xs:attribute`` nodes.)  

Seems like a reasonable thing to add (to ``nxdl.xsd``) but we need a way to express that ``dataRank`` might have a range of values.  The name ``minOccurs`` suggest that there *might* be more than one while ``required="false"`` (the default would be ``required="true"``) is more in keeping with the current definition.  When ``required="false"`` is given, it **must** be present on any ``<dim`` nodes with a higher ``index=`` attribute.

prjemian:
@yayahjb 

This might be the addition to [nxdl.xsd](https://github.com/nexusformat/definitions/blob/master/nxdl.xsd#L1218) as an additional attribute to the ``<dim`` node:

	<xs:attribute name="required" type="nx:NX_BOOLEAN" default="true">
		<xs:annotation>
			<xs:documentation>
				This dimension is required (true) or not required (false).

				The default value is ``true``.  When ``required="false"`` is
[...]

PR #550, opened 2017-03-12
Issue 546 minOccurs=0 to required=false

prjemian:
* #546: minOccurs suggests possible multiple use but this is not allowed for attributes.  Replaces with new ``required`` attribute


prjemian:
Because this corrects a syntax error that interrupts validation, it is high-priority to resolve.

Issue #556, opened 2017-04-09
NXmx has incorrect use of "dimensions"


prjemian:
In the `incident_beam_size` field definition, the structure given as

    <dimensions rank="1" value="2"/>

should be written as

    <dimensions rank="1">
        <dim index="1" value="2"/>
    </dimensions>

This will specify that `incident_beam_size` is a 1-D array of length 2.

Issue #547, opened 2017-03-12
NXtransformations adds units attribute to vector attribute


prjemian:
This is an error (in both HDF5 and XML).

Neither HDF5 nor XML attributes can have attributes themselves. As-is, this would result in a specification where a data file would specify a ``vector`` attribute which would, by the specification now, have a ``units`` attribute which could be assigned a value compatible with ``NX_DIMENSIONLESS``.

prjemian:
Here is the code with the error:

		<attribute name="vector" units="AXISNAME" type="NX_NUMBER">
			<doc>
				Three values that define the axis for this transformation.
				The axis should be normalized to unit length, making it
				dimensionless.  For ``rotation`` axes, the direction should be
				chosen for a right-handed rotation with increasing angle.
				For ``translation`` axes the direction should be chosen for
				increasing displacement.
			</doc>
			<dimensions rank="1" value="3" />
		</attribute>

prjemian:
Because is a design error that interrupts validation, it is high-priority to resolve.

prjemian:
suggest adding a ``vector_units`` attribute to provide the requested information

zjttoefs:
The documentation clearly states that vector is dimensionless. A units attribute is not required.

Issue #545, opened 2017-03-09
NXtransformations has invalid unit: AXISUNITS


prjemian:
The [travis-ci build server ](https://travis-ci.org/nexusformat/definitions/jobs/209217121) reports an error with [NXtransformations](https://github.com/nexusformat/definitions/blob/master/base_classes/NXtransformations.nxdl.xml#L104)

    NXDL_Invalid: base_classes/NXtransformations.nxdl.xml : Element '{http://definition.nexusformat.org/nxdl/3.1}field', attribute 'units': [facet 'enumeration'] The value 'AXISUNITS' is not an element of the set {'NX_ANGLE', 'NX_ANY', 'NX_AREA', 'NX_CROSS_SECTION', 'NX_CHARGE', 'NX_CURRENT', 'NX_DIMENSIONLESS', 'NX_EMITTANCE', 'NX_ENERGY', 'NX_FLUX', 'NX_FREQUENCY', 'NX_LENGTH', 'NX_MASS', 'NX_MASS_DENSITY', 'NX_MOLECULAR_WEIGHT', 'NX_PER_AREA', 'NX_PER_LENGTH', 'NX_PERIOD', 'NX_POWER', 'NX_PRESSURE', 'NX_PULSES', 'NX_SCATTERING_LENGTH_DENSITY', 'NX_SOLID_ANGLE', 'NX_TEMPERATURE', 'NX_TIME', 'NX_TIME_OF_FLIGHT', 'NX_VOLTAGE', 'NX_VOLUME', 'NX_UNITLESS', 'NX_WAVELENGTH', 'NX_WAVENUMBER'}., line 104

A comment in the [NXDL file ](https://github.com/nexusformat/definitions/blob/master/base_classes/NXtransformations.nxdl.xml#L37) says:

    nxdl.xsd rules do not allow us to show this as a variable name
    - we'll use AXISNAME in ALL CAPS for consistency with the
    use in NXDATA.  We'll use ``AXISUNITS`` for the units appropriate
    to the type of axis.

NXDL files must adhere to the rules of ``nxdl.xsd``.  As stated, there is no unit type that satisfies the intent in this NXDL class.

prjemian:
One way to resolve this is to add a unit type to ``nxdlTypes.xsd`` in the simpleType ``anyUnitsAttr`` that [describes this particular usage](https://github.com/nexusformat/definitions/blob/master/base_classes/NXtransformations.nxdl.xml#L53).

[...]

PR #558, opened 2017-04-09
Schema update 555

prjemian:
* changes to XML Schema, backwards-compatible
* #545 NXtransformations needs new units category: NX_TRANSFORMATION
* #546 NXmx use of `minOccurs` changed to `optional`
* #547 NXtransformations correction
* #556 allow declarations of dimensions for data values in attributes

locally: `utils/test_suite.py` passes and the Sphinx documentation builds


prjemian:
We should bump the minor version number of the XML Schema.  Here is the definition from `nxdl.xsd`:

    <xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
	targetNamespace="http://definition.nexusformat.org/nxdl/3.1"
	xmlns:nx="http://definition.nexusformat.org/nxdl/3.1" 
	elementFormDefault="qualified">

BUT, that will change the namespace and that can have consequences that affect anything that uses the XML Schema.

**Should** we bump the minor version?  Any other way to differentiate from the previous XML Schema?

prjemian:
@yayahjb Comments?  tacit approval is not satisfying

prjemian:
I will examine the consequences of changing the minor version number *without* changing the namespace string.  It may well be that we leave the namespace string as-is, at least for now.

zjttoefs:
Sorry, forgot to comment on the version number. I would say: Bump it. 
We might get away not changing it, but having version numbers we're too afraid to ever change is a bit pointless. We will see what breaks and have to prepare that for future updates.

PR #564, opened 2017-04-21
Renames relative_molecular_mass

mtwharmby:
relative_molecular_mass was carried over from the contributed 
definition NXcontainer. For consistency with the chemical_formula field
and also with (e.g.) the CIF definitions, it should be renamed 
chemical_formula_weight.

This was discussed (some time ago) with @markbasham 

### DIFF
--- a/base_classes/NXsample.nxdl.xml
+++ b/base_classes/NXsample.nxdl.xml
@@ -182,8 +182,11 @@
 			<dim index="1" value="n_comp"/>
 		</dimensions>
 	</field>
-	<field name="relative_molecular_mass" type="NX_FLOAT" units="NX_MASS">
-		<doc>Relative Molecular Mass of sample</doc>
+	<field name="chemical_formula_weight" type="NX_FLOAT" units="NX_MASS">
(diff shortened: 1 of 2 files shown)


prjemian:
Added to [2017-07-25 telco](https://github.com/nexusformat/wiki/edit/master/content/Telco_20170725.md) that this PR *will be merged unless there are objections raised*.

zjttoefs:
relative_molecular_mass is the historical NeXus term and I see no reason to change that by the way of a pull request, especially since the field was only re-confirmed in NXsample_component last October.

mtwharmby:
Rather than closing, might it not be worth discussing?
As a reason to change the name: consistency with other fields in NeXus (e.g. chemical_formula) and also with CIF format. Also by having similar names, when looking at the NeXus tree, chemical_formula and chemical_formula_mass/weight appear together, which helps readability.

Also, the last time I heard anyone use the term relative_molecular_mass was in a high school science text book (though searching the web, hit number five on Google is the IUPAC gold book)

PR #565, opened 2017-04-21
Adds missing nComp dimensions

mtwharmby:
Added nComp arrays to four fields which need them. This is only necessary 
if it is intended to continue using the nComp arrays, rather than using 
NXsample_component. Nxsample_component looks like the better solution,
however.


zjttoefs:
The idea at the time was to prefer NXsample_component and maintain legacy support for nComp. It was a conscious decision not to add nComp to new fields.

mtwharmby:
Fine. That makes sense, I just noticed the discrepancy and thought it worth pointing out. Thanks! :-)

PR #566, opened 2017-04-21
Remove NXcontainer duplicates

mtwharmby:
This PR removes fields which were moved over into NXsample and are therefore no longer necessary in the NXcontainer class. This was discussed (again, some time ago!) with @markbasham 

### DIFF
--- a/contributed_definitions/NXcontainer.nxdl.xml
+++ b/contributed_definitions/NXcontainer.nxdl.xml
@@ -67,11 +67,14 @@
 		   the changes to the beampath would need to be accounted for.
 		
 		This class encodes the position of the container with respect to the 
-		sample and allows the calculation of the beampath through the container.
-		It also includes sufficient data to model beam absorption of the 
-		container and a link to a dataset containing a measurement of the 
-		container with nothing inside, to allow data corrections (at a specific
-		beam energy/measurement time) to be made.
+		sample and allows the calculation of the beampath through the 
+		container. Other data necessary to model beam absorption by the 
+		container should be included in the reference dataset. Typically, the 
+		reference dataset is a measurement of the empty container (included in 
+		this NXcontainer as a link) which can be used to make data corrections 
+		(at a specific beam energy/measurement time). The beam, orientation 
+		and shape fields are included to allow the same reference dataset to 
+		be used even if the container is used with a different geometry/energy.
 	</doc>
 	<field name="name">
 		<doc>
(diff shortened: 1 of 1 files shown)


zjttoefs:
I don't understand. What are the new fields in NXsample?

mtwharmby:
I reviewed NXcontainer with @markbasham a couple of months back and realised that density, chemical_formula, chemical_formula_weight (relative_molecular_mass) had either been added to NXsample since I wrote NXcontainer or were already present. Also, @markbasham and I discussed using an NXsample class to store these data for a container as this is/was what the NIAC had decided was a better method for storing these data.

The only problem I can see for a Pair Distribution Function experiment usecase is there will be no packing_fraction for the sample, which is a necessary field. I'll discuss this offline with @markbasham

zjttoefs:
But doesn't NXcontainer describe the container and NXsample the sample? These two different things can have different chemical formulae and densities. No? Wasn't that not the point for having NXcontainer?

stuartcampbell:
@zjttoefs yes i think you are correct and i agree the container material could be something that is useful to know.

mtwharmby:
That was the original idea, but I was given the impression that as a measurement of a container needs very similar information storing as an NXsample, we should be using NXsample. And as we need a few extra fields (e.g.  some way of showing the relationship between multiple containers) we still need to NXcontainer class.

If I should be using NXcontainer to describe container only measurements, I will need to add a few new fields and it runs the risk of duplicating a lot of what is already in NXsample/NXsample_component. I don't mind doing this, but it'll be a different pull request!

zjttoefs:
Agree, ignoring aspects like geometry and shape NXcontainer is just a special case of NXsample_component. Those are important aspects, though and they usually mean that you can measure the container on its own, but often not the sample (under the same conditions). Especially when you have an involved protocol of measurements to identify separate contributions to the data I see value in NXcontainer. 
We ran out to time to approve that in the last NIAC meeting as far as I can tell from the minutes and remember. There was no official line against it.

zjttoefs:
Either way it would be good to see a full working example of how you see that used, with or without these fields. If we don't have that, it could have been the reason we didn't discuss that in detail at the NIAC.
For the time being: Are you interested in keeping this pull request open?

mtwharmby:
I'll close it for now. I have an offline meeting with @markbasham in a couple of weeks to discuss this sort of thing.

PR #568, opened 2017-06-05
Updates from NIAC 2016

phyy-nx:
Action items from minutes completed:
-H,K,L change to NX_number for incommensurate structures
-Discussion on flags, change to bitfield, no enum for efficiency
-Expand the description of partiality
-Need to keep mm and pixel positions because of parallax
-Write out predicted and observed
-Drop the mm in names, rather use units
-bounding_box as array, document the usage
-Spell out background
-Drop the val on the intensities, keep the _var version
code is not in place we defer it for now.
-Add polar_angle and azimuthal_angle

TODO
-Need documentation for polar_angle and azimuthal_angle
-Need to add some examples of flags. Would like to use the DIALS flags.

Deferred:
-overlaps is a list of overlapped reflections. This is a ragged array. As the

### DIFF
--- a/contributed_definitions/NXreflections.nxdl.xml
+++ b/contributed_definitions/NXreflections.nxdl.xml
@@ -79,7 +79,33 @@
     </field>
 
     <field name="flags" type="NX_INT" minOccurs="1">
-      <doc> Status flags describing the reflection. This is a bit mask. XXX possible flags enumeration here </doc>
+      <doc> Status flags describing the reflection. This is a bit mask. The bits in the mask follow the convention used by DIALS, and have the following names:
+        * bit 0: predicted
+        * bit 1: observed
+        * bit 2: indexed
+        * bit 3: used_in_refinement
+        * bit 4: strong
+        * bit 5: reference_spot
+        * bit 6: dont_integrate
+        * bit 7: integrated_sum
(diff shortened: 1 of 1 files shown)


zjttoefs:
This should have addressed all NIAC comments.

prjemian:
In the *flags* fields, there is bit 17: ``bad_shoebox``.  It is not obvious what is a *shoebox* in this context.

phyy-nx:
dials/dials#363 aims to document all these flags, including `bad_shoebox`.  When those doc strings are in place, I'll put them here.

Question about the first dim "n" for the bounding box.  Should all the other columns have 'n' as a dimension?  For example, 'h', 'k', and 'l' should all be arrays of dimension n.

prjemian:
> Should all the other columns have 'n' as a dimension? For example, 'h', 'k', and 'l' should all be arrays of dimension n.

The existing NXDL files are not consistent in this regard.  Yours is a good idea, fits the current circumstances, and would inspire us to revise NXDL files similarly.  One reason for the inconsistency in other NXDL files is that some *fields* have a variable [shape](https://duckduckgo.com/?q=hdf5+dataset+shape&t=ffsb&ia=web).

phyy-nx:
Cool.  I'll get that going then.

PR #579, opened 2017-06-22
Update NXdetector.nxdl.xml

jacobfilik:
Add serial_number field to explicitly identify different detectors.

The serial number on a detector is a useful key to allow different detectors identified without having to parse the multiple values in the descriptions field.

Having a explicit field for the serial number would make it much easier to determine which detector was actually used for a measurement, especially in facilities with detector loan pools.


prjemian:
This is useful, it defines the spelling of this common metadata, and a search of the existing NeXus manual does not show `serial` in use.  I agree with this PR.

Issue #562, opened 2017-04-17
use consistent terms when names are flexible


prjemian:
In some NXDL files, the name for a group, field, or attribute is flexible (the exact name given by the NXDL specification is not required) and is set at the time the HDF file is written.  Across the suite of NXDL files, different terms are used to mark a name as flexible.  In some cases, there is no indication that a name may be substituted.  In others, upper case is used,  In some other cases, such as for groups that have not been given explicit names, a default is shown in parentheses.  A few, select examples should describe the variations:

| NXDL  |  type  |  name given |
| ------  | ----  | ----------- |
| NXdata | field | data |
| NXdata | field | variable |
| NXdata | attribute | AXISNAME_indices |
| NXentry | group | (data) |
| NXscan | group | (instrument) |
| NXtransformations | field | AXISNAME |
| NXtransformations | field | AXISNAME_end |

The documentation publishing process applies the parentheses for groups where the name is not specified in the NXDL file.  Such as (from NXentry):

    <group type="NXsample" />

is documented as

    (sample): (optional) NXsample

It is proposed to choose either UPPER_CASE or (parentheses) to denote when a name is flexible and to apply this decision across all NXDL files.  The proposed should not change the meaning of any definition, but just use consistent nomenclature.

prjemian:
Note that the procedures that publish the manual use the parenthesis style (`(suggestion)`).  Perhaps we standardize that.

prjemian:
There is an attribute to an NXDL *field* element that addresses this point:  `nameType="any"`

```
<xs:attribute name="nameType" use="optional" default="specified">
	<xs:annotation>
		<xs:documentation>
			This interprets the name attribute as:
			* ``specified`` = use as specified
			* ``any`` = can be any name not already used in group
		</xs:documentation>
	</xs:annotation>
	<xs:simpleType>
		<xs:restriction base="nx:NX_CHAR">
			<xs:enumeration value="specified"/>
			<xs:enumeration value="any"/>
		</xs:restriction>
	</xs:simpleType>
</xs:attribute>
```

prjemian:
If we enclose the flexible part of the name within parentheses, this will require a change to the XML Schema and the regular expression.

Suggest that we use all capital letters (in addition to the `nameType="any"` attribute) to indicate in the documentation when a name is flexible.  "All caps" looks simpler.

prjemian:
[...]

Issue #380, opened 2015-01-21
add attribute called "default" at root and NXentry level


zjttoefs:
that attribute indicates the default entry or default NXdata, depending on where it is

prjemian:
completed in 5ae7f8ec1ec7460be4825e125e36b9d1612aabba

Issue #270, opened 2014-05-05
start each base class documentation with one-line summary


stuartcampbell:
**Original reporter**: *[prjemian](https://github.com/prjemian)*

Propose to start the documentation of each base class (and other NXDL files, too) with a one-line summary.  With that, then the page that lists all the base classes together at the start of the base class documentation in the manual, could show this summary.  This page:

http://download.nexusformat.org/doc/html/classes/base_classes/index.html

will become much more informative for the community and still be fully generated from source file.

zjttoefs:
I think that's a good idea. If there was an example of how that would need to be formatted writing the texts would be easy.

prjemian:
tested OK on Linux - manual page built properly

still inconsistencies in use of "template" v. "description", as well as lower/upper case at start of doc summary - fix those and then submit for PR

prjemian:
also, will test on Windows as part of those last consistency changes to make

zjttoefs:
Reopened to gather opinions on the right start of those one-line summaries at the next telco.

rayosborn:
I'm sorry I haven't had time to look at this closely yet, but I think this a great start. There are a couple of minor tweaks that I would recommend in the wording. We should probably decide if we are going to explicitly define whether components are specific to neutrons or x-rays and then be consistent in implementing this. NXmoderator is always a neutron moderator, for example, whereas I guess NXfresnel_zone_plate is always an x-ray component. I'm also not sure that the NXmonochromator description is ideal. There are so many small changes we could suggest that perhaps this should be delayed until the NIAC meeting.

prjemian:
To enable the infrastructure, one line summaries were created from the existing documentation.  Attempt was made to make the summary text consistent amongst the various base classes.

Discussion of the descriptive summary wording for each NXDL file is separate from this issue and should be discussed separately.  This issue is about creating and installing the infrastructure to make this change in the reference documentation.  That infrastructure has been accepted in a recent merge, noted above.  

I am closing this issue as complete.  Please create a new issue to discuss the documentation of any particular base class that needs further revision.

PR #580, opened 2017-06-22
NXexperiment and NXexperimentdata 

jacobfilik:
Contributed definitions to allow NXdata groups to be related to each other and an NXdetector.

With the aim of making writing software for visualisation and processing of data collections with multiple detectors in the same scan easier by explicitly describing how the NXdata groups are released to each other and the corresponding NXDetector.

### DIFF
--- a/contributed_definitions/NXexperiment.nxdl.xml
+++ b/contributed_definitions/NXexperiment.nxdl.xml
@@ -1,0 +1,56 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?xml-stylesheet type="text/xsl" href="nxdlformat.xsl" ?>
+<!--
+# NeXus - Neutron and X-ray Common Data Format
+# 
+# Copyright (C) 2012-2017 NeXus International Advisory Committee (NIAC)
+# 
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 3 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+#
+# For further information, see http://www.nexusformat.org
+-->
+
+<definition name="NXexperiment" extends="NXobject" type="group"
+	    version="1.0"  
+	    category="contributed"
+	    xmlns="http://definition.nexusformat.org/nxdl/3.1"
+	    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	    xsi:schemaLocation="http://definition.nexusformat.org/nxdl/3.1 ../nxdl.xsd">
+	<doc>
+      Describe the contents of an NXentry resulting from a data collection using multiple detectors,
+      generating raw and derived data, potentially being scanned across multiple dimensions.
+      
+      It is not uncommon for a beamline at a synchrotron to collect data from multiple detectors in during a single scan. NeXus files
+      from these scans will contain many :ref:`NXdata` either containing direct links to :ref:`NXdetector` data, or data derived from
+      this data. The aim of the :ref:`NXexperiment` class is to store NeXus classes in a way that better represents the *experiment* performed,
+      rather than the *instrument* the experiment was performed on.
+      
+      For example, a small-angle scattering beamline with a detector for SAXS,
+      WAXS, XRF, a diode for transmitted intensity and an XY stage performs a single grid scan: The XRF raw data
+      might be have multiple channels that are summed to give the MCA, then small regions of this sum may also
+      be summed to produce simple elemental maps.
+      
+      This file contains multiple plottable datasets, therefore multiple NXdata groups and NXdetectors, 
+      but with no obvious way to discern what was measured and how the plottable datasets are related.
+      
+      The :ref:`NXexperimentdata` class allows (multiple) :ref:`NXdata` to be related to each other (through the ``DATANAME_origin`` attribute in :ref:`NXexperimentdata`), and back to the :ref:`NXdetector`
+      used to collect the data.
+      
+      (**required**) :ref:`NXexperimentdata` container for :ref:`NXdata` and single :ref:`NXdetector` used for the measurement
+      
+  </doc>
+  <group type="NXexperimentdata" />
+  
+</definition>
--- a/contributed_definitions/NXexperimentdata.nxdl.xml
+++ b/contributed_definitions/NXexperimentdata.nxdl.xml
@@ -1,0 +1,56 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?xml-stylesheet type="text/xsl" href="nxdlformat.xsl" ?>
+<!--
+# NeXus - Neutron and X-ray Common Data Format
+# 
+# Copyright (C) 2012-2017 NeXus International Advisory Committee (NIAC)
+# 
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 3 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+#
+# For further information, see http://www.nexusformat.org
+-->
+
+<definition name="NXexperimentdata" extends="NXobject" type="group"
+	    version="1.0"  
+	    category="contributed"
+	    xmlns="http://definition.nexusformat.org/nxdl/3.1"
+	    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	    xsi:schemaLocation="http://definition.nexusformat.org/nxdl/3.1 ../nxdl.xsd"
+	    >
+	<doc>
+		Describe all the plottable data from a single detector in a single entity. To be used
+		inside :ref:`NXexperiment`
+		
+		Allows data to be related back to the detector that collected it.
+		
+		(**required**) :ref:`NXdetector` detector used to acquire data.
+		
+		(**required**) :ref:`NXdata` at least one NXdata, related to the NXdetector
+  </doc>
+  <attribute name="DATANAME_origin">
+    <doc>
+      Best description of the origin of this data, use . for the :ref:`NXdetector`. ``DATANAME`` refers to the 
+      name of an :ref:`NXdata` that must be in the same :ref:`NXexperimentdata`.
+    </doc>
+  </attribute>
+  <attribute name="primary">
+    <doc>
+      Default :ref:`NXdata` to plot to represent this detector.
+    </doc>
+  </attribute>
+  <group type="NXdetector" />
+  <group type="NXdata" />
+ 
+</definition>


contributed_definitions/NXexperiment.nxdl.xml
```diff
@@ -0,0 +1,56 @@
...
+-->
+
+<definition name="NXexperiment" extends="NXobject" type="group"
+	    version="1.0"  
+	    category="contributed"
+	    xmlns="http://definition.nexusformat.org/nxdl/3.1"
+	    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	    xsi:schemaLocation="http://definition.nexusformat.org/nxdl/3.1 ../nxdl.xsd">
+	<doc>
+      Describe the contents of an NXentry resulting from a data collection using multiple detectors,
```
prjemian: per #469 & #270, make the first line a one-line summary since only the first line will be used to generate the summary documentation about each NXDL shown on http://download.nexusformat.org/doc/html/classes/contributed_definitions/index.html

contributed_definitions/NXexperiment.nxdl.xml
```diff
@@ -0,0 +1,56 @@
...
+      generating raw and derived data, potentially being scanned across multiple dimensions.
+      
+      It is not uncommon for a beamline at a synchrotron to collect data from multiple detectors in during a single scan. NeXus files
```
prjemian: remove `in`

contributed_definitions/NXexperimentdata.nxdl.xml
```diff
@@ -0,0 +1,56 @@
...
+
+<definition name="NXexperimentdata" extends="NXobject" type="group"
+	    version="1.0"  
+	    category="contributed"
+	    xmlns="http://definition.nexusformat.org/nxdl/3.1"
+	    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	    xsi:schemaLocation="http://definition.nexusformat.org/nxdl/3.1 ../nxdl.xsd"
+	    >
+	<doc>
+		Describe all the plottable data from a single detector in a single entity. To be used
```
prjemian: per #469 & #270, make the first line a one-line summary since only the first line will be used to generate the summary documentation about each NXDL shown on http://download.nexusformat.org/doc/html/classes/contributed_definitions/index.html

contributed_definitions/NXexperimentdata.nxdl.xml
```diff
@@ -0,0 +1,56 @@
...
+		inside :ref:`NXexperiment`
+		
+		Allows data to be related back to the detector that collected it.
+		
+		(**required**) :ref:`NXdetector` detector used to acquire data.
+		
+		(**required**) :ref:`NXdata` at least one NXdata, related to the NXdetector
+  </doc>
+  <attribute name="DATANAME_origin">
```
prjemian: add `nameType="any"` (per #562) to indicate that *DATANAME* is not required text in the name

contributed_definitions/NXexperimentdata.nxdl.xml
```diff
@@ -0,0 +1,56 @@
...
+    <doc>
+      Best description of the origin of this data, use . for the :ref:`NXdetector`. ``DATANAME`` refers to the 
+      name of an :ref:`NXdata` that must be in the same :ref:`NXexperimentdata`.
+    </doc>
+  </attribute>
+  <attribute name="primary">
```
prjemian: change name of this attribute to `default` (per #380)  

documentation should mention that an absolute HDF5 path address (starts with `/`, such as `/entry/instrument/detector1/ccd3`) must be used.

contributed_definitions/NXexperimentdata.nxdl.xml
jacobfilik: Just to check, the issue suggests using this in fields, whereas this is an attribute. No other attributes (such as AXISNAME_indices) follow this convention.

contributed_definitions/NXexperimentdata.nxdl.xml
jacobfilik: I would prefer not to have to use the absolute HDF5 path for this, just the name of (one of) the NXdata in this class.

This would be expected to be used more like the `signal` attribute in `NXData`, that is, from my collection of objects, which is the one i should actually care about. If it is a path to a location outside the group, it loses all the connections in the group.
For example, NXexperimentdata contains
```
Xspress3 (/entry/instrument/detector)[NXDetector]
AllElementSum(/entry/AllElementSum)[NXData]
FeKAlpha(/entry/FeKAlpha)[NXData]
```
with `default` wanting to refer to FeKalpha (which has an `origin` of AllElementSum, which has an `origin` of Xspress3). If default was /entry/FeKalpha I have to strip the HDF5 path to make any use of the `origin` attributes.
There is the requirement that all the `NXdata` groups in this class must have unique names (which could be explicitly stated in the doc) but if there are mulitple NXdata, derived from the same NXDetector, with the same name, that suggests to me bigger issues with the file.

Since this behaviour breaks with the use of `default` in the issue (by not using the full path), but would work with the external `default` definition (I assume there is no reason why the `default` in an `NXentry` couldn't contain the full path to the `NXdata` in an `NXexperimentaldata`), maybe a different name would be better.

rayosborn:
Thanks for making this suggestion, which looks interesting, but I'm struggling to understand exactly how these classes would be used. Could you provide an example file tree, such as your SAXS example, with its various detectors?

jacobfilik:
Happy to, but it will take a few days to find suitable example data and construct the file by hand.

The main issue the classes aim to solve is relating `NXdata` to each other and `NXdetector` groups. For simple files with one `NXdata` and one `NXdetector` the problem is trivial.

The minute there are multiple `NXdata` and `NXdetector` groups there is no way to relate them, except potentially by comparing names/strings which can't be good. (If there is an official NeXus way to achieve this that I have overlooked I am happy to be educated). Relating an `NXdata` to the `NXdetector` is crucial, if you want to, say, convert from pixel to q space.

Also, for a grid scan, if one 'NXdata' contains the detector sum (which would be a 2D image), and another contains the detector data (a 3 or 4D block) there seems to be no way to encode this information. Again for one detector you could guess, but for N detectors you are back to comparing names.

I have no emotional attachment to these classes. @markbasham and I have discussed around this issue for a long time, and this seems to be one method of encapsulating the information we need, without just adding more fields to the existing base classes. I'm sure there are many elegant solutions, and I am happy to discuss the most NeXus way to get the information we need from the file.

rayosborn:
It is possible that the functionality that you are looking for is already provided by NeXus links. Here is part of the tree structure of one of the example files distributed with NeXpy:

    >>> print(chopper.tree)
    chopper:NXroot
      entry:NXentry
        analysis = 'TOFNDGS'
        data:NXdata
          @axes = ['polar_angle' 'time_of_flight']
          @signal = 'data'
          data = int32(148x750)
          polar_angle -> /entry/instrument/detector/polar_angle
            @target = '/entry/instrument/detector/polar_angle'
          time_of_flight -> /entry/instrument/detector/time_of_flight
            @target = '/entry/instrument/detector/time_of_flight'
        instrument:NXinstrument
          detector:NXdetector
            polar_angle = float32(148)
              @target = '/entry/instrument/detector/polar_angle'
            time_of_flight = float32(751)
              @target = '/entry/instrument/detector/time_of_flight'

So `polar_angle` and `time_of_flight` are added to the `/entry/data` NXdata group as links to fields that are stored in the `/entry/instrument/detector` group. The target attribute can be used to identify the NXdetector group from which the NXdata axes are derived. One of the rationales for treating these as soft links within NeXus (even though they are stored as hard links in the HDF5 file) is so that you can identify what detector group it comes from. 

It also allows the mixing of axes from multiple groups. For example, if you are scanning both incident energy and detector angle, the former could be stored in a NXmonochromator group and the latter in a NXdetector group.

Your method may be more transparent, although high-level APIs can make the association clearer. For example, in the attached image from NeXpy, the chain-link icon shows that the item is a link. If you right-click on the link, one of the options is `Show Link`, which will automatically select the target field in the detector group.

tree-view

Let us know if this meets your needs, or if you think new classes will still be useful. We will certainly consider them if you do.

jacobfilik:
Thank you, I had seen the `@target` attribute in Nexus files before, but I could never find anything in the documentation, so assumed it was something internal. This would work perfectly for my first use case (better in-fact including the axes example you show), but I would want to add some doc to the NXData class to describe this use-case. Assuming `NXdata` is the main place `@target` is used.

I think the second use case (multiple `NXData`, some derived from processing others) could be met by clever use of `@target` and `NXprocess`. I will have a think about this and possibly submit an alternative pull request, so the relative merits of each can be discussed.

Thanks again for taking time to review and discuss this request, it is greatly appreciated.

prjemian:
Good point.  The `@target` attribute is more general than just the `NXdata` base class, it is common to all links in a NeXus data file.  And you have identified another [TODO](http://download.nexusformat.org/doc/html/fileformat.html?highlight=hardlink#mapping-nexus-into-hdf) item in the NeXus manual.  The [only documentation is an h5dump example of a NeXus link](http://download.nexusformat.org/doc/html/fileformat.html?highlight=hardlink#h5dump-link).

That documentation will be addressed with (new) issue #581.

prjemian:
This is where the `@target` attribute is defined in NeXus: http://download.nexusformat.org/doc/html/nxdl_desc.html#linktype

prjemian:
The [Design](http://download.nexusformat.org/doc/html/design.html#links) chapter of the manual describes NeXus links and gives an example use of the `@target` attribute.

jacobfilik:
Brilliant! Thank you, I will start making sure target attributes are added appropriately. I assume it is also correct to add them when the data is externally linked (e.g. unless the data is only written directly in the `NXdata` i should expect/write `@target`)?

Finally, is there an approved manner of representing that a certain `NXdata` contains the result of processing another `NXdata`? I am considering including an `NXprocess` in the processed 'NXdata`, showing the provenance. If this is acceptable (and ideally recommended) I will happily close this pull request, as all the functionality I hoped to add is already there.

prjemian:
There is not a standard set for `NXdata` contains `NXprocess`, it makes some sense.

Another way is to place both `NXprocess` and `NXprocess` as siblings within `NXentry`.  There may be one or more such `NXprocess` groups in a numbered sequence, describing a series of processing events.  Each one of those might result in a `NXdata` group.

prjemian:
The [NAPI code](https://github.com/nexusformat/code/blob/fe8ddd287ee33961982931e2016cc25f76f95edd/src/napi5.c#L2248) makes no `@target` attribute assignment for externally linked data.  

You'd best avoid using the `@target` attribute with external file links.  Should also be added to the documentation?  Seems reasonable [at this point](http://download.nexusformat.org/doc/html/design.html?highlight=external#nexus-base-classes).

jacobfilik:
Thank you so much for this again, I just want to check with one final example and then I will close the request (although at this point it sounds like I should try the NAPI if that is the reference implementation).

If I have some data in myfile.h5, and I make an eternal link to it in my NeXus file in the `NXdetector` at /entry/instrument/detector/data, this should have no `@target` (it would make absolutely no sense to have one).
If then in the same NeXus file I make an `NXdata` /entry/mydata/data which is just a link to the data in the `NXdetector` this _should_ have a `@target`, because it is important to show that this is the data from the detector (and it is a link to another dataset in the same file, even though the data is in an external file)?

prjemian:
The procedure sounds right.  BUT, HDF5 will not let you be do this ("Interfile hard links are not allowed").  Just tried it with Python and h5py, following the [external file example in the manual](http://download.nexusformat.org/doc/html/examples/h5py/index.html#source-code-externalexample-py).  Continuing that example:

```
In [62]: f = h5py.File(FILE_HDF5_MASTER, "a")

In [63]: ds = f["/entry/data/counts"]

In [64]: ds.name
Out[64]: u'/entry/instrument/detector/counts'

In [65]: ds.attrs["target"] = ds.name

In [66]: f["/entry/nexus_link"] = ds
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-66-659843c8241d> in <module>()
----> 1 f["/entry/nexus_link"] = ds

h5py\_objects.pyx in h5py._objects.with_phil.wrapper (C:\Minonda\conda-bld\h5py_1474482483473\work\h5py\_objects.c:2705)
()

h5py\_objects.pyx in h5py._objects.with_phil.wrapper (C:\Minonda\conda-bld\h5py_1474482483473\work\h5py\_objects.c:2663)
()

D:\Apps\Anaconda\lib\site-packages\h5py\_hl\group.pyc in __setitem__(self, name, obj)
    275
    276         if isinstance(obj, HLObject):
--> 277             h5o.link(obj.id, self.id, name, lcpl=lcpl, lapl=self._lapl)
    278
    279         elif isinstance(obj, SoftLink):

h5py\_objects.pyx in h5py._objects.with_phil.wrapper (C:\Minonda\conda-bld\h5py_1474482483473\work\h5py\_objects.c:2705)
()

h5py\_objects.pyx in h5py._objects.with_phil.wrapper (C:\Minonda\conda-bld\h5py_1474482483473\work\h5py\_objects.c:2663)
()

h5py\h5o.pyx in h5py.h5o.link (C:\Minonda\conda-bld\h5py_1474482483473\work\h5py\h5o.c:3619)()

RuntimeError: Unable to create link (Interfile hard links are not allowed)
```

Conclusion is that when you want to make a link to an object in a NeXus HDF5 data file and that object is itself externally linked, then instead of making a hard link at the new location, you make an external file link at the new location back to the data source.  Clear enough?

rayosborn:
Your external link example does reveal a limitation in using the `@target` attribute to logically associate two groups. The only way to add such an attribute to the externally linked HDF5 dataset is for it to be added to the dataset itself in the remote file, where it might not make sense, e.g., it might be an invalid path. In practice, or at least in my own experience, external links are most often used for very large multidimensional arrays, i.e., signal arrays rather than one-dimensional axis arrays. If you stored the axis arrays in the local file, you could still use them to associate the NXdata and NXdetector groups, as in the chopper example, even if the signal array is externally linked in both groups. But I agree that external links expose a flaw in this approach.

jacobfilik:
That's a shame, the `@target` attribute looks like exactly what I want, but a significant amount of our data is externally linked.

I don't suppose there is a plan to move the `@target` attribute off the data set into the group (in the same way `@signal` was moved)? Defining `@DATASETNAME_target` with all the axes and axes_indicies etc, feels logical to me.

(As an aside, should I close this pull request - i'm now not convinced this is the most NeXus approach - and open an issue pointing at it? Are more people likely to read pull request comments or issues?)

prjemian:
The `@target` attribute can be used with both fields (a.k.a. HDF5 datasets) **and** groups to indicate a link to another item in the same file.  No plans to change this as this is one of the older internal structures in NeXus, many software tools expect this to remain as they know it now.

Perhaps we re-examine your original intention, which was *to associate the NXdata and NXdetector groups*.  I agree that closing this PR without a merge is the right course, and open a new issue to discuss how best to implement your intention.  The scope may expand to include associations between NXprocess, NXlog, and NXnote as well.

mkoennecke:
Hi,

to add my 2c's worth: In a normal in-file link look at the target attribute. For an external link, NAPI 
writes the napimount attribute. Which contains a URL like string pointing into the external file. 

So, yes, we have the use case for NXexperiment_data well covered.

rayosborn:
Hi Mark,
This discussion has now been transferred to #583. I'm not sure the `@napimount` attribute does what Jacob needs. It points to the object in the remote file, but Jacob is asking how to link an NXdata and NXdetector group in the same file that happen to have the same external link. Also, does the `@napimount` attribute still exist once the HDF5 file has been written? I thought it was just used by the API in creating the external link. In HDF5, the external link in the parent file cannot have different attributes to the dataset in the remote file, so once the link has been made, these extra attributes must disappear. That is what happens in the Python nexusformat API.

jacobfilik:
Hi,

Thanks, but if this points to the external file, the only way I can find the corresponding NXDetector is to parse all the NXDetectors looking for the same url which I don't really want to do. Also, is the napimount atttribute well documented NeXus standard (I haven't seen it before, but I am certainly not a NeXus expert)?

Finally structuring code like:
```if (contains target attribute) {
//Can go directly to NXdetector
} else if (contains napimount attribute) {
//walk detectors finding same url
} else {
//walk detectors finding same HDF5 dataset id
}
```
doesn't seem ideal. I admit that the NXexperiment/NXexperimentdata is not the way we want to do things, but the above can't be the recommended approach?

jacobfilik:
Not the write solution for this functionality, discussion moved to issue #583

PR #589, opened 2017-07-20
NXdata: prefixed axes attribute?

tschoonj:
An example in the documentation contains a reference to an attribute called `@I_axes`, which suggests that these axes belong to a particular dataset in the NXdata, even though the standard does not mention this case.

Typo?


prjemian:
You're right, thanks!

Issue #530, opened 2017-01-13
add machine-readable unit string for each NeXus units type


prjemian:
To validate the engineering units provided with a field, each field is assigned one of the NeXus units types listed in the file ``nxdlTypes.xsd``.  At this time, NeXus states that the ``units`` attribute is not validated.  With use of the UDUNITS-2 C library, it is possible validate such strings.  To make that a possibility, each of the specifications for units will have to be modified such that the example becomes both machine-readable (defined in its own XML element) and the example must be compatible with what UDUNITS-2 will accept (which is very broad).

This change will require an addition to the NeXus XML Schema (``nxdl.xsd``) to allow for the new content.

prjemian:
For example, this existing specification:

    <xs:enumeration value="NX_CROSS_SECTION">
        <xs:annotation>
            <xs:documentation>example: barns</xs:documentation>
        </xs:annotation>
    </xs:enumeration>

could (for example) be changed to:

			<xs:enumeration value="NX_CROSS_SECTION">
				<xs:annotation>
					<xs:documentation>
						cross sectional area of a scattering process
						<xs:element name="example">barn</xs:element>
						<xs:element name="example">m2</xs:element>
					</xs:documentation>
				</xs:annotation>
			</xs:enumeration>

note: The process that builds the manual documentation from these enumerations will need to be revised in consequence to this change.

eugenwintersberger:
Sounds like avery reasonable idea to me.

PR #592, opened 2017-09-09
for #530: adds machine-readable unit strings

prjemian:
The unit categories specified in `nxdlTypes.xsd` have examples when appropriate.  This PR makes the examples machine-readable, as demonstrated by the change in the documentations tool ``utils2rst.py``.  The change will be of benefit to software that will act on the units, such as data file validation software.  Use the UDUNITS-2 library support.  In Python, the cfutils package may be useful.

### DIFF
--- a/nxdlTypes.xsd
+++ b/nxdlTypes.xsd
@@ -87,7 +87,8 @@
 	<xs:simpleType name="NX_ANGLE">
 		<xs:annotation>
 			<xs:documentation>
-				units of angle, example: m
+				units of angle
+				<xs:element name="example">m</xs:element>
 			</xs:documentation>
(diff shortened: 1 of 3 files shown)


zjttoefs:
I like that. Was something we were discussing here on Friday. Found one (old) cut and paste mistake, which is fixed now. I'd be happy to merge. Let's give @eugenwintersberger some time, though.

eugenwintersberger:
Looks good to me.

rayosborn:
I think the reference to cfutils in the pull request means the cfunits package.

prjemian:
As @rayosborn wrote, that's right, this one: https://pypi.python.org/pypi/cfunits

PR #593, opened 2017-09-11
allow NX_NUMBER for arrays with times

zjttoefs:
The exiting use at ISIS is floats. We wanted to allow ints.
NX_NUMBER sounds like a sensible compromise.
Don't think we deliberately picked NX_INT.


zjttoefs:
Something is strange about the current state of the class. I'm quite sure we had a @scaling attribute, which isn't there any more. I'll investigate.

zjttoefs:
Okay, we only introduced the scaling in the NXlog. That was an oversight. I'll add this to this pull request and raise that again at the next telco or code camp.

prjemian:
per telco 2017-11-07

PR #603, opened 2017-12-07
Fix NxTransformations matrix order

matthew-d-jones:
If a transformation depends on another, the one it depends on should be applied first and should therefore be further to the right in the expression for T_f. Therefore, in the NxTransformation documentation this does not look correct to me:

> For a chain of three transformations, where T_1 depends on T_2 and that in turn depends on T_3, the final transformation T_f is
T_f = T_3 T_2 T_1

 Please check I have not misunderstood the existing documentation before merging!

### DIFF
--- a/base_classes/NXtransformations.nxdl.xml
+++ b/base_classes/NXtransformations.nxdl.xml
@@ -2,9 +2,9 @@
 <?xml-stylesheet type="text/xsl" href="nxdlformat.xsl" ?>
 <!--
 # NeXus - Neutron and X-ray Common Data Format
-# 
+#
 # Copyright (C) 2014-2017 NeXus International Advisory Committee (NIAC)
-# 
+#
 # This library is free software; you can redistribute it and/or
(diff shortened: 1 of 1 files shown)


PeterC-DLS:
The depends_on chain is physical not mathematical. So the ordering is different for passive transformations.

prjemian:
@PeterC-DLS: Can you supply a few words to clarify so others are not also unclear?

PeterC-DLS:
The difference between active and passive transformations are the former moves things in a fixed coordinate frame whereas the latter is changing between coordinate frames.

To interpret information from, for example,  a detector (say a pixel location), we are transforming from the detector's frame via each positioner to the laboratory frame fixed to the last positioner that  depends on `"."`.

matthew-d-jones:
@PeterC-DLS Thank you, makes sense. I will create a new PR for just the equation rendering fixes.

matthew-d-jones:
Created #604 for the rendering fixes.

PR #604, opened 2017-12-08
Fix rendering of mathematical expressions for transformations

matthew-d-jones:
Rendering of mathematical expressions for transformations in the NxTransformations class definition is broken, see http://download.nexusformat.org/sphinx/classes/base_classes/NXtransformations.html
There seems to be a problem with multiline `math` blocks, these changes work around the problem by reducing them to a single line.

Before fix:
before_fix

After fix:
after_fix


prjemian:
I approve this.  Built the docs locally, viewed the NXtransformations page.  Looks good.  Thanks.

PR #612, opened 2018-01-24
Document concepts of the new geometry description

matthew-d-jones:
**#601 should be merged before this PR**

Documented concepts of the new shape description (`NXoff_geometry` and `NXcylindrical_geometry`).
See `html/design.html#design-geometry` section for changes.


mkoennecke:
I have looked at this again. To this purpose I merged the other branch with this. There are some rough edges in the writing, for example: 
 
Conversion between OFF files and the NeXus description is straightforward allows geometry to be  defined 

I think there is a period and a few words missing there. 

The next thing I stumbled over is the detector_faces array in NXoff_geometry. It took me a while to understand that l is just an ordering dimensions and that the 2 in the dimensions actually describes a pair of detectorID, faceID. IMHO, this needs to be documented more clearly. Another question: is this array sorted by detectorID? Such that I can expect to find all faces making up a detector volume in a consecutive order. 

And I think the NIAC should vote on this

matthew-d-jones:
Thanks for the feedback @mkoennecke. I'll try to clarify these points.
Would NOBUGS be the next opportunity for a NIAC vote?

mkoennecke:
In principle NOBUGS would be the next point. But we could also vote by email on this. May be this is a topic for the next Telco?

zjttoefs:
I would say we have a sufficient mandate from the code camp to continue down this route. No need to accelerate a vote.
Presenting it at one of the next teclos for information sounds like a good idea, though.

matthew-d-jones:
I have tried to improve the wording in the couple of places @mkoennecke pointed out.

Sorting of the `detector_faces` is a good question. I've specified that it should be in ascending order of `faces` index, which is what my python scripts implement, but perhaps ordering by detector ID would be more useful to readers of the file...

zjttoefs:
I would suggest also presenting the NeXus equivalent to the cube.off file. That would explain the need for the winding order.

manual/source/design.rst
```diff
@@ -849,13 +862,97 @@ This is also a nice example of the application of transformation matrices:
...
+        :width: 15%
+        :align: center
+
+        Cylinder definition with three vertices
+
+
+Detector Shape Descriptions
+---------------------------
+
+An ``NXoff_geometry`` or ``NXcylindrical_geometry`` group named ``shape`` can
```
ACLay: The field name added was `detector_shape`, not `shape`

PR #622, opened 2018-05-10
add choice element to NXDL Schema

### LINKED ISSUE #619: add choice element to NXDL
As part of [adding new geometry definitions to `NXdetector`](https://github.com/nexusformat/definitions/pull/601), it is proposed to extend the NXDL to allow for specification of multiple choices for a named group or field.  

The proposed use seems clear, to create a group with a defined name that is one specific NXDL class from a list of possibilities.  The immediate children of the `<choice>` should all be of the same type (either all `<field>` or all `<group>` elements) and all should have the same `@name` attribute.

prjemian:
fixes #619, pre-requisite for [adding new geometry definitions to `NXdetector`](https://github.com/nexusformat/definitions/pull/601)

If we *only* want choice to apply to groups, this is ready to review and *squash merge**.

If we also want to include fields (and explain why we want that), a bit more work will be needed.

After this is merged, this should be tagged Schema-3.4 (similar to [Schema-3.3](https://github.com/nexusformat/definitions/releases/tag/Schema-3.3))


prjemian:
We *could* relax the requirement on the `@name` attribute for this PR or remove it altogether from the `choice` element.  I don't see how we can enforce that (easily) through the NXDL Schema without defining a special `groupGroup` just for this in the nxdl.xsd file.  Don't want to start that trend.

But, if this PR is good enough for today, these would be changes to be made upon request in the future.  That's my vote.

zjttoefs:
I'd be happy to merge. Let's give @mkoennecke some limited time to review, though.

PR #644, opened 2018-12-14
Fix NXtransformation example

### LINKED ISSUE #643: Incorrect use of NXtransformations in documentation
This example use of `NXtransformations` looks wrong, or at least misleading:
https://github.com/nexusformat/definitions/blob/master/manual/source/examples/euler-cif.txt

I think the correct form should be:
```
sample:NXsample
    transforms:NXtransformations
        rotation_angle
          @transformation_type=rotation
          @vector=0,1,0
          @offset=0,0,0
          @depends_on=.
        chi
          @transformation_type=rotation
          @vector=0,0,1
          @offset=0,0,0
          @depends_on=rotation_angle
```

matthew-d-jones:
Fixes #643 

According to http://download.nexusformat.org/sphinx/classes/base_classes/NXtransformations.html transformation axes should be fields of an NXtransformations group, not direct fields of the component to which the transformations apply. This fixes the example shown on the design page of the manual.

I think the `depends_on` field (not attribute) should be in the parent component rather than the `NXtransformations` group, as this results in one fewer operation required when parsing, but this should also be clarified in the documentation.

### DIFF
--- a/manual/source/examples/euler-cif.txt
+++ b/manual/source/examples/euler-cif.txt
@@ -1,18 +1,19 @@
   sample:NXsample
-    rotation_angle
-      @transformation_type=rotation
-      @vector=0,1,0
-      @offset=0,0,0
-      @depends_on=.
-    chi
-      @transformation_type=rotation
-      @vector=0,0,1
-      @offset=0,0,0
-      @depends_on=rotation_angle
-    phi
-      @transformation_type=rotation
-      @vector=0,1,0
-      @offset=0,0,0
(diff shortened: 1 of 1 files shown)


ACLay:
Assuming it follows the same rules as the `depends_on` attributes of the transforms (http://download.nexusformat.org/doc/html/design.html#coordinate-transformation-field-and-attributes), should the `depends_on` attribute of the sample be updated to `transforms/phi`?

prjemian:
Ready to merge?

PR #649, opened 2019-03-19
Changes to fields and description of NXdisk_chopper

matthew-d-jones:
~Since the existing `NXdisk_chopper` definitions appeared not to be fit for purpose I have been bold in making breaking changes.~ edit: changes are now non-breaking

Changes follow discussion of requirements for recording chopper data at the ESS.

I think the `delay` description requires further refinement.

### DIFF
--- a/base_classes/NXdisk_chopper.nxdl.xml
+++ b/base_classes/NXdisk_chopper.nxdl.xml
@@ -64,7 +64,8 @@
 		<doc>Number of slits</doc>
 	</field>
 	<field name="slit_angle" type="NX_FLOAT" units="NX_ANGLE">
-		<doc>Angular opening</doc>
+		<doc>Angular opening of each slit</doc>
+		<dimensions><dim index="1" value="n"/></dimensions>
 	</field>
 	<field name="pair_separation" type="NX_FLOAT" units="NX_LENGTH">
 		<doc>Disk spacing in direction of beam</doc>
@@ -93,7 +94,8 @@
 		<doc>Radius of the disk</doc>
 	</field>
 	<field name="slit_height" type="NX_FLOAT" units="NX_LENGTH">
-		<doc>Total slit height</doc>
+		<doc>Total height of each slit</doc>
+		<dimensions><dim index="1" value="n"/></dimensions>
 	</field>
 	<field name="phase" type="NX_FLOAT" units="NX_ANGLE">
 		<doc>Chopper phase angle</doc>
--- a/base_classes/NXdisk_chopper.nxdl.xml
+++ b/base_classes/NXdisk_chopper.nxdl.xml
@@ -28,9 +28,9 @@
 	type="group" extends="NXobject">
 
 	<symbols>
-    <doc>This symbol will be used below to coordinate datasets with the same shape.</doc>
-    <symbol name="n"><doc>Number of slits in the disk</doc></symbol>
-  </symbols>
+		<doc>This symbol will be used below to coordinate datasets with the same shape.</doc>
+		<symbol name="n"><doc>Number of slits in the disk</doc></symbol>
+	</symbols>
 
 	<doc>
 		A device blocking the beam in a temporal periodic pattern.
@@ -106,10 +106,10 @@
 		</doc>
 	</field>
 	<field name="ratio" type="NX_INT">
-	  <doc>
-      Pulse reduction factor of this chopper in relation to other
-      choppers/fastest pulse in the instrument
-	  </doc>
+		<doc>
+			Pulse reduction factor of this chopper in relation to other
+			choppers/fastest pulse in the instrument
+		</doc>
 	</field>
 	<field name="distance" type="NX_FLOAT" units="NX_LENGTH">
 		<doc>Effective distance to the origin</doc>
--- a/base_classes/NXdisk_chopper.nxdl.xml
+++ b/base_classes/NXdisk_chopper.nxdl.xml
@@ -2,9 +2,9 @@
 <?xml-stylesheet type="text/xsl" href="nxdlformat.xsl" ?>
 <!--
 # NeXus - Neutron and X-ray Common Data Format
-# 
+#
 # Copyright (C) 2008-2018 NeXus International Advisory Committee (NIAC)
-# 
+#
 # This library is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
 # License as published by the Free Software Foundation; either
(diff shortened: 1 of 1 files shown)


matthew-d-jones:
@zjttoefs I've created this as a draft PR for you to take a look at first. The `phase` description still makes little sense to me.

zjttoefs:
Is there a need to make the changes breaking?

I don't think we need to remove anything. And the phase in time might better be called _delay_. No?

base_classes/NXdisk_chopper.nxdl.xml
```diff
@@ -24,56 +24,78 @@
...
+		<dimensions><dim index="1" value="2n"/></dimensions>
 	</field>
-	<field name="pair_separation" type="NX_FLOAT" units="NX_LENGTH">
-		<doc>disc spacing in direction of beam</doc>
+	<field name="top_dead_center" type="NX_NUMBER" units="NX_TIME">
+		<doc>
+			Timestamps of the top-dead-center signal
+		</doc>
+	</field>
+	<field name="beam_position" type="NX_FLOAT" units="NX_ANGLE">
```
zjttoefs: I don't think we need this if we record the slit pattern starting from the TDC angle.

matthew-d-jones:
The other breaking changes were removing the `type` and `wavelength_range` fields which don't make sense for a single disk, and changing the meaning of the radius field which made little sense to me.

base_classes/NXdisk_chopper.nxdl.xml
matthew-d-jones: That depends whether the TDC position means the "position of the TDC sensor" or the "position of the beam when the disk is in the position where the TDC timestamp is recorded".

matthew-d-jones: Maybe the latter is more intuitive, just clunkier to describe.

matthew-d-jones:
Just realised I carefully used American spelling everywhere to be consistent with "disk" in "NXdisk_chopper" but then used the term "anticlockwise".

base_classes/NXdisk_chopper.nxdl.xml
prjemian: +1 it's an international standard, TDC is jargon (known in automotive)

prjemian:
Re [wavelength_range](https://github.com/nexusformat/definitions/pull/649#issuecomment-474414023), isn't there always a range of wavelengths that depends on the chopper opening size and the rotation speed?

matthew-d-jones:
> Re [wavelength_range](https://github.com/nexusformat/definitions/pull/649#issuecomment-474414023), isn't there always a range of wavelengths that depends on the chopper opening size and the rotation speed?

I suppose there is technically a low bound on the energy due to the thickness of the disk, but otherwise not really with a single disk.

mkoennecke:
There is a contradiction here: in the top documentation it says that there ought to be a NXdisk_chopper group per disk. Then the pair_separation does not make any sense as it implies a chopper pair. 

Regarding the phase. Given there is a distance between two choppers. Then the second chopper has to open a little later then the first one in order to account for the flight time between the two choppers. This is expressed as an angular offset, the phase, in the chopper systems I had to deal with so far. This is something  we read from the chopper controller, thus I want to have this field in order to record it. I do not mind the additional fields, they serve to describe the chopper in a more accurate way.

matthew-d-jones:
@mkoennecke 
There is already an angular `phase` field.

I originally removed the fields which don't make sense for a single disk, but Tobias is keen not to make breaking changes at the same time.
Given that there is nowhere else to store potentially useful information like `wavelength_range`, perhaps it is not unreasonable to store it in one, or both, of the two separate `NXdisk_chopper`s making up the pair? Similarly, specifying `type` of "synchro_pair" could imply that this disk is one of those in the pair. We would need to change the wording to make this clear.

prjemian:
Ready to merge?

zjttoefs:
I have given my approval. If no one else comments with requests for changes in the next days, I would merge.

prjemian:
Ok, then.  If no additional comments by May 22, then I will merge this.

mkoennecke:
I am mostly OK with this. But I have a minor point: There is this slits field which indicates that a chopper may have more than one slit (They usually do). Should then slit_angle and slit_edges not be arrays of length nSlits? This might apply to the radius etc. fields too.

matthew-d-jones:
> I am mostly OK with this. But I have a minor point: There is this slits field which indicates that a chopper may have more than one slit (They usually do). Should then slit_angle and slit_edges not be arrays of length nSlits? This might apply to the radius etc. fields too.

slit_edges is already described as an array of length 2*nSlits:
https://github.com/nexusformat/definitions/pull/649/files#diff-965d206bd4d315d80b5f08de5a97d475R79

but I forgot to add the array description for `slit_angle` and `slit_height`, thanks for catching that.

zjttoefs:
I don't like the new change. It changes existing fields and hence breaks existing files.

The height is rarely different for different cut outs, I believe. And changing the slit angle to an array makes the slit_edges redundant.

prjemian:
Seems this is not yet ready to merge.

base_classes/NXdisk_chopper.nxdl.xml
```diff
@@ -24,13 +24,26 @@
...
+	<symbols>
+    <doc>This symbol will be used below to coordinate datasets with the same shape.</doc>
+    <symbol name="n"><doc>Number of slits in the disk</doc></symbol>
+  </symbols>
+
 	<doc>
-		A device blocking the beam in a temporal periodic pattern.
-		
-		TODO: need documentation
+		A disk which blocks the beam but has one or more slits to periodically
```
prjemian: First line of this documentation should be a single-line summary (to be used in creating this summary: http://download.nexusformat.org/doc/html/classes/base_classes/index.html#base-class-definitions).  A blank line afterwards, then any multi-line text.

zjttoefs:
Since @mkoennecke  framed his suggestions yesterday as "minor comments" I would think we're okay to merge after reverting the last commit and addressing the documentation improvement that @prjemian has highlighted.

base_classes/NXdisk_chopper.nxdl.xml
```diff
@@ -42,37 +57,65 @@
 		</enumeration>
 	</field>
 	<field name="rotation_speed" type="NX_FLOAT" units="NX_FREQUENCY">
-		<doc>chopper rotation speed</doc>
+		<doc>
+			Chopper rotation speed. Positive for anticlockwise rotation when
+			facing the sample, negative otherwise.
```
zjttoefs: I would make this "facing away from the source". Removes confusion if you ever put a chopper after the sample.

base_classes/NXdisk_chopper.nxdl.xml
```diff
@@ -42,37 +57,65 @@
...
+	<field name="slit_edges" type="NX_FLOAT" units="NX_ANGLE">
+		<doc>
+			Angle of the each edge of each slit from the position of the
+			top-dead-center timestamp sensor, anticlockwise when facing the sample.
+			The first edge must be the opening edge of a slit, thus the last edge
+			may have an angle greater than 360 degrees.
+		</doc>
+		<dimensions><dim index="1" value="2n"/></dimensions>
+	</field>
+	<field name="top_dead_center" type="NX_NUMBER" units="NX_TIME">
```
zjttoefs: We should allow a start_time attribute (or whatever is consistent with NXlog).

base_classes/NXdisk_chopper.nxdl.xml
matthew-d-jones: :+1: done

base_classes/NXdisk_chopper.nxdl.xml
matthew-d-jones: :+1: done

zjttoefs:
Since we have three green ticks from reviewers, I would almost feel comfortable to merge. Especially since there are no changes (let alone breaking ones) to existing definitions.
But I'll give it a day or so, if @prjemian or @mkoennecke want to reassess the PR.

base_classes/NXdisk_chopper.nxdl.xml
```diff
@@ -24,56 +24,76 @@
...
+			facing the sample, negative otherwise.
+		</doc>
 	</field>
 	<field name="slits" type="NX_INT">
 		<doc>Number of slits</doc>
 	</field>
 	<field name="slit_angle" type="NX_FLOAT" units="NX_ANGLE">
-		<doc>angular opening</doc>
+		<doc>
+			Angle of the each edge of each slit from the position of the
```
prjemian: change `the each` to `each` and I'm good overall

zjttoefs: To cut this short, while I can't edit @matthew-d-jones source for the PR, I'll merge and drop the "the" in master.

zjttoefs:
Brilliantly referenced the wrong pull request, but the deed is done in 935afe4bd3ee0bdd458cd9cfa61bed731ae62723

PR #652, opened 2019-03-27
Clarify data_origin and data_size.  Resolves #650.

### LINKED ISSUE #650: data_origin and data_size documentation is unclear
Hi all, there are two issues here:

1. NXdetector_module (base class) and NXmx (application definition that cites NXdetector_module) have different documentation for data_origin and data_size.  The former states these are 2 values for the origin and size, but the latter specifies 2 or 3 values
2. Assume the NXmx doc string is correct.  data_origin specifies the order the dimensions are listed (slow to fast), but data_size does not.  Presumably it also should be slow to fast.

phyy-nx:


phyy-nx:
Ok to merge?  If no response, will merge in an hour :)

prjemian:
Will 24 hour notice be too long?

phyy-nx:
Nope that'll be fine. I'll wait until 9am tomorrow, PST (24 hours from
now).  Thanks,
-Aaron

prjemian:
Reading this fresh, `dimension-2` and `dimension-3` are still not so clear.  Perhaps it is that I'm used to seeing the terms *2D* and *3D* and this is reversed syntax from what I have seen.  Refers to arrays with ranks 2 & 3, respectively.  Will the referenced `data_origin` & `data_size` fields be rank 1 arrays with the same length as the rank of the image array? (actually one less than the rank of the image array due to the additional frame index)

Could be clearer still.  Not ready to merge.

phyy-nx:
How about this?  I've retained the dimension-2/dimension-3 nomenclature since it's the same pharasing as used for the data field, but I've been more explicit as to what is happening here.

prjemian:
I expect the term *rank* to be used as a scalar value that describes the number of indices of an array.  That puzzles me when it is used thus:  *dataset of rank (np, i, j)* yet I understand from the context what is meant.

Which one of these is correct? I assume this one:

			The frame number dimension (``np``) is omitted.  Thus the
			``data_origin`` field for a dimension-2 dataset with indices ``(np, i, j)``
			will be an array with indices ``(i, j)``, and for a dimension-3
			dataset with indices ``(np, i, j, k)`` will be an array with indices ``(i, j, k)``.

but the existing wording might be interpreted as describing this one:

			The frame number dimension (``np``) is omitted.  Thus the
			``data_origin`` field for a dimension-2 dataset with indices ``(np, i, j)``
			will be a dimension-1 array with values ``(i, j)``, and for a dimension-3
			dataset with indices ``(np, i, j, k)`` will be a dimension-1 array with values ``(i, j, k)``.

phyy-nx:
Fair points.  I've updated using this wording.

prjemian:
I agree this is clearer.

prjemian:
Give it a day before merge, allowing for those out of our at-work time zone to have a chance at review.

phyy-nx:
👍

PR #654, opened 2019-03-28
Allow multi-image masks. Resolves #651.

### LINKED ISSUE #651: Need per-image pixel mask
The NXmx specification of pixel_mask is:

`pixel_mask[i, j, k]: (optional) NX_INT`

This specifies a single mask applicable to all images in the dataset.  This is a valid use case.  I believe it could also be defined as:

`pixel_mask[np, i, j, k]: (optional) NX_INT`

i.e., it is valid for the user to provide a separate mask for each image.  This is the case for the EuXFEL.

Therefore, the pixel mask always has a number of dimensions equal to the dimensionality of the data OR the dimensionality of the data -1, where the former is for a mask for each image and the later is a mask for the whole dataset.

Thoughts?

phyy-nx:
Putative solution for multi-image masks and a proposed solution to the issues raised here: https://github.com/HDRMX/NXmx/wiki/Treatment-of-(Dynamic)-Shadows-in-Eiger-Data-Sets

Adds this text: Can be either one mask for the whole dataset (rank i, j, k) or each frame can have its own mask (rank np, i, j, k).

Also sync some of the documentation between NXdetector and NXmx.

### DIFF
--- a/applications/NXmx.nxdl.xml
+++ b/applications/NXmx.nxdl.xml
@@ -375,7 +375,9 @@
 
 					<field name="pixel_mask" type="NX_INT" minOccurs="0">
 						<doc>
-							The 32-bit pixel mask for the detector.
+							The 32-bit pixel mask for the detector. Can be either one mask
+							for the whole dataset (rank i, j, k) or each frame can have its own
+							mask (rank np, i, j, k).
 							Contains a bit field
 							for each pixel to signal dead,
 							blind or high or otherwise unwanted
--- a/base_classes/NXdetector.nxdl.xml
+++ b/base_classes/NXdetector.nxdl.xml
@@ -587,7 +587,10 @@
   </field>
   <field name="pixel_mask" type="NX_INT" >
     <doc>
-      The 32-bit pixel mask for the detector.
+      The 32-bit pixel mask for the detector. Can be either one mask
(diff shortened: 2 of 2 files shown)


phyy-nx:
@graeme-winter has argued we want a separate dynamic_mask. Closing this for now, but it can be reopened pending discussion.

zjttoefs:
This pull request only clarifies the status quo. I see no reason to not merge this.

phyy-nx:
Hey, so this would be independent then of whether #655 is merged in?  If so, I'll merge in an hour and we can let the issue of whether to add a dynamic mask have some more discussion, if needed.

prjemian:
[Again](https://github.com/nexusformat/definitions/pull/652#issuecomment-479068075), a day for final review is recommended.

phyy-nx:
👍

graeme-winter:
So, in people words you are allowing the dimensionality of pixel_mask to be extended to the same dimensionality of the data set, then the other PR is to add what is effectively a second mask? Makes sense to me.

phyy-nx:
Yep. I don’t see why we can’t have both, if there is a good use case like this. 

-Aaron

phyy-nx:
Thanks all.

PR #655, opened 2019-03-29
Add per-image dynamic_mask as option for NXdetector. Resolves #651.

### LINKED ISSUE #651: Need per-image pixel mask

phyy-nx:
Also sync documentation between NXmx and NXdetector.

Alternate solution to #651 than what I did in #654.  Pending discussion.

### DIFF
--- a/applications/NXmx.nxdl.xml
+++ b/applications/NXmx.nxdl.xml
@@ -382,8 +382,9 @@
 					<field name="pixel_mask" type="NX_INT" minOccurs="0">
 						<doc>
 							The 32-bit pixel mask for the detector. Can be either one mask
-							for the whole dataset (rank i, j, k) or each frame can have its own
-							mask (rank np, i, j, k).
+							for the whole dataset (i.e. an array with indices i, j) or
+							each frame can have its own mask (in which case it would be
+							an array with indices np, i, j).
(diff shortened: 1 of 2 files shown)


phyy-nx:
Updated branch taking into account #654.

phyy-nx:
Updates from telco applied

prjemian:
@yayahjb : Can you review?

prjemian:
@phyy-nx:  This PR is languishing.  Should we merge without review from @yayahb?

phyy-nx:
Hi, I've rebased against master.  Unless there are no objections, I'm happy to merge.  I'll merge by Wednesday, end of day PST?  Thanks!

Issue #645, opened 2019-01-11
Document variant decision


mkoennecke:
Hi,

following up on a recent question posed to me I found out that at NIAC 2014 we decided on a scheme 
for encoding variants of a value in NeXus. The scheme looks like this(I believe):

distance
    @variant=distance_from_reading
distance_from_reading
    @variant=distance_as_guessed_by_cleaner
distance_as_guessed_by_cleaner

I.e. a value for which a better value becomes available gets a variant attribute pointing to improved 
value. The name of which is built from the original name plus a description appended with underscores. The whole can become a chain of some length.

I had to dig this information up from Telco minutes. The decision ism actually in the NIAC2014_Minutes.md but without details. But this is not in the manual.

Thus I suggest:
- Please someone check that this is correct: (Herbert: it came from the concordance discussions with CIF)
- Then put this in the manual, probably in the section describing storage of individual data items in the file.

yayahjb:
Variants in CBF are a little different, since we are dealing with the
tables of a relational database instead of a tree, so in that case, variant
is
a column in any table and the name of the table does not have to change, a
variant of a row is just another row.  The scheme Mark proposes is a
natural way to accomplish a similar result without having relations.

Regards,
    Herbert

rayosborn:
I don't recall that discussion, although I assume I was there. I am concerned about the idea that the variant is the improved value, because it would mean that the value stored in the NXDL-approved field is not the best one to use, even though it is the value that will be used by any analysis code. The value in a standard field like 'distance' should, in my view, always be the value that analysis software is expected to use. It also seems an impractical naming scheme. Trying to add what is effectively additional metadata to the name seems like a bad idea. I hope I didn't vote for it, but I agree that we shouldn't just ignore previous votes.

rayosborn:
It's the last point I find worrying. Someone writing a third-party software package, not knowing anything about a particular experiment, is going to use the NXDL-approved field. If we discover, after the experiment, that we got the distance wrong, we have to change that value, not create another undocumented field to store it. I'm not against storing variants even if I don't like the naming scheme; I'm just against stating that they contain "improved" values.

vasole:
[...]

PR #666, opened 2019-05-23
Added documentation for variants to datarules.rst

mkoennecke:
Refs #645

As discussed, documentation for variants as recovered from NIAC Telcos and based on a decision from NIAC 2014.


prjemian:
Will merge after August 31 if there are no objections

PR #669, opened 2019-07-22
NXmx: add NXsource and make name in NXsource and NXinstrument required

phyy-nx:
From HDRMX meeting at ACA 2019. 

Co-authored-by: Graeme Winter <graeme.winter@gmail.com>

### DIFF
--- a/applications/NXmx.nxdl.xml
+++ b/applications/NXmx.nxdl.xml
@@ -74,6 +74,13 @@
 			</field>
 
 			<group type="NXinstrument">
+			        <field name="name" minOccurs="1">
+                                       <doc>Name of instrument</doc>
+                                       <attribute name="short_name">
+                                               <doc>short name for instrument, perhaps the acronym</doc>
+                                       </attribute>
+                                </field>
+
 				<group type="NXattenuator" minOccurs="0">
 					<field name="attenuator_transmission" type="NX_NUMBER" units="NX_UNITLESS"
 						minOccurs="0" />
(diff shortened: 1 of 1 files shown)


prjemian:
@phyy-nx: This PR is languishing. Should we merge without review from @yayahb?

yayahjb:
Dear Pete,
  We had an HDRMX meeting in Covington and Aaron Brewster and Graeme Winter
were delegated to compose the necessary agreed to changes for NXmx to
implement the consensus of the meeting, which has strong MX community
support.  This is part of that consensus.  I am happy with the idea of the
name field for NXsource and for NXinstrument being mandatory in NXmx.  The
incomplete part has been whether the shortname attribute should or should
not also be mandatory in NXmx.  Personally I think it should with a default
value equal to the value of the name field, but I would not like to hold up
the agreement on name being mandatory while people consider that minor side
issue.
  Regards,
    Herbert

prjemian:
@yayahb: Thanks!

mkoennecke:
I agree with Herbert that making the names of the instrument and the source mandatory makes sense. 

I have issues with the short_name however. First of all, that is a new attribute, IMHO, not seen before. Then  which name do you really want? For example HRPT at SINQ can be described as:
- High Resolution Powder diffracTometer 
- HRPT
The first is the long name. Next to no one knows that. In the case of HRPT it was a byproduct of inventing the acronym. Moreover the combination HRPT together with the source name SINQ fully identifies both instrument and source. 

Thus I think the short name for both source and instrument should be the content of the name fields. The extended name (which may not even exist for some instruments) could be an attribute long_name. Though I do not see that it adds value except for understanding convoluted language use in instrument naming.

yayahjb:
The short name is already an attribute the name both for NXsource and for
NXinstrument.

*                      name*: (optional) NX_CHAR
<http://download.nexusformat.org/doc/html/nxdl-types.html#nx-char>

               Name of source

*              @short_name*: (optional) NX_CHAR
<http://download.nexusformat.org/doc/html/nxdl-types.html#nx-char>

          short name for source, perhaps the acronym

*name*: (optional) NX_CHAR
<http://download.nexusformat.org/sphinx/nxdl-types.html#nx-char>

      Name of instrument

*@short_name*: (optional) NX_CHAR
<http://download.nexusformat.org/sphinx/nxdl-types.html#nx-char>

short name for instrument, perhaps the acronym


  Each facility should call itself whatever it wants to call itself,  If
HRPT want to call itself HRPT or call itself High Resolution Powder
diffracTometer or call itself something else, it should not be up to us to
decide.  "name" means "name", the identification of the instrument or the
source that the proprieters of that instrument or source chose, whether it
be short, long or something in between.  That is why I suggested the the
short name default to the same string as the name.  If the proprieters do
want to speciify some other short name they should be free to do so, even
if in some cases it turns out to be longer than the name.  They can explain
why or smile knowingly or just do it, not us.

Please, I believe in standards, but not in pointless policing of harmless
variations in their use.  We really do need for people to put a few
essential names in their metadata.  The less we say about what they should
choose for their own names, the better.

mkoennecke:
Alright then, I overlooked the short_name in NXsource. No more objections

yayahjb:
If there are no further objections, I would suggest this pull request be
executed.  -- Herbert

phyy-nx:
Thanks folks.  My intent had been to go over this in the next Telco, but if there are no further objections then maybe just merging it is fine.  I'll merge it Wednesday, end of day PST?

Issue #524, opened 2017-01-03
documentation: make it obvious what is required and what is optional


prjemian:
community feedback received by email about one of the application definitions:

-- *You know, that manual really needs to make it obvious what is required and what is optional.*

This is the table in the manual that describes the defaults:
http://download.nexusformat.org/doc/html/defs_intro.html#minoccurs

prjemian:
I suggest tagging fields in application definitions based on the minOccurs attribute, such as the ``title`` field in NXcanSAS (which relies on the default minOccurs, which is `1` for application definitions):

* *(required)* : `minOccurs=1`
* *(optional)* : `minOccurs=0`

We could also tag fields where the name is flexible.
Beginning to sound like too much information but this question of what is optional v. required is quite common.

prjemian:
For example, in any application definition, this line for a field with `minOccurs=1`:

**definition**: [NX_CHAR](http://download.nexusformat.org/doc/html/nxdl-types.html#nx-char)

will change to

**definition**: [NX_CHAR](http://download.nexusformat.org/doc/html/nxdl-types.html#nx-char) (required)


Fields in a base class with `minOccurs=0`, such as this line in NXuser

**name**: [NX_CHAR](http://download.nexusformat.org/doc/html/nxdl-types.html#nx-char)

will change to

**name**: [NX_CHAR](http://download.nexusformat.org/doc/html/nxdl-types.html#nx-char) (optional)

prjemian:
must be applied to fields, groups, and attributes

prjemian:
With this change, it will be more obvious in the documentation what each NXDL has to say about minOccurs for each item.

Before merging this branch, it may be necessary to revise various NXDL files with respect to `minOccurs`.

prjemian:
Note that for NXDL *attribute* elements, the attribute is called **optional** (not ``minOccurs``).

prjemian:
Seems that waiting for [revise various NXDL files with respect to `minOccurs`](https://github.com/nexusformat/definitions/issues/524#issuecomment-271028265) might take some time.  Is that waiting necessary?  By proceeding with the merge, any NXDL can assess the urgency of such revision.  At this time, it is not obvious which NXDL files must be revised.

prjemian:
fixed by merge #527

PR #671, opened 2019-07-28
clarify naming convention

### LINKED ISSUE #544: clarify how to name fields and groups
At the 2017-03-08 [web conference](http://www.nexusformat.org/Telco_20170308.html), the recommendations and rules for naming fields and groups were discussed.  The manual says one thing in the text and another in the regular expression for the recommended practice.

Specifically, the question is do we recommend (or even allow) the use of an *underscore* character (`_`) as the first character of a field or group name?  The text does not describe this while the regular expression, added later, does.  It is clear the recommendation is to use lower case letters.  

The text does not state explicitly that field or group names must not begin with a number.  This should be stated clearly.

There is a second regular expression with accompanying text, that adds uppercase letters to the list of allowed starting characters in a field or group name.  There were no concerns raised during the meeting about this.

prjemian:
fixes #544


prjemian:
@vasole: Here's a snapshot of what the documentation looks like when it is built:

nexus

vasole:
I certainly preferred the interpretation of NeXus being attribute based leaving freedom to the names. 
This PR now makes clear the intention. It implies a huge change in the understanding of what it has been in the documentation for long time. If the NIAC is clear that should be a rule and not a recommendation then somebody else can approve it.

To me this whole thing of changing a recommendation to  a rule is a hidden attempt to workaround the limitations imposed by trying to access HDF5 fields not as keys in a dictionary containing valid paths in a file system but as attributes of some instance.

This rule is like forbidding to use file names in a file system if they contain a dot or they starting by a number. Therefore, I would prefer somebody else to approve this PR that in my opinion deserves discussion at highest level.

prjemian:
To be clear, this PR is a change in the documentation only.  The intent of the naming convention has always been that the names should be valid program variable names, dating to the original design of NeXus using HDF4.

I've add @mkoennecke as a reviewer since he is one of the founders.

mkoennecke:
If I remember correctly, this whole naming convention came into being in order to make our lives easier. Names such "*.-' will cause all sorts of problems when programming. May be, the whitespace language  can work with that. But as far as I know there is no HDF5 interface to whitespace.... 

The . (period) has special meaning in HDF-5 path descriptions and is therefore best avoided. 

Not using upper case has been our practice for a long time. Thus, I believe we make no error when we forbid it.

Also, I do not really understand Armandos argument:
>To me this whole thing of changing a recommendation to a rule is a hidden attempt to workaround >the limitations imposed by trying to access HDF5 fields not as keys in a dictionary containing >valid paths in a file system but as attributes of some instance.

Please elucidate.

rayosborn:
I don't have strong opinions on whether these should be rules or strong recommendations, but to answer @mkoennecke, HDF5 (or at least h5py, with which I am more familiar) does allow white spaces, periods, and, as far as I can tell, any other unicode characters in object names. As @vasole points out, this is not a problem when the object names are treated as dictionary keys, but doesn't allow them to be treated as attributes. NeXpy allows both when possible, because it's more convenient to type, but if the field name violates Python naming conventions, then you have to use the dictionary form, which is recommended for use in Python scripts.

rayosborn:
Just as an example, I was able to create the following tree using the [nexusformat](https://github.com/nexpy/nexusformat) package:
```
>>> w1['entry/anothér.variable']='something else'
>>> print(w1.tree)
w1:NXroot
  @HDF5_Version = '1.10.4'
  @file_name = '/Users/rosborn/Desktop/w1.nxs'
  @file_time = '2019-07-29T09:55:32.612858'
  @h5py_version = '2.9.0'
  @nexusformat_version = '0.4.20+3.g8c1c9c9.dirty'
  entry:NXentry
    a variable = 'entry'
    another.variable = 'something'
    anothér.variable = 'something else'

w1['entry/anothér.variable']
>>> NXfield('something else')
```
So the naming restrictions are probably a little outdated now that unicode is in such widespread use, but it would certainly be more convenient for me if they were preserved, since I prefer typing in the attribute form.

vasole:
@mkoennecke 

Elucidation: today the whole point of not allowing a dot at all or a number as starting character of a field is to allow to use things like nexusformat accessing the fields as attributes of some object instead of using the correct way: accessing them by the path.

As Ray points out, if you want to make sure you do not run into trouble, you should avoid doing so in scripts to avoid collisions between file names accessed as attributes and actual class methods or actual attributes. That makes the interest of the whole approach doubtful.

For instance, keys, attrs and value are perfectly valid NeXus names that could potentially give troubles trying to access them as attributes as they could interfere with actual h5py methods.

vasole:
comment updated (I wrote file names instead of field names)

rayosborn:
I should probably clarify my position about attribute names. NeXpy allows the attribute form, `root.entry.data.x', purely as a convenience for use in interactive sessions. I am a strong believer in reducing the friction a scientist has when 'playing' with their data - trying things out quickly to see what works and what doesn't. If I have to use the dictionary form, `root['entry/data/x']`, all the time, that would slow me down and discourage exploring the data. That's how I work, and the attribute form, particularly with IPython autocompletion, makes it quick and convenient. If I discover that the attribute conflicts with a class method, then it's not a big deal because it generates an exception and I just retype the object reference in dictionary form.

Of course, it's different if you are writing robust code. Then you should always use the dictionary form if the field name is arbitrary to avoid the kind of conflicts @vasole describes. The fact that the `nexusformat` package allows both probably violates computational science best practices, but if it bothers anyone, they probably shouldn't be using Python.

Resisting the use of unicode characters in NeXus object names might end up being a problem in the long term, but if it's a problem for people writing code in other languages, I don't object to keeping the restrictions for now.

vasole:
I would like to clarify that what bothers me about being a rule and not a recommendation are

- the prevention of the use of the dot
- the prohibition to name a field starting by a number

vasole:
As a reminder, for the ipython interactive users, the automatic completion is also available when using dictionaries. 

From the h5py documentation (http://docs.h5py.org/en/stable/config.html):

> The easiest way to enable the custom completer is to do the following in an IPython session:
> 
> In  [1]: import h5py
> 
> In [2]: h5py.enable_ipython_completer()
> 
> It is also possible to configure IPython to enable the completer every time you start a new session. For >=ipython-0.11, “h5py.ipy_completer” just needs to be added to the list of extensions in your ipython config file, for example ~/.config/ipython/profile_default/ipython_config.py (if this file does not exist, you can create it by invoking ipython profile create):

mkoennecke:
My 2c's worth:

I believe (but it is not in the manual) that the NIAC in some stage decided to use only english names for groups and fields. This avoids the unicode discussion. UTF8 content, for example in a text field like title, is OK.  

I was not aware of this python dot notation. Would this not be thrown off when a dot is part of the name?

Now I am entering rant mode: I have a deep dislike for special characters, spaces and such in names. This is a disease inherited from windows. Causes all sorts of problems with quoting, counting spaces and what not, especially when working on the command line. Please outvote me on this, if you like. 

Names starting with a number would work mostly, but are not allowed in some programming languages. Thus I always tend to use thing01, thin02, thing03 when I have multiple things. And cannot work with a list. But this is a matter of taste.

But I am afraid, deciding these things requires a proper vote by the NIAC in order to clarify majorities.

vasole:
@mkoennecke 

Variable names starting by a number are not allowed almost anywhere. Names started by a number are allowed everywhere.

If the documentation upgrades from recommendation to rule, it prevents things like having things like

```
entry_01
   1_background@NXprocess
       ...
   2_crop@NXprocess
       ...
   3_binning@NXprocess
       ...
```

rayosborn:
I'm sure I voted in favour of the more restricted name sets, although that was before Unicode became ubiquitous. I think we will need to discuss this at the next NIAC meeting before allowing a major relaxation of the rules, since there may be other side effects that we haven't thought of. In the meantime,`nexusformat` can read files with any object names allowed by HDF5.

prjemian:
Again, this PR is only a change in the documentation (#544) to clarify the NeXus naming convention for groups and fields, not a change in the NeXus rules.

A major point of this discussion focuses on this [specific text](http://download.nexusformat.org/doc/html/datarules.html#index-2) in the manual, mostly on the one word *recommended*:

> It is recommended that all group and field names contain only these characters:

The context is highlighted in this screen shot of the section:
naming


It is obvious from the intensity of the discussion of this PR that the manual has not clearly expressed the NeXus standard.  Some have interpreted the word `recommended` to mean that other naming choices (such as starting with a number, including a period or whitespace) are compliant with the NeXus standard.  To quote [feedback](https://github.com/nexusformat/definitions/issues/524#issue-198528257) received in a separate topic:

> You know, that manual really needs to make it obvious what is required and what is optional.

We need help, gentle yet insistent help, to make this manual better.

prjemian:
It is still a problem that we present two regular expressions.  Since both describe group and field names that are acceptable, providing two expressions is obviously a source of confusion.  

Looking at the various files contributed to our [examples repository](https://github.com/nexusformat/exampledata), restriction to lower case names is used by most of the example files but not by all of the NeXus community.  In my view, restriction from using upper case characters in group and field provides no value to the NeXus standard.

I propose we only state the one that allows both upper and lower case characters: `[A-Za-z_][\w]*`.

prjemian:
2020 Code Camp can discuss but NIAC should review this before resolving.

prjemian:
The first RE should apply to official NeXus field or group names.  Whatever is decided should become the second RE.

The doc could be updated to say that *if you do use the second RE, you might have more compatibility issues*.

prjemian:
* allow letter, number, underline, or dot
* no dot at end or beginning

vasole:
Is starting by an underscore allowed? If so, the regular expression `^\w+(?:[\w.]*\w)?$` fails for  the case `_`

prjemian:
This regexp

    ^[a-zA-Z0-9_](?:[\w.]*\w)*$

will match all the valid names but it also matches `_`.  It does not match any other invalid name.

vasole:
So, both have the same single point of failure. The single underscore case.

prjemian:
Right.  We need an OR expression to eliminate that one case.

axelboc:
I believe this regex should [do the trick](https://regex101.com/r/5Q7b6C/7):

```js
/^(?:[a-zA-Z0-9]|\w(?:[\w.]*\w))$/
```

prjemian:
Thanks! It matches two underlines (`__`), otherwise good.

prjemian:
This [search](https://regex101.com/r/5Q7b6C/11) has more test cases:

```
# valid names
_123
_123.123
_123.abc
_abc.123
_abc.abc
_valid
_Valid
_valid123
1
1.1
123
a
a____a
a___a
a__a
a.a
A.A
Valid
valid_
valid_name
valid.name
valid123
validname
# invalid names
_
__
___
____
_.
.
._
.1
.123
.a
.A
.no_starting_period
1.
a.
A.
invalid name
invalid-name
invalid,name
invalid;name
invalid:name
invalid!name
invalid?name
invalid'name
invalid"name
invalid(name
invalid)name
invalid@name
invalid/name
invalid\name
invalid&name
invalid#name
invalid%name
invalid`name
invalid^name
invalid^name
invalid+name
invalid<name
invalid=name
invalid>name
invalid|name
invalid~name
invalid$name
no_ending_period.
```

axelboc:
I find it strange to allow underscores everywhere but not underscores only. Sure people shouldn't do it, but is there a reason to forbid it entirely?

Anyways, [it can be done](https://regex101.com/r/5Q7b6C/12) with a negative lookahead:

```js
/^(?!_+$)\w(?:[\w.]*\w)?$/
```

PeterC-DLS:
Note the meaning of ``\w`` is not universal. See https://www.regular-expressions.info/shorthand.html

vasole:
Well, expanding the expression of @axelboc to `^(?!_+$)[A-Za-z0-9_](?:[A-Za-z0-9_.]*[A-Za-z0-9_])?$` seems to work:

https://regex101.com/r/Yknm4v/1

prjemian:
Here is how these revisions will appear in the manual:

Clipboard01

Clipboard03

and later

Clipboard04

prjemian:
Travis-ci fails:

```
lxml.etree.XMLSchemaParseError: Element '{http://www.w3.org/2001/XMLSchema}pattern': The value '^(?!_+$)[A-Za-z0-9_](?:[A-Za-z0-9_.]*[A-Za-z0-9_])?$' of the facet 'pattern' is not a valid regular expression., line 80

The command "python ./utils/test_suite.py" exited with 1.
```

vasole:
@prjemian 

Could it be you need to pre-pend an `r` in front of the string to avoid interpretation of characters by python?

prjemian:
The string is read from the XML Schema file.

prjemian:
The lxml package takes responsibility for reading the XML Schema file.  [Here is the code](https://github.com/lxml/lxml/blob/fa1d856cad369d0ac64323ddec14b02281491706/src/lxml/xmlschema.pxi#L87-L91) that generates the message reported:

```
        if self._c_schema is NULL:
            raise XMLSchemaParseError(
                self._error_log._buildExceptionMessage(
                    u"Document is not valid XML Schema"),
                self._error_log)
```

probably because this previous line returned `NULL`:

    self._c_schema = xmlschema.xmlSchemaParse(parser_ctxt)

That looks like a call into the libxml2 library.  Checking NXDL file validation with command line tool using libxml2:

```bash
$ xmllint --noout --schema nxdl.xsd base_classes/NXentry.nxdl.xml 
regexp error : failed to compile: xmlFAParseAtom: expecting ')'
regexp error : failed to compile: xmlFAParseRegExp: extra characters
nxdl.xsd:80: element pattern: Schemas parser error : Element '{http://www.w3.org/2001/XMLSchema}pattern': The value '^(?!_+$)[A-Za-z0-9_](?:[A-Za-z0-9_.]*[A-Za-z0-9_])?$' of the facet 'pattern' is not a valid regular expression.
WXS schema nxdl.xsd failed to compile
```

Is this regexp beyond the skills of libxml2?

prjemian:
PNG?

phyy-nx:
I mean, if there is actually no way to specify this crazy string without breaking the parser, a screen grab of it from another editor would work too.  Not a great solution of course.

prjemian:
Maybe it is time to [reconsider](https://github.com/nexusformat/definitions/pull/671#issuecomment-707362014) and allow group and field names composed of only underscore (`_`) characters.  Can the [RegExp](https://github.com/nexusformat/definitions/pull/671#issuecomment-707767192) (`^[a-zA-Z0-9_](?:[\w.]*\w)*$`) be refactored to avoid the `?:` (non-capturing group) expression?

vasole:
Allowing only underscore characters `^[a-zA-Z0-9_]{1}(([a-zA-Z0-9_]|\.)*[a-zA-Z0-9_]{1})?$`

https://regex101.com/r/5Q7b6C/14

prjemian:
When the `^` and `$` are removed from the left and right sides, respectively (`[a-zA-Z0-9_]{1}(([a-zA-Z0-9_]|\.)*[a-zA-Z0-9_]{1})?`), this passes the local validation test.

```
(base) prjemian@poof ~/.../NeXus/definitions $ xmllint --noout --schema nxdl.xsd base_classes/NXentry.nxdl.xml 
base_classes/NXentry.nxdl.xml validates
```

<details>

```
(bluesky_2020_9) prjemian@poof ~/.../NeXus/definitions $ python utils/test_nxdl.py
test__1__base_classes__NXaperture (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXattenuator (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXbeam (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXbeam_stop (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXbending_magnet (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXcapillary (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXcite (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXcollection (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXcollimator (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXcrystal (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXcylindrical_geometry (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXdata (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXdetector (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXdetector_group (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXdetector_module (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXdisk_chopper (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXentry (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXenvironment (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXevent_data (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXfermi_chopper (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXfilter (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXflipper (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXfresnel_zone_plate (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXgeometry (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXgrating (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXguide (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXinsertion_device (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXinstrument (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXlog (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXmirror (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXmoderator (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXmonitor (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXmonochromator (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXnote (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXobject (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXoff_geometry (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXorientation (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXparameters (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXpdb (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXpinhole (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXpolarizer (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXpositioner (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXprocess (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXreflections (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXroot (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXsample (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXsample_component (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXsensor (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXshape (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXslit (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXsource (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXsubentry (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXtransformations (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXtranslation (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXuser (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXvelocity_selector (__main__.Individual_NXDL_Tests) ... ok
test__1__base_classes__NXxraylens (__main__.Individual_NXDL_Tests) ... ok
test__2__applications__NXarchive (__main__.Individual_NXDL_Tests) ... ok
test__2__applications__NXarpes (__main__.Individual_NXDL_Tests) ... ok
test__2__applications__NXcanSAS (__main__.Individual_NXDL_Tests) ... ok
test__2__applications__NXdirecttof (__main__.Individual_NXDL_Tests) ... ok
test__2__applications__NXfluo (__main__.Individual_NXDL_Tests) ... ok
test__2__applications__NXindirecttof (__main__.Individual_NXDL_Tests) ... ok
test__2__applications__NXiqproc (__main__.Individual_NXDL_Tests) ... ok
test__2__applications__NXlauetof (__main__.Individual_NXDL_Tests) ... ok
test__2__applications__NXmonopd (__main__.Individual_NXDL_Tests) ... ok
test__2__applications__NXmx (__main__.Individual_NXDL_Tests) ... ok
test__2__applications__NXrefscan (__main__.Individual_NXDL_Tests) ... ok
test__2__applications__NXreftof (__main__.Individual_NXDL_Tests) ... ok
test__2__applications__NXsas (__main__.Individual_NXDL_Tests) ... ok
test__2__applications__NXsastof (__main__.Individual_NXDL_Tests) ... ok
test__2__applications__NXscan (__main__.Individual_NXDL_Tests) ... ok
test__2__applications__NXspe (__main__.Individual_NXDL_Tests) ... ok
test__2__applications__NXsqom (__main__.Individual_NXDL_Tests) ... ok
test__2__applications__NXstxm (__main__.Individual_NXDL_Tests) ... ok
test__2__applications__NXtas (__main__.Individual_NXDL_Tests) ... ok
test__2__applications__NXtofnpd (__main__.Individual_NXDL_Tests) ... ok
test__2__applications__NXtofraw (__main__.Individual_NXDL_Tests) ... ok
test__2__applications__NXtofsingle (__main__.Individual_NXDL_Tests) ... ok
test__2__applications__NXtomo (__main__.Individual_NXDL_Tests) ... ok
test__2__applications__NXtomophase (__main__.Individual_NXDL_Tests) ... ok
test__2__applications__NXtomoproc (__main__.Individual_NXDL_Tests) ... ok
test__2__applications__NXxas (__main__.Individual_NXDL_Tests) ... ok
test__2__applications__NXxasproc (__main__.Individual_NXDL_Tests) ... ok
test__2__applications__NXxbase (__main__.Individual_NXDL_Tests) ... ok
test__2__applications__NXxeuler (__main__.Individual_NXDL_Tests) ... ok
test__2__applications__NXxkappa (__main__.Individual_NXDL_Tests) ... ok
test__2__applications__NXxlaue (__main__.Individual_NXDL_Tests) ... ok
test__2__applications__NXxlaueplate (__main__.Individual_NXDL_Tests) ... ok
test__2__applications__NXxnb (__main__.Individual_NXDL_Tests) ... ok
test__2__applications__NXxrot (__main__.Individual_NXDL_Tests) ... ok
test__3__contributed_definitions__NXcontainer (__main__.Individual_NXDL_Tests) ... ok
test__3__contributed_definitions__NXcsg (__main__.Individual_NXDL_Tests) ... ok
test__3__contributed_definitions__NXcxi_ptycho (__main__.Individual_NXDL_Tests) ... ok
test__3__contributed_definitions__NXelectrostatic_kicker (__main__.Individual_NXDL_Tests) ... ok
test__3__contributed_definitions__NXmagnetic_kicker (__main__.Individual_NXDL_Tests) ... ok
test__3__contributed_definitions__NXquadric (__main__.Individual_NXDL_Tests) ... ok
test__3__contributed_definitions__NXquadrupole_magnet (__main__.Individual_NXDL_Tests) ... ok
test__3__contributed_definitions__NXseparator (__main__.Individual_NXDL_Tests) ... ok
test__3__contributed_definitions__NXsnsevent (__main__.Individual_NXDL_Tests) ... ok
test__3__contributed_definitions__NXsnshisto (__main__.Individual_NXDL_Tests) ... ok
test__3__contributed_definitions__NXsolenoid_magnet (__main__.Individual_NXDL_Tests) ... ok
test__3__contributed_definitions__NXsolid_geometry (__main__.Individual_NXDL_Tests) ... ok
test__3__contributed_definitions__NXspecdata (__main__.Individual_NXDL_Tests) ... ok
test__3__contributed_definitions__NXspin_rotator (__main__.Individual_NXDL_Tests) ... ok

----------------------------------------------------------------------
Ran 105 tests in 0.055s

OK
```

</details>


With this revision, travis-ci is expected to pass.

vasole:
The `^` it is necessary to prevent starting by a dot.

prjemian:
@vasole Please review.  We believe this is ready to be merged.

yayahjb (approved):

vasole (approved):
LGTM

benajamin:
What we need to express are the boundaries between what is:
- recommended,  (validator is happy)
- allowed but not recommended, (validator gives warning)
- what is not allowed (validator throws error)

prjemian:
taking the idea to start from Ben's suggestion above, this regexp: `^[a-zA-Z0-9_]([a-zA-Z0-9_.]*[a-zA-Z0-9_]+)?$`

prjemian:
https://regex101.com/r/1nJ3xb/2

regexp: 

    ^[a-zA-Z0-9_]([a-zA-Z0-9_.]*[a-zA-Z0-9_]+)?$

<details>
<summary>revised test list</summary>

```
# current valid names
_123
_valid
_Valid
_valid123
a
a____a
a___a
a__a
Valid
valid_
valid_name
valid123
validname
# new valid names
_123.123
_123.abc
_abc.123
_abc.abc
1
1.1
123
a.a
A.A
valid.name
# new valid names that are difficult to exclude
_
__
___
____
# invalid names
_.
.
._
.1
.123
.a
.A
.no_starting_period
1.
a.
A.
invalid name
invalid-name
invalid,name
invalid;name
invalid:name
invalid!name
invalid?name
invalid'name
invalid"name
invalid(name
invalid)name
invalid@name
invalid/name
invalid\name
invalid&name
invalid#name
invalid%name
invalid`name
invalid^name
invalid^name
invalid+name
invalid<name
invalid=name
invalid>name
invalid|name
invalid~name
invalid$name
no_ending_period.
```

</details>

manual/source/datarules.rst
```diff
@@ -40,45 +41,52 @@ described by the following :index:`rules <rules; naming>`:
...
+     https://github.com/nexusformat/definitions/pull/671#issuecomment-708395846
+
+     Also, an online test is shown here:
+     https://regex101.com/r/Yknm4v/1
     
    .. code-block:: text
        :linenos:
    
-       [a-z_][a-z\d_]*
+       ^[a-zA-Z0-9_]([a-zA-Z0-9_.]*[a-zA-Z0-9_]+)?$
```
PeterC-DLS, line 74: Is the `+` necessary in the group?

prjemian: Seems you are correct.

nxdl.xsd
```diff
@@ -64,8 +64,8 @@
 			<xs:documentation>
 				Used for allowed names of elements and attributes.  
 				Need to be restricted to valid program variable names.
```
PeterC-DLS, line 66: This sentence no longer applies

manual/source/datarules.rst
```diff
@@ -40,45 +41,52 @@ described by the following :index:`rules <rules; naming>`:
...
+   this regular expression. Note: It is *preferred* [#lc]_ to
+   use lower case for all field and group names.
+   All NeXus base class definitions will use lower case.:
+
+   ..
+     To understand this complicated RegExp, see
+     https://github.com/nexusformat/definitions/pull/671#issuecomment-708395846
+
+     Also, an online test is shown here:
+     https://regex101.com/r/Yknm4v/1
```
PeterC-DLS, line 69: Is this update to date?

manual/source/datarules.rst
```diff
@@ -40,45 +41,52 @@ described by the following :index:`rules <rules; naming>`:
...
-   It is recommended that all group and field names 
-   contain only these characters:
+   The names of NeXus *group* and *field* items must 
+   contain only these characters (the names *must* match 
+   the regular expression below):
    
-   * lower case letters
-   * digits
-   * "_" (underscore character)
+   * underscore (``_``), period (``.``), digit, or letter (upper or lower case)
```
PeterC-DLS, line 52: latin alphabet letter

manual/source/datarules.rst
prjemian: https://regex101.com/r/1nJ3xb/3

nxdl.xsd
```diff
@@ -64,8 +64,8 @@
 			<xs:documentation>
 				Used for allowed names of elements and attributes.  
 				Need to be restricted to valid program variable names.  
-				Note:  This means no "-" or "." characters can be allowed and
-				you cannot start with a number.
+				Note: No ``-`` characters (among others) are allowed
+				and you cannot start or end with a period (``.``).
 				HDF4 had a 64 character limit on names 
 				(possibly including NULL) and NeXus enforces this 
 				via the ``NX_MAXNAMELEN`` variable with
```
PeterC-DLS, line 71: Is that constant leftover from NAPI?

manual/source/datarules.rst
prjemian: revision coming to this entire block ... please stand by

nxdl.xsd
prjemian: yes

```
(bluesky_2020_9) prjemian@poof ~/.../NeXus/definitions $ git grep NX_MAXNAMELEN
manual/source/napi-f90.rst:``NX_MAXNAMELEN``  Maximum length of NeXus name            ``64``
nxdl.xsd:                               via the ``NX_MAXNAMELEN`` variable with 
nxdl.xsd:                       <xs:maxLength value="63" />   <!-- enforce via NX_MAXNAMELEN -->
```

prjemian: Are you suggesting we remove the statement?

PeterC-DLS: No, just that variable/constant does not exist. Various validators will enforce the limit.

prjemian: changing text to indicate it is enforced by the NAPI

prjemian:
Clipboard01

benajamin:
Proposal: Clarify naming convention for the names of *groups* and *fields* as the following:

1. Recommended names
 - lower case words separated by underscores and, if needed, with a trailing number.
 - NOTE: this is used by the NeXus base classes.
2. Allowed names
 - any combination of upper and lower case letter, numbers, underscores and periods, except that periods cannot be at the start or end of the string.
 - passes validation according to the regex `^[a-zA-Z0-9_]([a-zA-Z0-9_.]*[a-zA-Z0-9_])?$`
3. Invalid names
 - does not pass validation according to the regex `^[a-zA-Z0-9_]([a-zA-Z0-9_.]*[a-zA-Z0-9_])?$`

All NIAC members are asked to cast their vote by attaching a thumbs up emoji to accept, a thumbs down to reject and any other emoji to abstain.

prjemian:
At NIAC2020, it was suggested to add this text:

> The NIAC recognises that the majority of the world uses characters outside of the basic latin ([Standard ASCII](https://coding.tools/ascii-table)) set currently included in the allowed names. The restriction given here reflects current technical issues and we expect to revisit the issue and relax such restrictions in future.

PR #682, opened 2019-09-09
Clarify how beam_center_x and beam_center_y are used

phyy-nx:
beam_center_x and beam_center_y are only used if the depends_on field is not present.   Clarify this and provide a definition of the default geometry given no dependency chain.

### DIFF
--- a/applications/NXmx.nxdl.xml
+++ b/applications/NXmx.nxdl.xml
@@ -324,6 +324,20 @@
 							detector. This is a length and can be outside of the actual
 							detector. The length can be in physical units or pixels as
 							documented by the units attribute.
+
+							The depends_on field has priority over this field and is
+							designed to fully specify the geometry of an arbitrarily
+							positioned detector. However, if the depends_on attribute is
+							not set, then beam_center_x, beam_center_y, and distance are
+							used to position the detector, given the following
+							conventions (all in the NeXus coordinate system):
+
+							Assume the detector is orthogonal to the beam without any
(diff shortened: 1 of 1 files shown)


prjemian:
@zjttoefs - Do you have time review this before Friday?

prjemian:
@zjttoefs  -- review today?

prjemian:
I'll merge this evening unless there is an objection.

yayahjb:
If you have no axis chain, you can still process data if you have the beam
center, wavelength and detector distance and
assume a simple head-on detector orientation.  That is what allows people
to process miniCBF Pilatus images and
Eiger images with most of the metadata missing.  Unfortunately it also
leads to lots of mistakes and mis-processed
images, which is why much of the MX community has decided to push for more
complete metadata.

prjemian:
@yayahjb - Are you implying this needs more discussion and should not be merged at this time?

yayahjb:
Yes this needs more discussion, but the proposed merge is a useful step in
the right direction.  When we bring the
Gold Standard to NIAC we well may improve this wording and introduce an
additional tag to allow finer control
over processing in cases where both an axis chain and beam center are
provided.

  -- Herbert

prjemian:
OK, I will move this to a later milestone.

zjttoefs:
I agree, this is probably not the last word on the matter. But the PR helps clarify what we know and agree on.

phyy-nx:
Thanks.  Merging.

prjemian:
Thanks, all.

PR #701, opened 2020-01-22
Clarify documentation to fix #307

### LINKED ISSUE #307: NXDL data type vs NAPI type vs NeXus data type: clarify and simplify!
Manual 1.3.3.2. has a section on "NeXus Data Types". 

2.2.1.1 has "NXDL Data Types (internal)", and an attribute "type"

2.2.1.2 has "NXDL: Data Types", "Data Types allowed in NXDL specifications", "Data types for use in NXDL" (three wordings for the same?), and also mentions "the NAPI type".

Then of course there is the data field name "type", which has nothing to do with the above.

The relation between NeXus, NXDL, and NAPI types needs to be clearly exposed, and if possible simplified.

1.3.3.2 introduces far too briefly a "binary data" type. 2.2.1.2 introduces "NX_BINARY". The relation between both is unclear.

The suggestion to use NX_BINARY for text with Microsoft linebreaks looks very much ad hoc. This is not a serious way to specify a text exchange format. What about character encoding?

PeterC-DLS:
To distinguish between types used in NXDL schema and the data types used
by NeXus specifications, the former is now called NXDL field types.


manual/source/nxdl-types.rst
```diff
@@ -8,10 +8,10 @@ NXDL: Data Types and Units
 
 .. _nxdl-types:
 		
-Data Types allowed in NXDL specifications
+Field Types allowed in NXDL specifications
```
prjemian: needs more # characters to match length of preceding line

PeterC-DLS: Done

PR #706, opened 2020-01-22
Per shot wavelength

### LINKED ISSUE #667: How to specify per-shot wavelength
At XFELs, each shot has a different wavelength.  How can we specify this in NeXus? Looking the DIALS code that reads NeXus data, we are assuming that if incident_wavelength is an array, than each entry in that array is a wavelength for a different shot. However, I think we are mis-interpreting incident_wavelength:
http://download.nexusformat.org/doc/html/classes/applications/NXmx.html
"In the case of a polychromatic beam this is an array of the wavelengths with the relative weights in incident_wavelength_weight."

So what's the right way to do this?

Current problems in the documentation:
1. NXmx/beam is a base class NXbeam, but its documentation does not match the base class NXbeam documentation.  Compare:
http://download.nexusformat.org/doc/html/classes/applications/NXmx.html
http://download.nexusformat.org/doc/html/classes/base_classes/NXbeam.html
2. incident_wavelength_spectrum, an NXdata field defined in NXmx in its beam group, has no doc string, and is not included in the base class NXbeam documentation.
3. The base class NXbeam has a NXdata field with this docstring: "Distribution of beam with respect to relevant variable e.g. wavelength. This is mainly useful for simulations which need to store plottable information at each beamline component."  I don't understand what this means, can someone clarify it for me?

I think I could specify per-shot wavelength using the NXdata field of NXbeam or using incident_wavelength_spectrum, but the latter seems wrong since I'm not recording a spectrum.

Thoughts?

phyy-nx:
Fixes #667 

Detail:
- Allow incident_wavelength to be up to 4D, supporting per-shot spectra.  Several use cases are directly supported now, including 1 wavelength for the whole dataset, a constant polychromatic beam, a varying per-shot wavelength, per-shot individual spectra, and a spread for each wavelength.
- Add incident_wavelength_axes, used to specify which axes are present in incident_wavelength
- Deprecate incident_wavelength_weight. This field is where one of the dimensions of the spectrum of a polychromatic beam was specified. This is now folded into a dimension of incident_wavelength.
- Deprecate incident_wavelength_spectrum. This group was undocumented.
- Clarify incident_wavelength_spread should have same dimensionality as incident_wavelength if specified

### DIFF
--- a/applications/NXmx.nxdl.xml
+++ b/applications/NXmx.nxdl.xml
@@ -570,14 +570,14 @@
 
 							Several other use cases are supported. For each of these,
 							usage of incident_wavelength_axes is recommended and is
-							required if the ray is 2D or higher.
+							required if the array is 2D or higher.
 
 							In the case of a polychromatic beam this is a 2D array
 							of the wavelengths vs. their relative weights.
 
 							In the case of a monochromatic beam with varying
-							wavelength per image, this is a 1D array of length np,
-							listing all the per-image wavelengths.
+							wavelength per shot, this is a 1D array of length np,
+							listing all the per-shot wavelengths.
 
 							In the case of a polychromatic beam that varies per
 							image, this is a 3D array of np vs. wavelengths vs.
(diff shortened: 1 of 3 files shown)


phyy-nx:
This wasn't the right way to do it.  Closing.

PR #710, opened 2020-01-22
Correct slow/fast pixel direction documentation to fix #695

### LINKED ISSUE #695: Slow/fast axis inconsistent in NXdetector_module
See [NXdetector_module](http://download.nexusformat.org/doc/html/classes/base_classes/NXdetector_module.html), the `slow_pixel_direction` is a copy of `fast_pixel_direction` and also compare with [NXmx](http://download.nexusformat.org/doc/html/classes/applications/NXmx.html) application definition.

Also how slow/fast is defined is missing.

PeterC-DLS:


yayahjb:
The wording "along the direction of `fastest varying' pixel direction" and "along the direction of `slowest varying' pixel direction" is unnecessarily complex.  A better and more commonly used
terminology, at least in crystallography, would be "along the fast pixel direction" and "along the slow pixel direction"

PeterC-DLS:
The field names are called so, the doc merely expands on this.

yayahjb:
This issue is not the field names, which are fine and commonly used in crystallography.  The issue is the explanatory wording used in the documentation, which seems unnecessarily complex and confusing to me.  I think it better either to assume that the user understands array indexing or to slowly and carefully explain array indexing.  The wording "along the direction of the fastest varying pixel direction" just does not seem to flow trippingly from the tongue, especially with the two uses of "direction", and does not seem to add anything clarifying to the field names themselves.

PR #716, opened 2020-01-22
Set minOccurs to zero for unneeded items in NXsource of NXmx

### LINKED ISSUE #696: NXmx has fields that may not have to be required
[NXmx](http://download.nexusformat.org/doc/html/classes/applications/NXmx.html) has a `NXsource` with many fields that are required. Perhaps these have been mistakenly marked as such and should be optional as they do not apply for synchrotron sources.

PeterC-DLS:
Fix #696 as documentation for application definitions adds items
as required if they do not set minOccurs as zero


PeterC-DLS:
This PR is not necessary with other updates to NXmx.

Issue #503, opened 2016-10-21
Recording time-stamped data


mkoennecke:
Update the manual with the decisions of the NIAC 2016 on how to store streamed data in NeXus files.

zjttoefs:
NXlog has something and in the features there is code to explain it. At least that could be linked, but some examples and the fact that NXlog can replace datasets needs to be mentioned.

prjemian:
@mkoennecke @zjttoefs Can either of you provide this documentation?

zjttoefs:
I found a volunteer!

prjemian:
2020 Code Camp: We might be able to resolve this by mining the features repo as suggested above (per NXlog).

PR #719, opened 2020-01-23
Improved documentation about time stamped data

mkoennecke:
Refs #503
I have updated the information on storing time stamped data in NeXus files. I find this now sufficiently documented.  Some one else, please review and merge. It is now much better then it used to be.

### DIFF
--- a/manual/source/introduction.rst
+++ b/manual/source/introduction.rst
@@ -23,7 +23,7 @@ data format has three purposes:
    the case of streaming data acquisition, where time stamped data are logged. 
 #. *processed data*:
    NeXus also defines standards for processed data. This is data which has underwent some form of data 
-   reduction or data analysis. NeXus allows to store the results of such processing together with 
+   reduction or data analysis. NeXus allows storing the results of such processing together with 
    documentation about how the processed data was generated. 
 #. *standards*:
    NeXus defines standards in 
--- a/manual/source/introduction.rst
+++ b/manual/source/introduction.rst
@@ -14,18 +14,21 @@ to define a common data exchange format for neutron, X-ray, and muon experiments
 NeXus is built on top of the scientific data format HDF5 and adds 
 domain-specific rules for organizing data within HDF5 files in addition 
 to a dictionary of well-defined domain-specific field names. The NeXus 
-data format has two purposes:
+data format has three purposes:
 
 #. *raw data*:
    NeXus defines a format that can 
    serve as a container for all relevant data associated with a scientific 
-   instrument or beamline. 
-   This is a very important use case.
+   instrument or beamline. This is a very important use case. This includes 
+   the case of streaming data acquisition, where time stamped data are logged. 
 #. *processed data*:
+   NeXus also defines standards for processed data. This is data which has underwent some form of data 
+   reduction or data analysis. NeXus allows to store the results of such processing together with 
+   documentation about how the processed data was generated. 
+#. *standards*:
    NeXus defines standards in 
    the form of *application definitions* for the exchange of data 
-   between applications.  NeXus provides structures for raw experimental 
-   data as well as for processed data.  
+   between applications. NeXus provides standards for both raw and processed data.
 
 A community of scientists and computer programmers working in neutron
 and synchrotron facilities around the world came to the conclusion that a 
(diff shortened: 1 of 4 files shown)


vasole:
To replace something expected to be a dataset by a group is in itself not a good idea.

Furthermore, we are replacing it by a group where everything is optional. It would have been much, much simpler for a program to test if at the same level as NXdetector/data one could find a time or timestamp or whatever name to find the additional information.

That would prevent breaking of any code.

A smarter solution, still going in the direction of what this PR implements, is to keep the dataset data as it is, and to create an NXlog inside the NXdetector where value is a link to NXdetector/data with all the required time information. You get your NXlog and nobody gets broken code.

mkoennecke:
Armando, I am sorry to disappoint you but this has been discussed and decided upon at NIAC 2016. It just took embarrassingly long to update the documentation. And anyway, you cannot use the same code for treating streaming data then for more traditional data collection schemes. Except if you adapt the code and then you can also put in the check for NXlog.

vasole:
I agree you cannot use the same code to analyze the data, but you can use existing code to visualize the data in one case but not in the other.

FreddieAkeroyd:
Related to #503

manual/source/introduction.rst
```diff
@@ -14,18 +14,21 @@ to define a common data exchange format for neutron, X-ray, and muon experiments
...
 #. *raw data*:
    NeXus defines a format that can 
    serve as a container for all relevant data associated with a scientific 
-   instrument or beamline. 
-   This is a very important use case.
+   instrument or beamline. This is a very important use case. This includes 
+   the case of streaming data acquisition, where time stamped data are logged. 
 #. *processed data*:
+   NeXus also defines standards for processed data. This is data which has underwent some form of data 
+   reduction or data analysis. NeXus allows to store the results of such processing together with
```
PeterC-DLS: ~~allows to~~ use `can`

manual/source/strategies.rst
```diff
@@ -149,10 +149,17 @@ which also allows "energy" to be specified if one is so inclined.
...
-:ref:`NXevent_data` is used for storing neutron event data and :ref:`NXlog` would be used 
-for storing any other time-stamped data, e.g. sample temperature, chopper 
-top-dead-centre, motor position etc.
+
+Time-stamped data can be stored in both :ref:`NXlog` and :ref:`NXevent_data` structures. 
+Of the two, :ref:`NXlog` is the most important one, :ref:`NXevent_data` is used only for storing neutron event data 
+and :ref:`NXlog` would be used for storing any other time-stamped data, e.g. sample temperature, chopper top-dead-centre, 
+motor position, detector data etc.
+
+Regarding the NeXus file structure to use, there is one simple rule: just use the standard NeXus file structure but replace
```
benajamin:
```suggestion
Regarding the NeXus file structure to use, there is one simple rule: just use the standard NeXus file structure but insert 
```

manual/source/strategies.rst
```diff
@@ -174,7 +181,11 @@ to grab the data between 20 seconds and 40 seconds, and then trim that data to g
 want.
 Obviously in this simple example this does not gain us a lot, but it is easy to see that in a 
 large dataset having appropriately placed cues can save significant computational time when looking 
-up values in a certain time-stamp range.
+up values in a certain time-stamp range. NeXus has actually borrowed the cueing table concept 
+from video file formats where it allows viewing software to quickly access your most favourite scene.
```
benajamin:
```suggestion
from video file formats where it allows viewing software to quickly access your favourite scene. 
```

manual/source/strategies.rst
```diff
@@ -174,7 +181,11 @@ to grab the data between 20 seconds and 40 seconds, and then trim that data to g
 want.
 Obviously in this simple example this does not gain us a lot, but it is easy to see that in a 
 large dataset having appropriately placed cues can save significant computational time when looking 
-up values in a certain time-stamp range.
+up values in a certain time-stamp range. NeXus has actually borrowed the cueing table concept 
+from video file formats where it allows viewing software to quickly access your most favourite scene. 
+Correspondingly, cueing in NeXus allows you to quickly access your most favourate morsel of time stamped
```
benajamin:
```suggestion
Correspondingly, cueing in NeXus allows you to quickly access your favourite morsel of time stamped 
```

vasole:
@mkoennecke 

Another resolution against common sense approved in a meeting where I was not present.

How many NIAC representatives are actually developing and supporting code?

This resolution is the equivalent of changing the signature of a function instead of adding an optional argument.

PR #734, opened 2020-01-24
clarify the nature of application definitions

### LINKED ISSUE #697: Clarify the nature of NeXus classes described by Application Definitions
From Steve Collins, the chair of DLS's NeXus working group,

```
The nature of NeXus classes described by Application Definitions.

The question is, for what purpose are NeXus classes described by application definitions, where they are 
already described independently? 

Specific questions:

a.	Can instances of classes inside definitions require new fields that were previously optional?

b.	Can instances of classes inside definitions make optional new fields that were previously not 
mentioned? (E.g. ‘depends_on’ in NXsample/NXmx)

c.	Can instances of classes inside definitions require new fields that were previously not 
mentioned?

d.	If the answer to all three is ‘yes’ then can we view the process of defining classes within an 
application definition as defining a subclass of the original class? That is, all instances of the class 
within the definition are valid instances of the original class, but not vice-versa?

Also, how can application definitions be extended? We will probably want to use the NXmx definition but add optional fields for the sample surface orientation and the response of a polarization analyser (Jones matrix). NXmx contains just about everything else that we can think of for physical science diffraction applications such as magnetic and surface diffraction. Can we propose to achieve a new Application Definition by formally extending NXmx?
```

prjemian:
FAQ that help clarify the nature of application definitions to fix #697

### DIFF
--- a/manual/source/faq.rst
+++ b/manual/source/faq.rst
@@ -202,3 +202,49 @@ This is a list of commonly asked questions concerning the NeXus data format.
    See the many examples in the NeXus exampledata repository. 
    (https://github.com/nexusformat/exampledata)
 
+#. Can instances of classes inside definitions require 
+   new fields that were previously optional?
+
+   Yes. That is one of the motivations to have 
+   application definitions. By default, all content 
+   in an application definition is required.
+
+   For example, the ``radiation`` field in :ref:`NXcanSAS`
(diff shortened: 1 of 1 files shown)


prjemian:
@PeterC-DLS - Do you have time to review this before Friday?  Could see if this resolves the questions posed in #697 by Steve Collins.

PR #737, opened 2020-01-26
Update example for new axes syntax and for HDF 1.8.0 and above

### LINKED ISSUE #665: [DOC] C example uses old style NXdata
Nothing urgent. Just to comment that the C example about how to create a NeXus file with the original HDF5 API uses old style NXdata.

http://download.nexusformat.org/doc/html/examples/code_native.html

FreddieAkeroyd:
Attribute code refactored, updated for HDF 1.8.0, new axes syntax

fix #665

### DIFF
--- a/manual/source/examples/code_native.rst
+++ b/manual/source/examples/code_native.rst
@@ -19,9 +19,8 @@ Compare these code examples with :ref:`NAPI-Examples`.
 Writing a simple NeXus file using native HDF5 commands in C
 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 
-.. note:: This example uses the signal/axes attributes applied to the
-   data field, as described in :ref:`Design-FindPlottable-ByName`.
-   New code should use the method described in :ref:`Design-FindPlottable-NIAC2014`.
+.. note:: This example uses the new method described
+   in :ref:`Design-FindPlottable-NIAC2014` for indicating plottable data.
 
 .. literalinclude:: nxh5write.c
     :tab-width: 4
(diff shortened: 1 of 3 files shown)


rayosborn:
This example is fine as far as it goes, but `nxh5read.c` does not attempt to read the attributes in order to determine the axes (or signal, for that matter). I think we also need an example of how such a file should be parsed in order to determine the signal and axes without any prior knowledge, since that is the point of the attributes in the first place.

prjemian:
> 
> 
> This example is fine as far as it goes, but `nxh5read.c` does not attempt to read the attributes in order to determine the axes (or signal, for that matter). I think we also need an example of how such a file should be parsed in order to determine the signal and axes without any prior knowledge, since that is the point of the attributes in the first place.

@rayosborn - Do you see this change as required for this PR or an addition to be added later?  I'd like this PR to be part of the next release which will happen on Friday this week.  I could construct such a example for the [*h5py* examples](https://manual.nexusformat.org/examples/h5py/index.html) in the manual.  I'd leave it to others to write the *c* example.

rayosborn:
I don't want to hold up any release schedule. When I saw the issue, I thought that the intention the code was to show how to use the new indices, but this example is much more limited in scope.

prjemian:
OK, time to open a new issue for the deeper example.

The example I just created is nwo in the manual: https://manual.nexusformat.org/examples/h5py/index.html#finding-the-default-plottable-data

FreddieAkeroyd:
Hi, I think an h5py example as @prjemian has provided is more useful to most people, we can add a C axes reading example later though I wonder if a C++ HDF API example would be more useful?

PR #744, opened 2020-01-27
From code camp, Make NXmx/NXdata optional.

### LINKED ISSUE #746: NXmx/NXdata is required which makes NXreflections fail validation
This is for the NXreflections case where the processed data file could have NXreflections and NXmx sub entries, but the link to the original dataset is gone, therefore there cannot be a NXdata field.

(discussion in #744)

phyy-nx:
This is for the NXreflections case where the processed data file could have NXreflections and NXmx sub entries, but the link to the original dataset is gone, therefore there cannot be a NXdata field.

I imagine this one could require a bit more discussion?  Any thoughts on this?  It would be cool to get this into the next milestone too, if possible.

Fixes #746

### DIFF
--- a/applications/NXmx.nxdl.xml
+++ b/applications/NXmx.nxdl.xml
@@ -688,6 +688,6 @@
 					</attribute>
 				</field>
 			</group>
-			<group type="NXdata" />
+			<group type="NXdata" minOccurs="0" />
 		</group>
 </definition>
--- a/applications/NXmx.nxdl.xml
+++ b/applications/NXmx.nxdl.xml
@@ -171,6 +171,7 @@
 
 					<field name="data" type="NX_NUMBER">
 						<doc>
+							The pixel data recorded, if available.
 							For a dimension-2 detector, the rank of the data array will be 3.
 							For a dimension-3 detector, the rank of the data array will be 4.
 							This allows for the introduction of the frame number as the
@@ -688,6 +689,14 @@
 					</attribute>
 				</field>
 			</group>
-			<group type="NXdata" minOccurs="0" />
+			<group type="NXdata" minOccurs="0">
+				<doc>
+					The pixel data recorded, if available.
+					For a dimension-2 detector, the rank of the data array will be 3.
+					For a dimension-3 detector, the rank of the data array will be 4.
+					This allows for the introduction of the frame number as the
+					first index.
+				</doc>
+			</group>
 		</group>
 </definition>


prjemian:
Is there an issue related?

prjemian:
I'm marking this for the milestone this week.  If it's not resolved by OCB Thursday, I'll move it to the next milestone.

yayahjb:
I agree that you can have an interesting and useful MX dataset without raw
data.  Indeed you could also have an interesting
and useful MX dataset without reflections.  However, if you have no raw
diffraction images, but do have reflections, you certainly
could present those reflections in NXdata.  They can be plotted.  Indeed
those plots can be made to look very much like
the raw diffraction images.

prjemian:
I agree with @yayahjb that the one writing the file should make the best effort to identify a plottable data set and set up the attributes to make that default plottable data set easy to find.  This is one of the three [motivations](https://manual.nexusformat.org/motivations.html#simple-plotting) of NeXus.

phyy-nx:
Sure, but that would be a separate issue, right?  This pull request removes NXdata from the NXmx application definition which itself doesn't refer to NXreflections at all.  Requiring NXreflections to have NXdata is interesting, but is another proposal altogether, in my opinion.

phyy-nx:
(by remove I mean makes optional)

phyy-nx:
> Is there an issue related?

Added an issue. Thanks.

phyy-nx:
@yayahjb and @prjemian, I think you are proposing adding NXdata as an optional field to NXreflections.  I could create an issue for this.  Would that be acceptable?  If so, could I merge this tomorrow as part of the milestone?

prjemian:
Something is getting lost.  I disagree with the original assertion stated above:

> therefore there cannot be a NXdata field.

Should this be true, a data file is moot if there cannot be data.  These data files are not merely to document that an activity has occurred.

In NeXus, it is always possible to add an NXdata group.  In the case of NXmx, an optional NXdata group with no specified name, such as the one in question

    <group type="NXdata" />

is not necessary at all.  NXmx is not specifying what to name this group if it were provided, not providing any documentation about how to use it, and only relying on the existing rules to allow it under the most generic circumstances.  The NXmx definition will be easier to follow if you leave it out.   Remove this line entirely.  It provides no value here.

However, I advocate anyone writing a NeXus data file should always provide a NXdata group at the NXentry level unless there is absolutely no possible means to define one.  This is one of the three motivations for the NeXus standard.

phyy-nx:
Thanks for the details.  This pull request was motivated by NXreflections but is only for NXmx.  So if it helps, forget NXreflections for the context of this pull request.

If no pixel data is available, I'd still think I can create a valid NXmx file that describes the experimental geometry.  This file would be very useful as a blueprint.  It seems odd that nxvalidate would say it's not a valid NXmx file.  Hence, this pull request.

phyy-nx:
I put in some documentation to the NXdata field.  Hopefully that makes it more clear as what is going on.

yayahjb:
Actually I was proposing that the top level /entry/data:NXdata either
contain the reflection data or a link to reflection
data elsewhere.  That makes it easy to find the reflections for plotting.
However right now the NXreflections base
class does not provide any NXreflections group into which to gather all the
fields so they can all be linked to from
entry/data:NXdata.  I would suggest the structure of the base class
NXreflections have the mention of NXentry
be removed and that one or more MX application definitions use the base
class NXreflections to show examples
of how reflections should best be included in a dataset.  Among the
application definitions for which this would
be appropriate would be NXmx.  I think we should do this carefully, so I
would suggest we make the clean-up
change to NXreflections to it does not force its fields directly under
NXentry, and just put NXreflections as an
optional top-level group under NXentry with a comment recommending
appropriate links into NXdata.  Then
we can improve the examples as we gain experience in including indexed
reflections, merged or unmerged,
scaled or unscaled, in our H5 datasets with or without the raw image data.

Regards,
    Herbert

prjemian:
Let's separate our interest in any changes to NXreflections from this PR.  Make a separate issue, for next milestone. 

Is the field named ``data`` required to have that name *exactly*?  If exactly, then please note in the NXdata group documentation that the name ``data`` cannot be used since it is already reserved for a field.  Alternatively, if the name is flexible, then use ``<field name="DATA" type="NX_NUMBER">`` per #698 and #562.

phyy-nx:
@yayahjb I agree with your comments, but I agree it's a separate issue.  I'll make it later.

@prjemian I'm sorry, I'm not sure I understand.  There is a field named data under NXdetector, is this what you are referring to?  That is not at the same level of NXentry/NXdata so using the name 'data' for the NXdata group should be fine.

prjemian:
I'm looking at the classes in this or and see the field I described

prjemian:
Auto correction!

Looking at the changes in this PR ...

applications/NXmx.nxdl.xml
```diff
@@ -171,6 +171,7 @@
 
 					<field name="data" type="NX_NUMBER">
```
phyy-nx: @prjemian This line is under NXdetector and must be named data.

applications/NXmx.nxdl.xml
```diff
@@ -688,6 +689,14 @@
 					</attribute>
 				</field>
 			</group>
-			<group type="NXdata" />
+			<group type="NXdata" minOccurs="0">
+				<doc>
```
phyy-nx: @prjemian this line is under NXentry, is of class NXdata, and can be named anything

phyy-nx:
@prjemian do those comments help or is there still a question?  Thanks for your review.

applications/NXmx.nxdl.xml
prjemian: Understood.  Needed to view the whole file, not just the changes.

phyy-nx:
Thanks.  Merging.

PR #752, opened 2020-01-30
NXreflections tidy

### LINKED ISSUE #727: NXreflections still looks like an application definition even though it is now a base class
Ever since #574, NXreflections is a base class instead of an application definition, but it still looks like an application definitions, namely that it has NXentry at the top of its structure.

Propose dropping NXentry at the top of the structure for NXreflections.  
Propose also dropping 

"definition: (optional) NX_CHAR
NeXus NXDL schema to which this file conforms
Obligatory value: NXreflections"

I think this is also left over from when this was an application definition.

phyy-nx:
Fixes #727

Lots of whitespace here.  Look at commit 7b7e08b to see the changes.

### DIFF
--- a/base_classes/NXreflections.nxdl.xml
+++ b/base_classes/NXreflections.nxdl.xml
@@ -13,533 +13,533 @@
     <symbol name="m"><doc>number of experiments</doc></symbol>
   </symbols>
   <doc>This is a definition for reflection data from diffraction experiments </doc>
-    <field name="experiments" minOccurs="1">
-      <doc>The experiments from which the reflection data derives </doc>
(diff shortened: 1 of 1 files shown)


base_classes/NXreflections.nxdl.xml
```diff
@@ -13,543 +13,533 @@
     <symbol name="m"><doc>number of experiments</doc></symbol>
   </symbols>
   <doc>This is a definition for reflection data from diffraction experiments </doc>
```
prjemian: remove superfluous text: "This is a definition for"

phyy-nx:
👍

PR #774, opened 2020-05-07
NXmx: NXbeam can be on sample or instrument

phyy-nx:
As part of prepping the SwisFEL Jungfrau 1M detector example, @yayahjb moved the beam group from sample to instrument (see [Zenodo entry](https://zenodo.org/record/3526738#.XrNAABNKipr)).  NXbeam is an allowed group on the [NXsample](https://manual.nexusformat.org/classes/base_classes/NXsample.html) and [NXinstrument](https://manual.nexusformat.org/classes/base_classes/NXinstrument.html) base classes so it makes sense to modify [NXmx](https://manual.nexusformat.org/classes/applications/NXmx.html) to make NXbeam optional on sample but note it should be present on one of sample or instrument.

### DIFF
--- a/applications/NXmx.nxdl.xml
+++ b/applications/NXmx.nxdl.xml
@@ -576,7 +576,11 @@
 
 				<field name="temperature" units="NX_TEMPERATURE" minOccurs="0" />
 
-				<group type="NXbeam">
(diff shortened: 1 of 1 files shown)


phyy-nx:
Ok to merge? @prjemian 

Thanks

PR #776, opened 2020-05-27
Add ORCID as a field to NXuser

PeterC-DLS:
The ORCID system is widely used in scientific literature to identify authors. This adds the URI as an optional field

### DIFF
--- a/base_classes/NXuser.nxdl.xml
+++ b/base_classes/NXuser.nxdl.xml
@@ -68,7 +68,7 @@
 	</field>
 	<field name="ORCID">
 		<doc>
-			an author code, Open Researcher and Contributer ID,
+			an author code, Open Researcher and Contributor ID,
 			defined by https://orcid.org and expressed as a URI
 		</doc>
 	</field>
--- a/base_classes/NXuser.nxdl.xml
+++ b/base_classes/NXuser.nxdl.xml
@@ -66,5 +66,11 @@
 			address/contact database
 		</doc>
 	</field>
+	<field name="ORCID">
+		<doc>
+			an author code, Open Researcher and Contributer ID,
+			defined by https://orcid.org and expressed as a URI
+		</doc>
+	</field>
 </definition>
 


prjemian:
Wait to merge until all checks pass (they should since they passed last time and this is a documentation change).

prjemian:
All checks pass.

PeterC-DLS:
Thanks, shall I merge?

prjemian:
This does not seem controversial.  Still, the question hinges on *Do we need broader approval to merge this?*  @benajamin , @mkoennecke , @FreddieAkeroyd ?

prjemian:
Perhaps give a short deadline for comments but at least a business day before merging.

vasole:
I have nothing against this PR, but I wonder how all these NXuser fields are acceptable concerning the European GDPR. At the ESRF we aim at writing the DOI associated to the data inside NXuser instead of user information.

rayosborn:
IANAL, and I certainly don't know all the intricacies of GDPR, but it is my understanding that the ORCID itself is public, but the owner of the ORCID has the right to determine what information is publicly available. So I guess if a there is no problem adding a user name to the NeXus file (?), there should be no problem adding the ORCID. When facilities ask for the ORCID, do they say what it will be used for?

rayosborn:
Thinking about it a little more, I don't think this should concern NeXus. I presume that the GDPR can't forbid a database having a particular field in it. It can only affect the policy under which that field is filled in.

vasole:
For NeXus is clearly not a problem.

Concerning the facility, as far as is the user who writes or agrees to write the information it could be fine for the facility but clearly some NXuser fields like email, address and telephone_number are against GDPR and a facility providing that information should allow the user to request the information to be deleted.

prjemian:
@vasole : Should cautions be added to the documentation for the affected fields?  Can you supply some suggested text?

vasole:
I would add a caution for the whole NXuser field but give me time to think about the form.

The role field is basically the only one warranted to be GDPR compliant. The facility_user_id field might also provide enough anonymization.

ORCID is already a good step in the correct direction because the user has control about the information related to it and the user has the possibility to de-activate the ORCID account.

PR #777, opened 2020-06-02
Fix spelling in schema definition

PeterC-DLS:
Also fix mixed indentations and remove trailing spaces


prjemian:
@PeterC-DLS : merge at will.  Thanks for the review.

PR #784, opened 2020-06-26
NXroot. Use NX_DATE_TIME and not NX_CHAR

### LINKED ISSUE #783: NXroot - Wrong type for @file_time and @file_update_time
It should be NX_DATE_TIME and not NX_CHAR as written in documentation

@file_time: (optional) NX_CHAR

    Date and time file was originally created

@file_name: (optional) NX_CHAR

    File name of original NeXus file

@file_update_time: (optional) NX_CHAR

    Date and time of last file change at close

vasole:
Specify NX_DATE_TIME instead of NX_CHAR

closes #783

### DIFF
--- a/base_classes/NXroot.nxdl.xml
+++ b/base_classes/NXroot.nxdl.xml
@@ -37,13 +37,13 @@
             <item value="NXroot"></item>
         </enumeration>
     </attribute>
-    <attribute name="file_time">
+    <attribute name="file_time" type="NX_DATE_TIME">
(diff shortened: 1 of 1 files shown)


prjemian:
I added this for review at NIAC2020

prjemian:
NIAC2020 adds that we should update NX_DATE_TIME (& ISO8601) type to advise that time zone is recommended, otherwise local time zone is the default, per ISO8601

PR #799, opened 2020-09-29
Added missing required dim fields

RussBerg:
3 fields: **_sensor_size, region_origin_** and **_region_size_** all specified a **_rank_** but did not then specify the required **_dim_** fields as well, this pull request adds the missing dim fields so that the nxdl can be auto generated as well as pass **cnxvalidate**.

### DIFF
--- a/applications/NXarpes.nxdl.xml
+++ b/applications/NXarpes.nxdl.xml
@@ -105,15 +105,24 @@
         </field>
         <field name="sensor_size" type="NX_INT">
           <doc>number of raw active elements in fast and slow pixel dimension direction</doc>
-          <dimensions rank="2" />
+          <dimensions rank="2">
+          	<dim index="1" value="1" /> 
+          	<dim index="2" value="2" /> 
+          </dimensions>
         </field>
         <field name="region_origin" type="NX_INT">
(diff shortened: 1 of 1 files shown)


PeterC-DLS:
The placeholders should be letters or a word. See for example the `data` field in `NXdata` of [`NXmx`](https://github.com/nexusformat/definitions/blob/f5e596af971ce9e736403c8f9bf32eeb3de684bc/applications/NXmx.nxdl.xml#L120) where the placeholder is defined at [L44](https://github.com/nexusformat/definitions/blob/f5e596af971ce9e736403c8f9bf32eeb3de684bc/applications/NXmx.nxdl.xml#L44)

RussBerg:
I thought the same thing, but then that made the meaning of sensor_size, region_origin and region_size different, the docs for each do not seem like they are wanting to specify 2 dimensional **arrays** for each, the way I read them it seems like they are wanting to specify 2 values "in the form of an array" like [sizeY, sizeX] for example which is why I set the values to 1 and 2 respectively.

prjemian (approved):
LGTM

benajamin (approved):

PR #806, opened 2020-10-13
recommend providing time zone in NX_DATE_TIME

### LINKED ISSUE #805: recommend including time zone in NX_DATE_TIME
The documentation of the `NX_DATE_TIME` and `ISO8601` data types should recommend that time zone be included.  Per the ISO8601 standard, the default time zone (when not explicit) is the local time.  Similar/same language should appear in the [`datarules`](https://manual.nexusformat.org/datarules.html?highlight=nx_date_time#nexus-dates-and-times) page under the *NeXus dates and times* section.

https://github.com/nexusformat/definitions/blob/3a88c936b95a775ef3023fe2abc86bb49a89ba04/nxdlTypes.xsd#L509-L521

prjemian:
fixes #805

### DIFF
--- a/manual/source/datarules.rst
+++ b/manual/source/datarules.rst
@@ -11,9 +11,9 @@ This section describes the rules which apply for storing single data items.
 Naming Conventions
 ##################
 
-.. index::
+.. index:: 
    single: naming convention
-   single: NX; used as NX class prefix
+   single: NX; used as NX class prefix 
    single: attribute; NXclass
    single: NXclass (attribute)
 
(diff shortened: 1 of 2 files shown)


PeterC-DLS (changes requested):
Can you separate out the whitespace changes and text reflow from this PR?

PeterC-DLS (approved):
LGTM. Does raise the issue of where the local time zone is recorded.

vasole:
@PeterC-DLS 

To me it is enough to specify your local time zone in the ISO8601.

As far as I understand this proposal, you can provide any time zone. What is needed is to remove the ambiguity when there is no time zone specified. The norm is that if there is no time zone, it is assumed local time, however, when a file moves from one country to another it is undefined. If the local time zone is written, the ambiguity is gone.

yayahjb (approved):

PR #813, opened 2020-10-19
@default for every base class except NXdata

### LINKED ISSUE #812: Confirm that default attribute is allowed for any group that is not NXdata
as described in #641

Look in `nxdl.xsd`

prjemian:
fixes #812 (from NIAC2018), accepted by vote at [NIAC2018](https://www.nexusformat.org/NIAC2018Minutes.html).

### DIFF
--- a/base_classes/NXaperture.nxdl.xml
+++ b/base_classes/NXaperture.nxdl.xml
@@ -53,4 +53,17 @@
         <doc>Description of aperture</doc>
     </field>
     <group type="NXnote"><doc>describe any additional information in a note*</doc></group>
+    <attribute name="default">
+        <doc>
+            .. index:: plotting
+            
+            Declares which child group contains a path leading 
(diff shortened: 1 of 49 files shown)


PeterC-DLS:
Wiki link: wiki.nexusformat.org/2014_How_to_find_default_data does not exist!

vasole:
Two comments

- I would change the text:
``
Declares which child group contains a path leading to a :ref:`NXdata` group.
`` 
to
``
Declares which child group, if any, contains a path leading to a :ref:`NXdata` group.
`` 

- And I would remove any reference, except under NXentry, or NXroot to the default plot:

 `` 
It is recommended (as of NIAC2014) to use this attribute to help define the path to the default dataset to be plotted.
            See http://wiki.nexusformat.org/2014_How_to_find_default_data
            for a summary of the discussion.
``

yayahjb (approved):

PR #818, opened 2020-10-19
add NXdata/@auxiliary_signals  attribute

### LINKED ISSUE #815: NXdata auxiliary signals
Per [NIAC2018](https://www.nexusformat.org/NIAC2018Minutes.html) approved vote, 

> additional group level attribute auxiliary_signals which is an array of strings holding the names of additional signals to be plotted with the signal. They all need to be of the same shape.

prjemian:
fixes #815 per NIAC2018 approved vote

### DIFF
--- a/base_classes/NXdata.nxdl.xml
+++ b/base_classes/NXdata.nxdl.xml
@@ -54,8 +54,9 @@
 			.. index:: plotting
 			
 			Array of strings holding the names of additional signals to
-			be plotted with the signal. Each auxiliary signal needs to
-			be of the same shape as the default signal.
+			be plotted with the default signal (specified by the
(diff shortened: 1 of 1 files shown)


benajamin (commented):
The second sentence is ambiguous. I suggest:
"Each auxiliary signal needs to be of the same shape as the default signal."

benajamin (commented):
Further clarified:
"Array of strings holding the names of additional signals to
be plotted with the default signal (specified by the "signal" attribute).
Each auxiliary signal needs to be of the same shape as the default signal."

PeterC-DLS (approved):
LGTM

prjemian:
Thanks!

rayosborn (approved):

Issue #648, opened 2019-02-16
Documentation glitch?


prjemian:
@vasole reports through the email listserv:

In http://download.nexusformat.org/sphinx/design.html one can read:

rgba, hsla and cmyk are (4 x n x m) arrays, where the 4 channels are the colour channels appropriately

I would appreciate that being written as (n x m x 4) because it is actually what people are going to write/use. I cannot imagine people writing the R component and nxm bytes later find the corresponding G component and so on. Usually one has the RGBARGBARGBA... and not RRRGGGBBBAAA...

phyy-nx:
From Telco, @prjemian will fix (no branch needed)

PR #834, opened 2020-10-19
add 5 interpretation attribute values for color images

### LINKED ISSUE #829: “rgbimage”, “rgbaimage”, ...
Per [NIAC2014](https://www.nexusformat.org/NIAC2014_Meeting#Minutes.html) approved vote, 

> “rgbimage”, “rgbaimage”, “hslimage”, “hslaimage”, “cmykimage” to be added to the interpretation attribute of datasets to encode 3d datasets with colour as added dimension to the 2d image. Improve documentation as well.

prjemian:
fixes #829

### DIFF
--- a/manual/source/design.rst
+++ b/manual/source/design.rst
@@ -183,11 +183,12 @@ attributes can be found in the next table:
 			* ``scaler`` (0-D data)
 			* ``spectrum`` (1-D data)
 			* ``image`` (2-D data)
+			* ``rgb-image`` (3-D data)
 			* ``rgba-image`` (3-D data)
+			* ``hsl-image`` (3-D data)
 			* ``hsla-image`` (3-D data)
 			* ``cmyk-image`` (3-D data)
 			* ``vertex`` (3-D data)
-
 
 .. index::
    ! single: file attribute
--- a/manual/source/design.rst
+++ b/manual/source/design.rst
@@ -180,7 +180,8 @@ attributes can be found in the next table:
 			(`n x m x 3`).
 			Allowed values include:
 
-			* ``scaler`` (0-D data)
+			* ``scalar`` (0-D data)
+			* ``scaler`` DEPRECATED, use ``scalar``
 			* ``spectrum`` (1-D data)
 			* ``image`` (2-D data)
 			* ``rgb-image`` (3-D data)
--- a/nxdl.xsd
+++ b/nxdl.xsd
@@ -884,14 +884,13 @@
 					<xs:simpleType>
                         <xs:restriction base="nx:NX_CHAR">
                         	<xs:enumeration value="scalar"/>
-                        	<xs:enumeration value="scaler"/>
                         	<xs:enumeration value="spectrum"/>
                         	<xs:enumeration value="image"/>
-                        	<xs:enumeration value="rgbimage"/>
-                        	<xs:enumeration value="rgbaimage"/>
-                        	<xs:enumeration value="hslimage"/>
-                        	<xs:enumeration value="hslaimage"/>
-                        	<xs:enumeration value="cmykimage"/>
+                        	<xs:enumeration value="rgb-image"/>
+                        	<xs:enumeration value="rgba-image"/>
+                        	<xs:enumeration value="hsl-image"/>
+                        	<xs:enumeration value="hsla-image"/>
+                        	<xs:enumeration value="cmyk-image"/>
                         	<xs:enumeration value="vertex"/>
                         </xs:restriction>
 					</xs:simpleType>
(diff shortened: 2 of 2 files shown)


prjemian:
@vasole Can you review this PR?  Check the documentation strings for each new interpretation value.

vasole:
I understand the expected shape for RGB is then (nrows, ncolumns, 3) and for RGBA (nrows, ncolumns, 4) in a C-order representation. Is it worth to write it? Some people might confuse "last dimension" when using MatLab or Fortran where the same buffer is (3, ncolumns, nrows) or (4, ncolumns, nrows).

In the same file you mention "scalar" and "scaler" for the same interpretation. This ambiguity has led us to support both everywhere. It would be worth to write both of them.

nxdl.xsd
```diff
@@ -871,6 +871,11 @@
 							* ``scaler`` = 0-D data to be plotted
 							* ``spectrum`` = 1-D data to be plotted
 							* ``image`` = 2-D data to be plotted
+							* ``rgbimage`` = 2-D images with colour as added 3rd dimension
```
PeterC-DLS, line 874: Can these match https://github.com/nexusformat/definitions/blame/master/manual/source/design.rst#L186

PeterC-DLS (changes requested):

PeterC-DLS:
Also #648 clarified Armando's concern about storage order

prjemian:
@vasole:  you identified a typo:

```
~/.../NeXus/definitions $ grep scaler nxdl.xsd 
							* ``scaler`` = 0-D data to be plotted
~/.../NeXus/definitions $ grep scalar nxdl.xsd 
                        	<xs:enumeration value="scalar"/>
```

prjemian:
Don't know how to deprecate `scaler` within the `nxdl.xsd` file.  Can just mark that in the documentation.

nxdl.xsd
prjemian: Those names in the design chapter were added 6 years ago.  I'll update this PR to use the hyphenated versions AND the associated text.

prjemian:
@vasole @PeterC-DLS Please review again.

vasole:
It is fine with me. Thank you!

nxdl.xsd
prjemian: with recent changes, they match now

rayosborn:
@vasole, does SILX have a way of plotting CMYK and HSL images?

vasole:
I just notify @t20100 

RGB and RGBA are certainly supported but we never asked ourselves about CMYK and HSL.

PeterC-DLS (approved):

manual/source/design.rst
```diff
@@ -183,12 +183,13 @@ attributes can be found in the next table:
 			* ``scaler`` (0-D data)
```
PeterC-DLS, line 183: This needs updating

PeterC-DLS (changes requested):

PeterC-DLS (approved):

t20100:
In SILX only RGB(A) is supported, but I guess adding conversion of CMYK and HSL to RGB is not a big deal.

BTW, why not using the last dimension of the dataset to know whether is it RGB (3) or RGBA (4) (same for HSL (3), HSLA (4)) as already described in the `@interpretation` documentation rather than adding more `interpretation`s?

From a software reading NeXus point of view, to be resilient we'll still have to check the size of last dimension to decide, i.e., `rgb-image` and `rgba-image` will be handled the same way.

rayosborn:
If NeXpy, if the last dimension is of size 3 or 4, a 'Plot (RGBA) Image' option is automatically enabled, so the user is given a choice how to plot it. I guess that, if the `interpretation` attribute were added, we could make that the default, but I agree that it makes no difference if it's an 'rgb-image' or an 'rgba-image'. I'm pleased that SILX doesn't provide automatic support for CMYK images, because it's not something I particularly want to support in NeXpy either. I can do the conversion, but it will be unreliable, because CMYK values depend on the intended print method, and slow.

I just discovered that Python has a [colorsys module](https://docs.python.org/3.8/library/colorsys.html) that does conversions between RGB, HLS and HSV colors (but not CMYK). I mention this for two reasons. One is that they call it HLS, not HSL. A quick google suggests that both are used, but HSL seems more common, so I think we can probably stick with this proposal. The other is to ask whether we should also add `hsv-image` to the list for completeness. Personally, I would prefer it if we had stuck with RGB, since that is what screens display and it seems to be the only system that is universally supported, but we seem to have already voted on it, so it is already part of the NeXus cruft.

PR #840, opened 2020-10-20
Add _mask as a reserved suffix.

### LINKED ISSUE #814: Attribute Mask and Normalisation
Per [NIAC2018](https://www.nexusformat.org/NIAC2018Minutes.html) approved vote, 

> `DATASETNAME_mask` as a general way to specify a mask. We use the conventions as described for `NXdetector/pixel_mask` with the option to use less that 32 bit.

prjemian:
fixes #814 as approved at [NIAC2018](https://www.nexusformat.org/NIAC2018Minutes.html)


yayahjb (approved):

rayosborn:
I would suggest "Field containing a signal mask, where 0 means the pixel is not masked. If required, bit masks are defined in :ref:`NXdetector` ``pixel_mask``."

prjemian:
@rayosborn Could you review the changes?

rayosborn (approved):

Issue #842, opened 2020-10-20
fix target attribute for external file links


prjemian:
As noted in #841 and as directed by [NIAC2018](https://www.nexusformat.org/NIAC2018Minutes.html#decision_target), the `target` attribute does not work for external file links.  Fix the problem.

prjemian:
@rayosborn [commented](https://github.com/nexusformat/definitions/pull/843#issuecomment-713062589):

The nexusformat API use the `target` attribute to define an external link, e.g.,
```
cspbbr3_300K['f1/data/data'] = NXlink(target='/entry/data/data', file='300K/f1.h5')
```
This is used to define the path within the external file to the linked data, but it is not, of course, stored in the external file itself. When `nexusformat` reads the external link, it creates an NXlink object that then contains this target, derived from the HDF5 external link definition, and this link resolves to the external file path when it is accessed. Since HDF5 disguises external links - it's actually quite difficult to find out that it is an external link - you need special calls to the h5py get function. 

To summarize, our use of the `target` attribute is specific to `nexusformat` and is useful when using the API, but it is not a property of the stored parent file. It's not really a technical limitation - I think in a way, it shouldn't be part of the standard. It's up to the file writer whether they want the data to be external, but there is no difference to the innocent user. It's only an issue for any API the NIAC does or does not support.  I don't know if this helps or not.

vasole:
The problem of being or not stored in the external file it will depend on whether the target attribute is created or not.

If the target attribute is created, the destination field will also have it. It is much cleaner to use just pure HDF5 links (external or soft) and inquire the physical target of the link if that information is actually needed.

vasole:
The target attribute is not needed if only the HDF5 backend is considered because standard SoftLinks and ExternalLinks already provide that functionality.

If you want an XML backend, you may need it (or something similar) to emulate HDF5 links.

prjemian:
This was missed for the 2022-06 Code Camp.

PR #843, opened 2020-10-20
DOC: note the `target` attribute does not work for external file links

### LINKED ISSUE #841: DOC NeXus Target Attribute: only works for internal links
At [NIAC2018](https://www.nexusformat.org/NIAC2018Minutes.html#decision_target), it was agreed:

> The target attribute denoting the origin of a linked data item is broken as it only works for internal links.
> 
> We discussed this a little but could not come to an agreement. The technical issues raised by this problems need to be looked at in more detail.
> 
> Thus this Proposal: While the NIAC is working at the problem of the broken target attribute for external links, we document the problem with it.

prjemian:
Fixes #841 as directed by [NIAC2018](https://www.nexusformat.org/NIAC2018Minutes.html#decision_target), note the `target` attribute does not work for external file links.

### DIFF
--- a/manual/source/design.rst
+++ b/manual/source/design.rst
@@ -284,6 +284,10 @@ path [#absolute_address]_ to the *original* dataset.
       rather than a **relative** address (starting from the current group,
       such as: ``detector/polar_angle``).
 
+      .. note:: The ``target`` attribute does not work for
+         :ref:`external file links<design.external_file_link>`.
+         The NIAC is working at resolving the technical limitations
+
    .. [#hdf5_hard_link] HDF5 hard link:
       https://portal.hdfgroup.org/display/HDF5/H5L_CREATE_HARD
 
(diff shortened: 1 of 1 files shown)


rayosborn:
The nexusformat API use the `target` attribute to define an external link, e.g.,
```
cspbbr3_300K['f1/data/data'] = NXlink(target='/entry/data/data', file='300K/f1.h5')
```
This is used to define the path within the external file to the linked data, but it is not, of course, stored in the external file itself. When `nexusformat` reads the external link, it creates an NXlink object that then contains this target, derived from the HDF5 external link definition, and this link resolves to the external file path when it is accessed. Since HDF5 disguises external links - it's actually quite difficult to find out that it is an external link; you need special calls to the h5py get function. 

To summarize, our use of the `target` attribute is specific to `nexusformat` and is useful when using the API, but it is not a property of the stored parent file. I don't know if this helps or not.

prjemian:
@rayosborn I'm going to copy your comment to the #842, where the `target` attribute is to be fixed.

benajamin (approved):

PR #845, opened 2020-10-20
DOC: clarify target value is a suggestion

### LINKED ISSUE #600: Target attributes in application definitions
I encountered this when putting a NXsubentry following the NXtomo definition into an existing file. 
In entry/NXdata of NXtomo we specify certain places where the link should point to. I ran cnxvalidate against my file and it flagged those entries. Because, the way the file is constructed, the links point 
to wherever I found the data. To my understanding, this is valid NeXus. 

IMHO, we should remove prescribed target links from application definitions and correspondingly not validate them. Or leave them in as an example and only validate that links point to valid HDF5 objects.

prjemian:
fixes #600


prjemian:
@mkoennecke @zjttoefs You both commented on #600.  Please review.

benajamin (approved):

Issue #541, opened 2017-03-08
datarules.rst contains information on 2014 discussions


zjttoefs:
We should revise what has been accepted and put that in the manual directly.
Could be up for discussion what level of in progress material is right for the manual.

prjemian:
edit the part related to "2014 discussion"

prjemian:
Do this with a branch+PR

prjemian:
2020 Code Camp: Has any of this work already been completed?  Where?  Push this off to next milestone.

prjemian:
2014 minutes: https://www.nexusformat.org/NIAC2014_Meeting#Minutes.html

Is this about the uncertainties and axes attributes or are there other items to act on?

prjemian:
@benajamin Please identify what proposals were approved that still need to be acted on.

PeterC-DLS:
The page that's always not linked to correctly: https://www.nexusformat.org/2014_axes_and_uncertainties.html

prjemian:
That link is already in the manual, it appears here https://manual.nexusformat.org/datarules.html#examples as the footnote `(7)`.

> More examples are available in the NeXus webpage (7).

prjemian:
Back to the original statement:

> We should revise what has been accepted and put that in the manual directly.

Here is my analysis of approved items from [NIAC2014](https://www.nexusformat.org/NIAC2014_Meeting#Minutes.html) minutes that have not yet been acted upon:

* **CIF style marked as preferred method ...** (#)
* **Vote on proposal for NXdata axes ...** (#)
* **Clarification decision:  ...** (#)
* **“rgbimage”, “rgbaimage”,  ...** (#)

prjemian:
This reST link target in `datarules.rst`:

```
.. _Design-FindPlottable-NIAC2014:
```

is used elsewhere in the manual, such as:

```
manual/source/datarules.rst:   Perhaps substitute with the discussion from NIAC2014?
manual/source/examples/code_napi.rst:   New code should use the method described in :ref:`Design-FindPlottable-NIAC2014`.
manual/source/examples/code_native.rst:   in :ref:`Design-FindPlottable-NIAC2014` for indicating plottable data.
manual/source/history.rst:    See the description in :ref:`Design-FindPlottable-NIAC2014`
manual/source/rules.rst:  as described in :ref:`Design-FindPlottable-NIAC2014`.
```

The section linked, `:ref:`Design-FindPlottable-NIAC2014` contains the footnote mentioned above.

[...]

PR #847, opened 2020-10-20
Update URLs to correct destinations

PeterC-DLS:
as mentioned in https://github.com/nexusformat/definitions/issues/541#issuecomment-712194880


prjemian (approved):
Good catch!

mkoennecke (approved):
I am fine with this

PR #848, opened 2020-10-20
clarify notation of tree structure

### LINKED ISSUE #768: clarify the notation used in tree structure
It was asked (by private email)

> But, a related really simple question: what is the meaning of @ in the specification?  @signal: ‘I’ - that signal is an attribute not a dataset?

prjemian:
fixes #768

### DIFF
--- a/manual/source/design.rst
+++ b/manual/source/design.rst
@@ -48,6 +48,22 @@ used by NeXus. These are:
 
 In the following sections these elements of NeXus files will be defined in more detail.
 
+.. _tree.structure:
+
+.. note:: Notation used to describe a NeXus data file
+
+   In various places in the NeXus manual, contents of a NeXus
+   data file are described using a tree structure, such as
+   in the :ref:`Introduction<fig.verysimple-structure>`.
+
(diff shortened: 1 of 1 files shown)


mkoennecke (approved):
This is fine with me

rayosborn:
If we are going to define these trees, I would argue that we should use the `nexusformat` output. 
```
root:NXroot
  @file_name = '/content/example.h5'
  @file_time = '2020-10-20T21:43:16.578946'
  @default = 'entry'
  entry:NXentry
    @default = 'data'
    data:NXdata
      @axes = ['Qk', 'Qh']
      @signal = 'data'
      Qh = float64(51)
      Qk = float64(41)
      data = float64(41x51)
```
In any case, I'm not sure that it is necessary to give such an elaborate description. The original confusion was about the @ character, so we could say that it refers to attributes. Everything else is pretty self-explanatory.

rayosborn:
I guess I spent too much time writing my comment, since I missed the merge by 20 minutes. I will have to write my own PR.

PR #858, opened 2020-10-27
Fix up NXBeam symbols and polarization

### LINKED ISSUE #708: NXBeam symbols are redundant
The symbol table in NXBeam that is going to be added by #706 lists `i` and `j` as possible symbols, but they appear to mean the same thing in the rest of the documentation.  Should `j` be removed and we only have `i`?

phyy-nx:
Fixes #708.  Needs community review to see which if any of these fields are in common use, but I'd first like review from NIAC members to see if these changes are reasonable.

Set of fields in question:
- extent (length)
- incident_beam_divergence (angle)
- incident_polarization (any)
- final_polarization (any)
- final_beam_divergence (angle)


yayahjb:
I think it is premature to change what is in NXmx just for the difference between English and American spelling without a lot of discussion first.  -- Herbert

phyy-nx:
I purport that the only code actually reading/writing this field while creating NXmx files is using the American spelling already: https://github.com/dials/dials/blob/master/util/nexus/nx_mx.py#L208, which matches the spelling in NXbeam.  So renaming the field to incident_polarization_stokes synchronizes with existing use.

I'd need to back this up with an email survey, but that's the basis behind my proposal.

prjemian:
I recall a previous NIAC discussion (?decision?) where a particular preference would be chosen to avoid such debates. Anyone remember this discussion from more than 10 years ago?  As I recall, the preference was for the British spelling but this is not found by me in the manual.  One way to gather this is to search for items where the spelling would differ, such as _colour_ or _polarise_ or ...

phyy-nx:
> I recall a previous NIAC discussion (?decision?) where a particular preference would be chosen to avoid such debates. Anyone remember this discussion from more than 10 years ago? As I recall, the preference was for the British spelling but this is not found by me in the manual. One way to gather this is to search for items where the spelling would differ, such as _colour_ or _polarise_ or ...

No instances of _colour_ or _color_ in definitions.  _polarise_ only matches NXmx.  _polarize_ matches many files, including [NXpolarizer](https://manual.nexusformat.org/classes/base_classes/NXpolarizer.html).

prjemian:
```
(base) prjemian@poof ~/.../NeXus/definitions $ git grep -i polari
applications/NXmx.nxdl.xml:                                     <field name="incident_polarisation_stokes" recommended="true">
base_classes/NXbeam.nxdl.xml:        scattering by the sample, e.g., energy transfer, polarizations. </doc>
base_classes/NXbeam.nxdl.xml:    <field name="incident_polarization" type="NX_FLOAT" units="NX_ANY">
base_classes/NXbeam.nxdl.xml:        <doc>Polarization vector on entering beamline component</doc>
base_classes/NXbeam.nxdl.xml:    <field name="final_polarization" type="NX_FLOAT" units="NX_ANY">
base_classes/NXbeam.nxdl.xml:        <doc>Polarization vector on leaving beamline component</doc>
base_classes/NXguide.nxdl.xml:        To describe polarizing supermirrors such as used in neutron reflection,
base_classes/NXguide.nxdl.xml:        to include :ref:`NXpolarizer` and/or :ref:`NXmirror`.
base_classes/NXinstrument.nxdl.xml:    <group type="NXpolarizer" />
base_classes/NXpolarizer.nxdl.xml:    name="NXpolarizer" 
base_classes/NXpolarizer.nxdl.xml:      A spin polarizer.
base_classes/NXpolarizer.nxdl.xml:        <doc>description of the composition of the polarizing material</doc>
base_classes/NXpolarizer.nxdl.xml:        <doc>polarizing efficiency</doc>
contributed_definitions/NXsnsevent.nxdl.xml:        <group type="NXpolarizer" minOccurs="0"/>
contributed_definitions/NXsnshisto.nxdl.xml:        <group type="NXpolarizer" minOccurs="0"/>
```

```
(base) prjemian@poof ~/.../NeXus/definitions $ git grep -i colour
manual/source/design.rst:                       4 channels are the colour channels appropriately. If the image data
```

Search for `color` shows lots of code (.svg, .pdf, .xsl) that are false positives but no text items in the standard.  

So, no clear, *consistent* preference but a slight edge to `polarize` over `polarise` is indicated.  That, by itself, is not definitive.

rayosborn:
I think we should use 'polarize'. It's of course standard in the US, whereas Britain has gradually shifted to using the 'ise' suffix for most words in recent years. However, journals like Nature still use the 'ize' suffix and  I think [many/most etymologists](https://en.wikipedia.org/wiki/Oxford_spelling) think it is more correct. Even the [Oxford School of Neutron Scattering](https://www.oxfordneutronschool.org/2011/lectures/osns_stewart_polarised_2011.pdf) uses 'polarize'.

PeterC-DLS:
I remember that there's differing spelling preference in British dictionaries: the OED favours _-ize_; others _-ise_

phyy-nx:
Hey, I don't know why the build is failing.  When I run util/test_suite.py on my local system on the `nxbeam_fixup` branch it passes.  Am I missing something?

prjemian:
In the travis-ci logs (https://travis-ci.org/github/nexusformat/definitions/jobs/739717918), at line 364, a problem is reported with NXmx:

```
======================================================================

ERROR: test__2__applications__NXmx (test_nxdl.Individual_NXDL_Tests)

----------------------------------------------------------------------

Traceback (most recent call last):

  File "/home/travis/build/nexusformat/definitions/utils/test_nxdl.py", line 56, in validate_xml

    result = NXDL_SCHEMA.assertValid(xml_tree)

  File "src/lxml/etree.pyx", line 3623, in lxml.etree._Validator.assertValid

lxml.etree.DocumentInvalid: Element '{http://definition.nexusformat.org/nxdl/3.1}field': Character content other than whitespace is not allowed because the content type is 'element-only'., line 824

During handling of the above exception, another exception occurred:

Traceback (most recent call last):

  File "/home/travis/build/nexusformat/definitions/utils/test_nxdl.py", line 90, in test_wrap

    validate_xml(nxdl_file_name)

  File "/home/travis/build/nexusformat/definitions/utils/test_nxdl.py", line 61, in validate_xml

    raise NXDL_Invalid(msg)

test_nxdl.NXDL_Invalid: applications/NXmx.nxdl.xml : Element '{http://definition.nexusformat.org/nxdl/3.1}field': Character content other than whitespace is not allowed because the content type is 'element-only'., line 824
```

This passes validation test locally for me as well.  The only thing I see that is unusual is line 826 uses space indentation while the other lines nearby use tabs.  This does not show in the GitHub view but _does_ show up in my VSCode editor locally.

The report *Character content other than whitespace is not allowed* drives me crazy since I don't see the non-whitespace.  Also, other `field` elements in the file are split to multiple lines, so it is not the end-of-line character that triggers.

An alternative test is to use `xmllint` on the command line:

```
(bluesky_2020_9) prjemian@poof ~/.../NeXus/definitions $ xmllint --noout --schema nxdl.xsd applications/NXmx.nxdl.xml 
applications/NXmx.nxdl.xml validates
```

I'm not seeing it.

prjemian:
Clipboard01

phyy-nx:
Thanks @prjemian!  Worked!

prjemian:
@phyy-nx Will this be ready for review today?  Or move to next milestone?

phyy-nx:
Let's move it to the next milestone

applications/NXmx.nxdl.xml
```diff
@@ -821,8 +821,23 @@
...
+							component using Stokes notation
+						</doc>
+						<dimensions rank="2">
+							<dim index="1" value="np" />
+							<dim index="2" value="4" />
+						</dimensions>
+					</field>
 
-					<field name="incident_polarisation_stokes" recommended="true">
+					<field name="incident_polarization_stokes" recommended="true">
```
PeterC-DLS, line 836: Add field types as `NX_FLOAT` in both fields

PeterC-DLS:
> 
> Perhaps NX_NUMBER is more general?

They are numbers between -1 and 1 so NX_FLOAT is best!

yayahjb:
Actually NX_NUMBER is more appropriate.  The 4 Stokes parameters are
experimental measurements which may or may
not have been normalized.  If normalized, you are right that they lie
between -1 and 1, but those are not the actual measurements.

PeterC-DLS:
Is there any preference for the sign convention of the V component of the Stokes vector (i.e. it is defined looking down the beam towards its source or in the opposite direction)?

PeterC-DLS:
I think a reference plane will be needed to determine the meaning of the Stokes vector if another coordinate system is used. I propose another vector with an `NXtransformation` group, similar to #927:
```
polarization_reference_normal
  @transformation_type: translation
  @vector: three values that define the normal to the reference plane
```
If this is missing then default to using (0,1,0).

base_classes/NXbeam.nxdl.xml
```diff
@@ -43,84 +56,167 @@
...
         <dimensions rank="1">
-            <dim index="1" value="i"/>
+            <dim index="1" value="m"/>
         </dimensions>
     </field>
     <field name="energy_transfer" type="NX_FLOAT" units="NX_ENERGY">
         <doc>Energy change caused by beamline component</doc>
         <dimensions rank="1">
-            <dim index="1" value="i"/>
+            <dim index="1" value="m"/>
```
phyy-nx, line 71: This one I don't know if it should be `value="m"` or `value=nP`.  Thoughts?

phyy-nx:
Ok, I've applied all the feedback from Spring 2022 NIAC session B.  Ready for further review and hopefully a vote :)

base_classes/NXbeam.nxdl.xml
```diff
@@ -26,10 +26,23 @@
...
+            These symbols coordinate datasets with the same shape.
+        </doc>
+        <symbol name="nP">
+            <doc>Number of scan points.</doc>
+        </symbol>
+        <symbol name="m">
+            <doc>Number of channels in the incident beam spectrum, if known</doc>
+        </symbol>
+        <symbol name="c">
+            <doc>Number of components representing beam divergence (x, y, xy, etc.)</doc>
```
phyy-nx, line 40: Moments representing beam divergence

base_classes/NXbeam.nxdl.xml
```diff
@@ -43,84 +56,167 @@
...
+            the sample (or specimen) after all monochromation and
+            collimation.
+
+            The dimension **c** is a series of components of that represent
+            the standard uncertainty (e.s.d.) of the directions of
+            photons. The first and second components are in the XZ and YZ
+            planes around the mean source beam direction, respectively.
+
+            Further components in **c** characterize co-variance terms, so
+            the next component is the product of the first two, and so on.
```
phyy-nx, line 150: Drop second paragraph and change components to moments.

base_classes/NXbeam.nxdl.xml
```diff
@@ -43,84 +56,167 @@
...
+            of dimensions **nP** by **m** (slow to fast) with the
+            relative weights in ``incident_wavelength_weights`` as a 2D
+            array.
+
+            Note, :ref:`variants &lt;Design-Variants&gt;` are a good way
+            to represent several of these use cases in a single dataset,
+            e.g. if a calibrated, single-value wavelength value is
+            available along with the original spectrum from which it
+            was calibrated.
+            Wavelength on entering beamline component
```
phyy-nx, line 107: Move sections of this text to top and generally drop the dimnensionality of the other fields, where it's up to the application definition to specify the use of nP, m, both, or neither.

base_classes/NXbeam.nxdl.xml
yayahjb: m

base_classes/NXbeam.nxdl.xml
phyy-nx: After review, it's better to leave this text here.  See comment below.

phyy-nx:
As part of Spring NIAC 2022 it was suggested to move the explanation of use cases for polychromatic beams from incident_wavelength to the intro as the dimensionality explained there would apply to other fields.  After review, this is not the case.

The fields in question were `incident_energy`, `final_energy`, `energy_transfer`, `final_wavelength` and`final_wavelength_spread`.  I think they should retain their dimensionality of `**m**` (number of channels) as the other possible dimension `**nP**` does not seem relevant.  Thoughts?

Otherwise I believe this is ready for vote again.

phyy-nx:
@PeterC-DLS and @pshafer-als have requested we discuss the sign convention of the V component of the Stokes vector, right?  Can we cover this in our closeout meeting tomorrow?  Thanks!

pshafer-als:
> @PeterC-DLS and @pshafer-als have requested we discuss the sign convention of the V component of the Stokes vector, right? Can we cover this in our closeout meeting tomorrow? Thanks!

I would request assessing the sign convention for all 3 bi-directional components.

pshafer-als:
E.g., Here is is [a sign convention on the wikipedia page](https://en.wikipedia.org/wiki/File:StokesParameters.png), though I'm not sure if it specifies the time-dependence...i.e., is it traveling toward you or away from you?

pshafer-als:
...and here is the convention used in [Jo Stöhr's book](https://rd.springer.com/book/10.1007/978-3-540-30283-4), which matched the convention in my beamline controls.

image

**Left-circular polarization:**

image

**Right-circular polarization:**

image

phyy-nx:
Ok, the NXtransformations group @PeterC-DLS proposes in #927 for NXBeam establishes the coordinate system for the beam, and if not present then the beam is in the lab coordinate system of NeXus.  From there we follow the Stokes convention as you cite, since the propagation of the beam is in the Z direction, and the directions of X and Y are well [defined](https://manual.nexusformat.org/design.html#the-nexus-coordinate-system).

Perhaps obvious to other folks, but after my crash course in polarization this evening, I now know that S0..S3 is the [same](https://en.wikipedia.org/wiki/Stokes_parameters#Stokes_vectors) as IQUV, so I don't think we need to note one or the other in our definition.  Just citing that these are Stokes parameters seems sufficient to me.

base_classes/NXbeam.nxdl.xml
```diff
@@ -26,101 +26,197 @@
...
-        <doc>Divergence of beam entering this component</doc>
-        <dimensions rank="2"><!-- [2,:] -->
-            <dim index="1" value="2"/>
-            <dim index="2" value="j"/>
+        <doc>
+            Beam crossfire in degrees parallel to the laboratory X axis
+
+            The dimension **c** is a series of moments of that represent
+            the standard uncertainty (e.s.d.) of the directions of
+            photons. The first and second moments are in the XZ and YZ
```
phyy-nx, line 146: photons should be beam: "of the beam"

base_classes/NXbeam.nxdl.xml
```diff
@@ -26,101 +26,197 @@
...
     <field name="final_wavelength" type="NX_FLOAT" units="NX_WAVELENGTH">
         <doc>Wavelength on leaving beamline component</doc>
         <dimensions rank="1">
-            <dim index="1" value="i"/>
+            <dim index="1" value="m"/>
         </dimensions>
     </field>
-    <field name="incident_polarization" type="NX_FLOAT" units="NX_ANY">
+    <field name="incident_polarization" type="NX_NUMBER" units="NX_ANY"
+        deprecated="use incident_polarization_stokes, see https://github.com/nexusformat/definitions/issues/708">
```
phyy-nx, line 174: Don't deprecate

base_classes/NXbeam.nxdl.xml
```diff
@@ -26,101 +26,197 @@
...
-            <dim index="1" value="2"/>
-            <dim index="2" value="j"/>
+        <dimensions rank="2">
+            <dim index="1" value="nP"/>
+            <dim index="2" value="2"/>
         </dimensions>
     </field>
-    <field name="final_polarization" type="NX_FLOAT" units="NX_ANY">
+    <field name="final_polarization" type="NX_NUMBER" units="NX_ANY"
+        deprecated="use final_polarization_stokes, see https://github.com/nexusformat/definitions/issues/708">
```
phyy-nx, line 182: Don't deprecate

base_classes/NXbeam.nxdl.xml
```diff
@@ -26,101 +26,197 @@
...
-        <dimensions rank="2"><!-- [2,:] -->
-            <dim index="1" value="2"/>
-            <dim index="2" value="j"/>
+        <dimensions rank="2">
+            <dim index="1" value="nP"/>
+            <dim index="2" value="2"/>
+        </dimensions>
+    </field>
+    <field name="incident_polarization_stokes" type="NX_NUMBER" units="NX_ANY">
+        <doc>Polarization vector on entering beamline component using Stokes notation</doc>
```
PeterC-DLS, line 190: The Stokes parameters are four components labelled I,Q,U,V or S_0,S_1,S_2,S_3. These are defined with the standard Nexus coordinate frame unless it is overridden by an NXtransformations field pointed to by a depends_on attribute. The last component, describing the circular polarization state, is positive for a right-hand circular state - that is the electric field vector rotates clockwise at the sample and over time when observed from the source.

pshafer-als: Am I correct in presuming that +Q(S1) is parallel to the x-axis, and that +U(S2) is in the (+x,+y) quadrant (i.e., tan(angle from x toward y) > 0...as is the case [here](https://en.wikipedia.org/wiki/Stokes_parameters#Stokes_vectors)?

Whatever the sign conventions for these 3 components, please document them **explicitly**.  This will avoid confusion for end users, and there is no obvious downside to documenting this.

phyy-nx:
Ok, ready for review again.  I took @PeterC-DLS's polarization description verbatim using a co-commit, and then following  @pshafer-als's suggestions I added what I think is enough to specify the sign conventions.  Thoughts?  Thanks!

base_classes/NXbeam.nxdl.xml
```diff
@@ -170,31 +170,48 @@
...
-        <doc>Polarization vector on entering beamline component using Stokes notation</doc>
+        <doc>
+             Polarization vector on entering beamline component using Stokes notation
+
+             The Stokes parameters are four components labelled I,Q,U,V or S_0,S_1,S_2,S_3.
+             These are defined with the standard Nexus coordinate frame unless it is
+             overridden by an NXtransformations field pointed to by a depends_on attribute.
+             The last component, describing the circular polarization state, is positive
+             for a right-hand circular state - that is the electric field vector rotates
+             clockwise at the sample and over time when observed from the source.
```
pshafer-als, line 196: Because sign conventions are needed for all polarized comments, I would favor a uniform approach as suggested below.

base_classes/NXbeam.nxdl.xml
```diff
@@ -170,31 +170,48 @@
...
+
+             I is the beam intensity (often normalized to 1), and Q, U, and V are derived
+             from p (the fractional degree of polarization), and the elliptical components
+             Ψ and Χ (psi and chi). Ψ is the orientation angle of the polarization ellipse,
+             relative to the x axis and increasing counter-clockwise around z. Χ is the
+             ellipticity angle, or the arctan of the ratio of the minor/major axes of the
+             polarization ellipse.
```
pshafer-als, line 203: The references to Psi & Chi & 'p' are useful background info, but are not used in NXbeam or elsewhere in NeXus. I recommend to instead state the sign conventions of S1..S3 (Q, U, V).

```suggestion
              I (S_0) is the beam intensity (often normalized to 1). Q, U, and V scale 
              linearly with the total degree of polarization, and indicate the relative
              magnitudes of the pure linear and circular orientation contributions.
              
              Q (S_1) is linearly polarized along the x axis (Q > 0) or y axis (Q < 0).
              
              U (S_2) is linearly polarized along the x==y axis (U > 0) or the
              -x==y axis (U < 0). 

              V (S_3) is circularly polarized. V > 0 when the electric field vector rotates
              clockwise at the sample with respect to time when observed from the source; 
              V < 0 windicates the opposite rotation.
```

pshafer-als (changes requested):
@phyy-nx :  Thank you for making these updates to NXbeam.  We are looking forward to using them.

I've made some suggestions for how to clearly document the sign conventions of the Stokes parameters in NX.

base_classes/NXbeam.nxdl.xml
phyy-nx: Looks great, thanks!  Reviewed in code camp with @PeterC-DLS

base_classes/NXbeam.nxdl.xml
```diff
@@ -26,101 +26,220 @@
...
+        </symbol>
+    </symbols>
     <doc>
         Properties of the neutron or X-ray beam at a given location. 
-        
-        It will be referenced
-        by beamline component groups within the :ref:`NXinstrument` group or by the :ref:`NXsample` group. Note
-        that variables such as the incident energy could be scalar values or arrays. This group is
+
+        This group is referenced
```
benajamin, line 46: I suggest "This group is intended to be referenced..."

base_classes/NXbeam.nxdl.xml
```diff
@@ -26,101 +26,220 @@
...
+
+        Note that incident_wavelength and related fields can be a scalar values or arrays, depending on the use case.
+        To support these use cases, the explicit dimensionality of these fields is not specified, but it can be inferred
+        by the presense of and shape of accompanying fields, such as incident_wavelength_weights for a polychromatic beam.
+        </doc>
     <field name="distance" type="NX_FLOAT" units="NX_LENGTH">
         <doc>Distance from sample</doc>
     </field>
     <field name="incident_energy" type="NX_FLOAT" units="NX_ENERGY">
         <doc>Energy on entering beamline component</doc>
```
benajamin, line 61: I assume this refers to the energy of each particle/photon in the beam, rather than the power of the beam. Might be helpful to make the doc string more specific. Suggest: "Energy carried by each particle of the beam on entering the beamline component"

base_classes/NXbeam.nxdl.xml
```diff
@@ -26,101 +26,220 @@
...
         <dimensions rank="1">
-            <dim index="1" value="i"/>
+            <dim index="1" value="m"/>
         </dimensions>
     </field>
     <field name="final_energy" type="NX_FLOAT" units="NX_ENERGY">
         <doc>Energy on leaving beamline component</doc>
```
benajamin, line 67: Suggest: "Energy carried by each particle of the beam on leaving the beamline component"

base_classes/NXbeam.nxdl.xml
```diff
@@ -26,101 +26,220 @@
...
         <dimensions rank="1">
-            <dim index="1" value="i"/>
+            <dim index="1" value="m"/>
         </dimensions>
     </field>
     <field name="energy_transfer" type="NX_FLOAT" units="NX_ENERGY">
         <doc>Energy change caused by beamline component</doc>
```
benajamin, line 73: Suggest: "Change in particle energy caused by the beamline component"

base_classes/NXbeam.nxdl.xml
phyy-nx: 👍

benajamin:
Need to replace < and > characters with &lt; and &gt; in order for code to work properly and pass CI checks. I tried to do it but we had a commit collision.

phyy-nx:
Thanks, looks like I got it.  All conflicts are fixed and suggestions are applied.

pshafer-als (approved):
These revision look great.  Thank you for all your effort on this!

PeterC-DLS (approved):
LGTM

applications/NXmx.nxdl.xml
```diff
@@ -858,17 +858,30 @@
                             <item value="rectangular"/>
                         </enumeration>
                     </field>
-
-
-                    <field name="incident_polarisation_stokes" recommended="true">
+                    <field name="incident_polarisation_stokes" type="NX_NUMBER" recommended="false"
```
soph-dec, line 861: This change makes "incident_polarisation_stokes" required (see [here](https://nexusformat.github.io/definitions/classes/applications/NXmx.html#nxmx-entry-instrument-beam-incident-polarisation-stokes-field)). I guess it should have been changed to `optional="true"`.

PeterC-DLS: The release has already been tagged. Does this prevent you from validating files?

PR #859, opened 2020-10-27
Add _errors to NXreflections

### LINKED ISSUE #724: Consider changing _var to _errors in NXreflections
Based on 2018 NIAC, we should be using _errors to track error in a measurement (See https://www.nexusformat.org/NIAC2018Minutes.html#uncertainties-continuation)

However, NXreflections uses _var to describe variances on the standard deviation.  Should we change to using _error instead?  Would we then store standard deviations instead of variances?  How would we document this?  Would we deprecate the affected fields, remove them, or leave them but note they are not recommended?

This would affect these fields:
- observed_frame_var
- observed_px_x_var
- observed_px_y_var
- observed_phi_var
- observed_x_var
- observed_y_var
- int_prf_var
- int_sum_var

phyy-nx:
Fixes #724

Considered deprecating the _var fields but I'm not convinced there is a need too, even though the spec did get a bit busy through this change.

### DIFF
--- a/base_classes/NXreflections.nxdl.xml
+++ b/base_classes/NXreflections.nxdl.xml
@@ -267,6 +267,19 @@
     </attribute>
   </field>
 
+  <field name="observed_frame_errors" type="NX_FLOAT" units="NX_UNITLESS" minOccurs="1">
+    <doc>
+      The standard deviation of the estimate of the frame at which the central
+      impact of the reflection was recorded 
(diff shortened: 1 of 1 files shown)


yayahjb (approved):
I am approving, but I agree with Pete's comment that this had gotten rather busy and suggest we revisit for a clean-up in a year or so.

PR #866, opened 2020-11-16
NXdetector: move start_time & end_time OUT of efficiency group

### LINKED ISSUE #865: [NXdetector] Bug on documentation layout?
We would like to put a sort of timestamp on a per frame basis. Looking at the information concerning NXdetector it seems the proper place is start_time[np] and end_time[np]. However, as it is presented, it has to be inside an NXdata named efficiency where a dataset efficiency is mandatory....

https://manual.nexusformat.org/classes/base_classes/NXdetector.html#nxdetector

Is that a documentation layout bug and start_time, end_time, and real_time should be **outside** the efficiency NXdata group?

prjemian:
fixes #865

### DIFF
--- a/base_classes/NXdetector.nxdl.xml
+++ b/base_classes/NXdetector.nxdl.xml
@@ -434,7 +434,8 @@
       Most often there is a single real time value that is constant across
       an entire image frame.  In such cases, only a 1-D array is needed.
       But there are detectors in which the real time
-      changes per pixel. In that case, more than one dimension is needed.
+      changes per pixel. In that case, more than one dimension is needed. Therefore
+      the rank of this field should be less than or equal to (detector rank + 1).
     </doc>
     <dimensions rank="3">
       <dim index="1" value="np" />
--- a/base_classes/NXdetector.nxdl.xml
+++ b/base_classes/NXdetector.nxdl.xml
@@ -428,8 +428,13 @@
 
   <field name="real_time" type="NX_NUMBER" units="NX_TIME">
     <doc>
-      real-time of the exposure (use this if exposure time varies for
-      each array element, otherwise use ``count_time`` field)
+      Real-time of the exposure (use this if exposure time varies for
+      each array element, otherwise use ``count_time`` field).
+	    
+      Most often there is a single real time value that is constant across
+      an entire image frame.  In such cases, only a 1-D array is needed.
+      But there are detectors in which the real time
+      changes per pixel. In that case, more than one dimension is needed.
     </doc>
     <dimensions rank="3">
       <dim index="1" value="i" />
--- a/base_classes/NXdetector.nxdl.xml
+++ b/base_classes/NXdetector.nxdl.xml
@@ -425,6 +425,20 @@
       </dimensions>
     </field>
 
+    <field name="real_time" type="NX_NUMBER" units="NX_TIME">
+      <doc>
+        real-time of the exposure (use this if exposure time varies for
+        each array element, otherwise use ``count_time`` field)
+      </doc>
+
+      <dimensions rank="3">
+        <dim index="1" value="i" />
+        <dim index="2" value="j" />
+        <dim index="3" value="k" />
+      </dimensions>
+    </field>
+  </group>
+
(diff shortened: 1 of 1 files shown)


vasole:
I still see real_time should be out of the efficiency NXdata group.

base_classes/NXdetector.nxdl.xml
```diff
@@ -424,6 +424,19 @@
...
       </dimensions>
     </field>
+  </group>
+
+  <field name="real_time" type="NX_NUMBER" units="NX_TIME">
+    <doc>
+      real-time of the exposure (use this if exposure time varies for
+      each array element, otherwise use ``count_time`` field)
+    </doc>
+    <dimensions rank="3">
```
PeterC-DLS, line 434: Should this really be 3D?

vasole: Most likely not and 1D is enough. After all we just want a number per frame.

prjemian: The question is: _Is there such a detector that can vary the time across the frame?_  I believe the answer is _yes_ when looking at fast event-counting (so called fast pixel) detectors which a frame represents a number of disparate counting events.  It is not realistic to gather the many individual events from such a detector as raw data.  Rather, the composite frame is accumulated from the event stream.

prjemian: @ronpandolfi: Is the fast pixel detector at ALS such an example?

vasole: @prjemian Let's put it like this. Can one use a 1D dataset when rank is set to 3 in the definition?

prjemian: I agree that a 1-D should be allowed.  That's the most typical case.  How do we describe this in NXDL?  How is it described in other base classes when a variable rank array is allowed?

vasole: I guess it will be the same as for the polar angle:

``
This is the polar angle of the detector towards the previous component in the instrument; most often the sample. The usage depends on the nature of the detector. Most often it is the polar_angle of the detector assembly. But there are irregular detectors. In this case, the polar_angle must be specified for each detector pixel.
``

We could state a similar syntax and sentence here:

``
Most often there is a real time for the detector signal but there are detectors in which the real time changes per pixel and in that case more than one dimension is needed.
``

PeterC-DLS: So should be <= detector rank? (Still not 3D.)

base_classes/NXdetector.nxdl.xml
```diff
@@ -424,6 +424,24 @@
...
+      Real-time of the exposure (use this if exposure time varies for
+      each array element, otherwise use ``count_time`` field).
+	    
+      Most often there is a single real time value that is constant across
+      an entire image frame.  In such cases, only a 1-D array is needed.
+      But there are detectors in which the real time
+      changes per pixel. In that case, more than one dimension is needed.
+    </doc>
+    <dimensions rank="3">
+      <dim index="1" value="i" />
```
PeterC-DLS, line 440: If same as `polar_angle` then 1:`np`, 2:`i`, 3:`j`

PeterC-DLS (changes requested):

base_classes/NXdetector.nxdl.xml
PeterC-DLS: @prjemian, @vasole https://github.com/nexusformat/definitions/pull/866/commits/01de4733ed1b2cc488ffc9154de9a4a45c3c5662 okay?

prjemian: @PeterC-DLS Could you add your comment as additional advice in the documentation?

> If same as polar_angle then 1:np, 2:i, 3:j

PeterC-DLS: How's this https://github.com/nexusformat/definitions/pull/866/commits/3430099d4974faa61303e0325c01b21f1ffb5836

prjemian: Clear.  Thanks.

PR #878, opened 2021-01-02
create HTML anchors for all fields, groups, attributes, and links in NXDL documentation

### LINKED ISSUE #855: Technical demonstration of terminology anchors
Per https://github.com/nexusformat/NIAC/issues/73, add HTML anchors for every defined term in the manual.

### LINKED ISSUE #877: multiple hyperlink targets
During work on #855, the study revealed that NXaperture has two anonymous NXgeometry groups.  Since the default name of `GEOMETRY` will be used for each, this situation *now* generates an error from Sphinx while building the manual since there will be duplicate hyperlink targets.  Same situation for NXcanSAS and NXdata.

_Originally posted by @prjemian in https://github.com/nexusformat/definitions/issues/855#issuecomment-753529818_

prjemian:
FIX #855 (and FIX #877)


prjemian:
Aaack! Fails unit testing since we still support Py3.5 and f-strings were added in py3.6.

prjemian:
Searching for the hyperlink targets in the generated rst files and the HTML anchors in the generated HTML files, the numbers are *not* identical.  2411 hyperlink targets were created from which 1843 HTML anchors were created.  This means that more work is needed to identify which targets are not generating anchors.  And why.

@hgoerzig : Before I continue with that analysis, can you confirm this work is proceeding in the direction you wish?

[html-anchors.txt](https://github.com/nexusformat/definitions/files/5760910/html-anchors.txt)
[rst-targets.txt](https://github.com/nexusformat/definitions/files/5760911/rst-targets.txt)

---
FWIW: here are the commands to recreate the two files just uploaded.

```
make all
grep ".. _/NX" build/manual/source/classes/*/*.rst > /tmp/rst-targets.txt
grep "id=\"nx" build/manual/build/html/classes/*/*.html | grep "class=\"target\"" > /tmp/html-anchors.txt
```

prjemian:
Looking through the HTML anchors file, there are no matches to search for this text: `-group"` which means none of the *group* names are anchored yet.  Might be easy to fix.

prjemian:
Spot-checked the NXroot NXDL file and found HTML anchor `nxroot-entry-group`.  It did not show up in the results above because it did not have the `class="target"` attribute on the HTML field.  Here is a revised HTML anchors file:

[html-anchors.txt](https://github.com/nexusformat/definitions/files/5760926/html-anchors.txt)

produced by: `grep "id=\"nx" build/manual/build/html/classes/*/*.html  > /tmp/html-anchors.txt`

prjemian:
* 2411 rst targets
* 2517 HTML anchors

closer than before, tell me what you think

prjemian:
In the HTML anchors for example, there is:

```
build/manual/build/html/classes/contributed_definitions/NXspin_rotator.html:  <div class="section" id="nxspin-rotator">
```

which was created by some other part of the code, not this issue. So, nothing more to change that is easy for me to see.

hgoerzig:
Yes this is what I meant! :-)
I did not pull the code a made it run to check. There is one comment or idea. I don't know how you see it. I looked at the html-anchors.txt file:
```
build/manual/build/html/classes/applications/NXarchive.html:  <div class="section" id="nxarchive">
build/manual/build/html/classes/applications/NXarchive.html:<div><p id="nxarchive-entry-group"><strong>entry</strong>: (required) <a class="reference internal" href="../base_classes/NXentry.html#nxentry"><span class="std std-ref">NXentry</span></a></p>
build/manual/build/html/classes/applications/NXarchive.html:<div><span class="target" id="nxarchive-entry-index-attribute"></span><p id="index-2"><strong>&#64;index</strong>: (required) <a class="reference internal" href="../../nxdl-types.html#nx-char"><span class="std std-ref">NX_CHAR</span></a></p>
```
The ids that are generated always have appdefname-entry-term-type` id="nxarchive-entry-index-attribute"` .   
Am I right that the entry part of the id is only generated in appdefs ? In this case it might look easier for persons to read if there is the prefix appdef- and the 'entry' could be left away?
Would there be doubles if the last part (attribute, field, group) would be left away? This really might be a matter of taste.
Anyway the whole thing is mainly for machines.

prjemian:
Fantastic!

You are right that the `-entry-` part appears in application definitions. It shows the nesting in the defined structure.  It does not appear in the base classes (except for NXroot, NXentry, and NXsubentry, all of which define an `entry`).  Here are a few of the anchors in the `NXdata.html` and also the matching `entry` ids from the base classes:

```
nxdata-variable-last-good-attribute
nxdata-variable-axis-attribute
nxdata-variable-errors-field
nxdata-data-signal-attribute
nxdata-data-axes-attribute
nxentry-entry-identifier-field
nxentry-entry-identifier-uuid-field
nxentry-entry-identifier-uuid-version-attribute
nxroot-entry-group
nxsubentry-entry-identifier-field
```

And, yes, the final assignment of attribute, group, field, or link prevents duplicates.

prjemian:
@PeterC-DLS - This is ready for review

PeterC-DLS:
Looks okay to me. I've generated a [list](https://github.com/nexusformat/definitions/files/5820207/iri-targets.txt) of anchor references using on a local build
```
$ grep -R target\"  build/manual/build/html/classes/ | sed -r 's/^([^:]*):.*target" id="([^"]+).*/\1#\2/' | cut -d/ -f6-
```
@hgoerzig are you happy with theses?

PeterC-DLS (approved):

prjemian:
Would it be useful to have a complete list of all the anchors generated, as part of this pull request?

prjemian:
This is a nice addition that, since it publishes the anchors for all to see, completes the request.  Here is an example from `NXaperture`: 
Clipboard01

prjemian:
@PeterC-DLS : Since I have used the [pyRestTable package](https://github.com/prjemian/pyRestTable) to generate the nice table of anchors (noting that package started here in the NeXus documentation to generate such tables), it uses [python f-strings](https://realpython.com/python-f-strings/) which have a minimum requirement of Python 3.6.

OK if we drop py3.5 support now?  Otherwise need to refactor the code that writes such tables.

PeterC-DLS:
> 
> 
> @PeterC-DLS : Since I have used the [pyRestTable package](https://github.com/prjemian/pyRestTable) to generate the nice table of anchors (noting that package started here in the NeXus documentation to generate such tables), it uses [python f-strings](https://realpython.com/python-f-strings/) which have a minimum requirement of Python 3.6.
> 
> OK if we drop py3.5 support now? Otherwise need to refactor the code that writes such tables.

It's EOL so no problem.

prjemian:
I'll update the CHANGES.rst.  Feel free to merge after the CI finishes after the next commit.

prjemian:
We should drop the travis-ci integration now.  Now that we have a [Github Actions workflow](https://github.com/nexusformat/definitions/blob/master/.github/workflows/syntax-checks.yml) to check NXDL syntax, the travis-ci integration is no longer needed.

PR #883, opened 2021-02-11
Reformat long URLs in NXmx documentation

### LINKED ISSUE #882: Some links are not complete as hyperlinks
Hello,

Just found these links didn't fully function on this [page](https://manual.nexusformat.org/classes/applications/NXmx.html). An extra space seems to be the cause of the problem.

https://github.com/nexusformat/definitions/blob/96ebd852e03016490e9ab4087ce02ad683945a54/applications/NXmx.nxdl.xml#L193-L197

https://github.com/nexusformat/definitions/blob/96ebd852e03016490e9ab4087ce02ad683945a54/applications/NXmx.nxdl.xml#L860-L863

prjemian:
Reformat long URLs in NXmx documentation to fix #882.

Also added a comment to keep URLs all on one line.  In this case, there was cosmetic line-breaking that caused the URLs to not function properly in the published documentation.


.travis.yml
```diff
@@ -0,0 +1,28 @@
+# :file: .travis.yml
```
PeterC-DLS, line 1: Accidental re-add?

applications/NXmx.nxdl.xml
```diff
@@ -859,7 +860,7 @@
 					<doc>
 						Name of source.  Consistency with the naming in
 						http://mmcif.wwpdb.org/dictionaries/mmcif_pdbx_v50.dic/Items
-						/_diffrn_source.pdbx_synchrotron_site.html controlled
```
PeterC-DLS, line 862: Still incorrect as wrapped

PeterC-DLS (changes requested):

.travis.yml
prjemian: Correct.  Was in my local directory as uncommitted content.  Removing.

prjemian:
@PeterC-DLS Thanks for the review.

PeterC-DLS (approved):

PR #887, opened 2021-03-22
Clarify NXtransformations to fix #885

### LINKED ISSUE #885: Clarify what a general transformation is expected to be in NXtransformations
A "general" transformation field (i.e. one that does not have a `transformation_type` attribute) is said to be allowed for documenting important directions. Can the field have zero length? Can it have an `offset` attribute or a `depends_on` attribute?

Also for any transformation is the `vector` attribute really optional?

PeterC-DLS:
Make the vector attribute mandatory, add note on offset usage.

### DIFF
--- a/base_classes/NXtransformations.nxdl.xml
+++ b/base_classes/NXtransformations.nxdl.xml
@@ -40,7 +40,7 @@
 		transformations are done, or in documenting important directions, such as the
 		direction of gravity.
 
-		A nested sequence of transformations lists the offset and rotation steps
+		A nested sequence of transformations lists the translation and rotation steps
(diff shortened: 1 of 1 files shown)


prjemian (approved):
LGTM.  

**observation**: 
As `NXtransformations` replaces `NXgeometry`, there will be increased use in describing a geometry that is virtual and not based on hardware.  The documentation refers to hardware (such as this phrase: _... the mechanical offset from mounting the axis ..._) as if the only use of NXtransformations is to describe orientations of hardware.  Perhaps the use of `NXtransformations` to describe dependencies in a virtual geometry should be considered in the documentation.

Issue #412, opened 2015-06-17
Full documentation of NXdetector and related classes


eugenwintersberger:
We have currently no reasonable documentation for NXdetector and related classes. In addition some fields in the NXdetector documentation cannot be described properly within the NXDL file as sketches would be required to make the text unambiguous.  

I will start working on this in a new branch related to this issue.

zjttoefs:
Sketches would help.

But the branch is stale. Is it worth keeping?

zjttoefs:
Dropping this for now. A more visual explanation is still a good idea, if someone wants to volunteer.

prjemian:
I will recover the drawings that were provided in hopes that they can be used.  And then close the associated [branch](https://github.com/nexusformat/definitions/tree/issue_412).

prjemian:
@eugenwintersberger : Added your drawings and documentation into master but have not merged the changes into `design.rst` yet.  Would you like to complete this work?

prjemian:
Any volunteers?

phyy-nx:
Proposal if for @yayahjb and myself to review NXdetector from the context of the HDRMX and Gold Standard paper

yayahjb:
We will be working in
  (https://github.com/nexusformat/definitions/blob/412-ff-NXdetector-documentation/manual/source/_static/412-detector-docs/detectors.txt)

for which I have created the new branch 412-ff-NXdetector-documentation to continue the work begun by Eugen and Pete.  We need to add in all the relevant information from the Gold Standard paper into detectors.txt.  Then that file will need to be added into design.rst.  For the 2020 code camp we should make enough of a start on updating detectors,txt so that others from outside of the synchrotron and XFEL worlds can comment and extend the work.   This is a big job, which will need a lot more drawings.  There will be HDRMX e-meetings between now and summer 2021.  If at all possible, we should try to have an agreed text in time for the summer 2021 ACA and IUCr XXV meetings.
  Regards,
    Herbert

phyy-nx:
Feature request: add depends_on as field in NXdetector_module.  See https://github.com/nexusformat/cnxvalidate/issues/27#issuecomment-712906720

phyy-nx:
Another thing to consider.  `gain_setting` is available as enumeration in NXdetector:
```
  <field name="gain_setting" type="NX_CHAR">
    <doc>The gain setting of the detector. This influences background etc.</doc>
    <enumeration>
      <item value="high"/>
      <item value="standard"/>
[...]

PR #897, opened 2021-03-23
Issue 681

mkoennecke:
This is the first version of adding NXtransformations, depends_on and NXoff_geometry to all base classes needing them.

### DIFF
--- a/base_classes/NXaperture.nxdl.xml
+++ b/base_classes/NXaperture.nxdl.xml
@@ -27,12 +27,13 @@
     name="NXaperture" 
     type="group" extends="NXobject">
 
-    <doc>A beamline aperture.</doc>
+    <doc>A beamline aperture. This group is deprecated, use NXslit instead.</doc>
     <!-- TODO compare with "screens" in SHADOW -->
 
     <group type="NXgeometry">
         <doc>
-            location and shape of aperture
+            location and shape of aperture. The use of this group is deprecated, use 
+	    depends_on and NXtransformations instead.
             
             .. TODO: documentation needs improvement, contributions welcome
             
@@ -44,7 +45,28 @@
         </doc>
     </group>
     <group name="BLADE_GEOMETRY" type="NXgeometry">
-        <doc>location and shape of each blade</doc>
+        <doc>location and shape of each blade. The use of this group is depreacted. 
+	Use NXoff_geometry instead in order to describe the shape of the aperture.</doc>
+    </group>
+    <group type="NXoff_geometry" minOccurs="0">
+      This group contains the geometry description of the aperture and replaces BLADE_GEOMETRY.
+    </group>
+    <field name="depends_on" type="NX_CHAR">
+      <doc>
+    NeXus positions components by applying a set of translations and rotations 
+    to apply to the component starting from 0, 0, 0. The order of these operations 
+    is critical and form a what NeXus calls a dependency chain. The depends_on 
+    field defines the path to the top most operation of the dependency chain. 
+    Usually these operations are stored in a NXtransformations group. But NeXus 
+    allows them to be stored anywhere.
+      </doc>
+    </field>
+    <group type="NXtransformations" minOccurs="0">
+      <doc>
+    This is the group recommended for holding the chain of translation 
+    and rotation operations necessary to position the component within
+    the instrument. 
+      </doc>
     </group>
     <field name="material"><!-- TODO Uniformity problem, "type" is used elsewhere for same context -->
         <doc>Absorbing material of the aperture</doc>
--- a/base_classes/NXaperture.nxdl.xml
+++ b/base_classes/NXaperture.nxdl.xml
@@ -30,23 +30,13 @@
     <doc>A beamline aperture. This group is deprecated, use NXslit instead.</doc>
     <!-- TODO compare with "screens" in SHADOW -->
 
-    <group type="NXgeometry">
+    <group type="NXgeometry" deprecated="Use NXtransformations for positioning the aperture and NXoff_geometry to describe its shape">
         <doc>
-            location and shape of aperture. The use of this group is deprecated, use 
-	    depends_on and NXtransformations instead.
-            
-            .. TODO: documentation needs improvement, contributions welcome
-            
-                * description of terms is poor and leaves much to interpretation
-                * Describe what is meant by translation _here_ and ...
-                * Similar throughout base classes
-                * Some base classes do this much better
-                * Such as where is the gap written?
+            location and shape of aperture.
         </doc>
     </group>
-    <group name="BLADE_GEOMETRY" type="NXgeometry">
-        <doc>location and shape of each blade. The use of this group is depreacted. 
-	Use NXoff_geometry instead in order to describe the shape of the aperture.</doc>
+    <group name="BLADE_GEOMETRY" type="NXgeometry" deprecated="Use NXoff_geometry instead to describe the shape of the aperture">
+        <doc>location and shape of each blade</doc>
     </group>
     <group type="NXoff_geometry" minOccurs="0">
       This group contains the geometry description of the aperture and replaces BLADE_GEOMETRY.
--- a/base_classes/NXattenuator.nxdl.xml
+++ b/base_classes/NXattenuator.nxdl.xml
@@ -43,9 +43,27 @@
         TODO  SHADOW: http://ftp.esrf.eu/pub/scisoft/shadow       or (older)   http://www.nanotech.wisc.edu/shadow  
     -->
 
-    <field name="distance" type="NX_FLOAT" units="NX_LENGTH">
+    <field name="distance" type="NX_FLOAT" units="NX_LENGTH" deprecated="Use transformations/NXtransformations/distance instead">
         <doc>Distance from sample</doc>
     </field>
+<field name="depends_on" type="NX_CHAR">
+  <doc>
+    NeXus positions components by applying a set of translations and rotations 
+    to apply to the component starting from 0, 0, 0. The order of these operations 
+    is critical and form a what NeXus calls a dependency chain. The depends_on 
+    field defines the path to the top most operation of the dependency chain. 
+    Usually these operations are stored in a NXtransformations group. But NeXus 
+    allows them to be stored anywhere.
+  </doc>
+</field>
+
+<group name="transformations" type="NXtransformations" minOccurs="0">
+  <doc>
+    This is the group recommended for holding the chain of translation 
+    and rotation operations necessary to position the component within
+    the instrument. 
+  </doc>
+</group>
     <field name="type">
         <doc>Type or composition of attenuator, e.g. polythene</doc>
     </field>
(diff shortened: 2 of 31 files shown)


base_classes/NXaperture.nxdl.xml
```diff
@@ -27,25 +27,37 @@
...
-                * Some base classes do this much better
-                * Such as where is the gap written?
+            location and shape of aperture.
         </doc>
     </group>
-    <group name="BLADE_GEOMETRY" type="NXgeometry">
+    <group name="BLADE_GEOMETRY" type="NXgeometry" deprecated="Use NXoff_geometry instead to describe the shape of the aperture">
         <doc>location and shape of each blade</doc>
     </group>
+    <group type="NXoff_geometry" minOccurs="0">
```
PeterC-DLS, line 41: If this class is to be deprecated then new additions should be avoided.

base_classes/NXattenuator.nxdl.xml
```diff
@@ -43,9 +43,27 @@
         TODO  SHADOW: http://ftp.esrf.eu/pub/scisoft/shadow       or (older)   http://www.nanotech.wisc.edu/shadow  
     -->
 
-    <field name="distance" type="NX_FLOAT" units="NX_LENGTH">
+    <field name="distance" type="NX_FLOAT" units="NX_LENGTH" deprecated="Use transformations/NXtransformations/distance instead">
```
PeterC-DLS, line 46: The suggested form is ambiguous with those slashes. Maybe use `transformations:NXtransformations` to indicate that the group's name and NeXus class.

base_classes/NXattenuator.nxdl.xml
```diff
@@ -43,9 +43,27 @@
...
+    NeXus positions components by applying a set of translations and rotations 
+    to apply to the component starting from 0, 0, 0. The order of these operations 
+    is critical and form a what NeXus calls a dependency chain. The depends_on 
+    field defines the path to the top most operation of the dependency chain. 
+    Usually these operations are stored in a NXtransformations group. But NeXus 
+    allows them to be stored anywhere.
+  </doc>
+</field>
+
+<group name="transformations" type="NXtransformations" minOccurs="0">
```
PeterC-DLS, line 60: Is this name fixed?

PeterC-DLS (changes requested):

base_classes/NXattenuator.nxdl.xml
mkoennecke: I thought it is a good idea to have that in the same place in each group. Is there a use case for another name?

matthew-d-jones: The name is not important from a parsing perspective because you are always following a `depends_on` path to find it anyway. I also can't think of a use case for allowing a different name though.

base_classes/NXaperture.nxdl.xml
```diff
@@ -27,25 +27,39 @@
...
+    field defines the path to the top most operation of the dependency chain. 
+    Usually these operations are stored in a NXtransformations group. But NeXus 
+    allows them to be stored anywhere.
+      </doc>
+    </field>
+    <group type="NXtransformations" minOccurs="0">
+      <doc>
+    This is the group recommended for holding the chain of translation 
+    and rotation operations necessary to position the component within
+    the instrument.
```
matthew-d-jones, line 60: Maybe worth adding something like "The chain may traverse similar groups in other component classes"? The current wording might imply that the full chain for this component must be inside this group.

base_classes/NXbeam_stop.nxdl.xml
```diff
@@ -43,16 +43,42 @@
...
+</field>
+
+<group name="transformations" type="NXtransformations" minOccurs="0">
+  <doc>
+    This is the group recommended for holding the chain of translation 
+    and rotation operations necessary to position the component within
+    the instrument. 
+  </doc>
+</group>
+<group type="NXoff_geometry" minOccurs="0">
```
matthew-d-jones, line 64: Or "NXcylindrical_geometry"

base_classes/NXtransformations.nxdl.xml
```diff
@@ -92,6 +92,21 @@
...
+		as needed in order to position the component. Feel free to add more axis if required. In the description 
+		given below, only those atttributes which are defined through the name are spcified. Add the other attributes 
+		of the full set:
+
+		* vector
+		* offset
+		* transformation_type
+		* depends_on
+      
+		as needed.
```
matthew-d-jones, line 109: Was it agreed by NIAC to prefer using these names? I sometimes use the name of a motion axis device for example.
Should we also explicitly mention that an `NXlog` can be used in place of a dataset?

mkoennecke: ? These are the standard transformation related attribute types.

At the beginning of the code camp we discussed this issue and decided to move all the older NeXus 
positioning fields into NXtransformations. The use case you describe, an axis name called hugo with a very weird operation, is still possible.

The NXlog is a good point and I see where I can fit this in.

matthew-d-jones: Sorry, by "these names" I meant "distance", "height", etc.

base_classes/NXdetector.nxdl.xml
```diff
@@ -804,6 +804,24 @@
       </doc>
     </group>
   </choice>
+<field name="depends_on" type="NX_CHAR">
+  <doc>
+    NeXus positions components by applying a set of translations and rotations 
+    to apply to the component starting from 0, 0, 0. The order of these operations 
+    is critical and form a what NeXus calls a dependency chain. The depends_on
```
phyy-nx, line 811: Should read "and forms what NeXus..."

base_classes/NXattenuator.nxdl.xml
phyy-nx: @prjemian Can you chime in on this?  From session G of CodeCamp2021, we are split on this issue.  Whether to specify the name as transformations or leave it optional.  (Note, nobody had a really strong opinion :)

prjemian: IMO, not sure why it is necessary to require the exact name.  Within the BES Bluesky communitym there is a group looking at XPCS data and has begun learning how to apply NX transformations.  They have assumed the name is not required to be exactly `transformations`.  Thus, in any NXDL spec, the name could be omitted unless there are two or more NXtransformations groups in the same directory.  In such case, they need names to avoid ambiguity.

base_classes/NXtransformations.nxdl.xml
```diff
@@ -176,6 +192,207 @@
...
+    <field name="distance" type="NX_FLOAT" units="NX_LENGTH" minOccurs="0">
+      <doc>
+	This is the distance to the previous component in the
+	instrument. 
+      </doc>
+      <attribute name="vector" type="NX_NUMBER">
+	<dimensions rank="1">
+	  <dim index="1" value="3" />
+	</dimensions>
+	<enumeration>
```
PeterC-DLS, line 204: This attribute is defined as having a number type thus can probably cannot be an enumeration.

PeterC-DLS: That's not quite right: arrays are not allowed as enumeration items.

base_classes/NXtransformations.nxdl.xml
```diff
@@ -176,6 +192,207 @@
 			differences.  Use of ``AXISNAME_end`` is recommended.
 		</doc>
 	</field>
+    <field name="distance" type="NX_FLOAT" units="NX_LENGTH" minOccurs="0">
```
PeterC-DLS, line 195: This fixes the coordinate frame's z axis direction. I think the frame should be left as a choice for beamlines/facilities.

base_classes/NXattenuator.nxdl.xml
yayahjb: For the crystallography community, prescriptive naming is not likely to work.

On the matter of deprecating distance in favor of deriving it from the axis chain, that also is not likely to work, since distance, wavelength and beam center are essential data for
processing.  The best we are likely to be able t do is to make it clear when they are derived versus when they are directly observed and make derived the default.

yayahjb (commented):
See comments on prescriptive naming and keeping explicit distance, beam center and wavelength, and note that the enumeration issue still needs to be resolved, but all that being said, it really is time to make use of NXtranformations the norm.

phyy-nx:
I believe that what @mkoennecke has done is to add depends_on and NXtransformations where they were missing, which I'm quite pleased about.  However, as discussed in one of the code camp sessions, going further and deprecating distance and beam center fields is likely not necessary.  I've reverted the deprectations and replaced them with comments here:

https://github.com/nexusformat/definitions/compare/issue-681...issue-681-aaron

Further, while I was working on this, I see that @mkoennecke added the deprecated fields explicitly to NXtransformations as example names for beamline components.  For example, compare this removal of [distance](https://github.com/nexusformat/definitions/pull/897/files#diff-958495767f26dd709fe7d1ab478f93f1183a252690722aeed15013b7b9f4ec81R46) to where it was [added as an example axis name](https://github.com/nexusformat/definitions/pull/897/files#diff-d5885ae955bdaadaf3e76bec7bfc38ffa4225cb3072708adaeee02116912726bR197) to NXtransformations.   

I think the additions to NXtransformations aren't needed in formal NXdl language.  I'd instead like to see examples as text, such as what @yayahjb and I are hoping to do when resolving #412.  That would make it more clear that axis names and conventions aren't proscriptive.  But as that's more work, I'd instead propose reverting most of the changes to NXtransformations as the next step in this PR.  Would love to hear any feedback though.

Thanks!

yayahjb:
Aaron is right -- Herbert

mkoennecke:
Hmmh, IMHO we should not give up to quickly on names for transformations. I am in favor to keep them at least as suggestions. Some of them are very old names, with NeXus since its inception. These are things like distance, rotation_angle, polar_angle, height and such. Then there are names which never got popular like azimuthal_angle, meridional_angle. May be drop those. We not only have to cater for the anarchic MX community but also for people who are relatively new to NeXus and would prefer a little hand holding when choosing a name for say: a translation in x or y.

rayosborn:
The comment by @mkoennecke about deprecating `polar_angle`, etc, alarmed me. Are people actually suggesting deprecating `distance`, `polar_angle`, `azimuthal_angle`, etc, in favor of abstract `depends_on` chains? I hope not because we use them all the time. Personally, I have never used transformations and I think it would be a mistake to force everyone to use them. It is another non-intuitive aspect of the standard for people to learn, and I don't think we even provide a tutorial (please point me to one if I'm wrong). If I have got the wrong end of the stick, please ignore this comment.

prjemian:
As far as I know, there is no _tutorial_ but there is documentation in addition to what is provided by the NXtransformation base class.  The manual has a [section](https://manual.nexusformat.org/design.html#coordinate-transformation-field-and-attributes) with one example.  More examples would be very useful.

yayahjb:
I think Aaron is right that we should put most of this aside for the moment
and talk it through slowly and carefully.

mkoennecke:
I have updated this pull request by merging in aaron's changes which remove most of the deprecated messages and merged with main. Please review again.

phyy-nx:
I am working on a unified PR for #897 and #999

Issue #899, opened 2021-03-23
Create a requirements.txt file for package installation in non-conda virtual environments


prjemian:
Start with `environment.yml`.  Related to #854 (tabs for language examples) and #870 (switch CI from using conda).

prjemian:
Right. The PR referenced this issue with `see #899`. That text did not trigger this issue to close automatically when the PR was merged.

PR #902, opened 2021-03-23
Add requirements.txt to automated build

FreddieAkeroyd:
This allows specification of items needed for python venv on ISIS doc build server
 
See #899

It is possible to use a `requirements.txt` from a conda `environment.yml`
 
```
name: test-env
dependencies:
  - python>=3.5
  - anaconda
  - pip
  - pip:
    - -r file:requirements.txt
```


FreddieAkeroyd:
have reduced `requirements.txt` to same 4 entries as `environment.yml` and it works fine

jenkins_build
```diff
@@ -4,10 +4,16 @@ set -o errexit -x
...
 
-export PATH=/usr/local/bin:$PATH
-
-# use local python venv
-. /usr/local/python/bin/activate
+#export PATH=/usr/local/bin:$PATH
+
+## install local python virtual environment
+## based on local requirements.txt
+PYDIR=`pwd`/python
```
prjemian, line 11: PYDIR=$(pwd)/python

prjemian: This is a standard recommendation from Codacy's bash script analyses

FreddieAkeroyd: does it show that I learnt my sh programming before that syntax was invented :-D

prjemian (commented):
Lots of items in the requirements, many of which are probably required by a smaller number of packages we name here.  pip will examine the requirements of each package named in this file. This list may need to be pruned later as the packages we _actually_ need update their requirements.  If it works now, we can defer the pruning actions to a later issue.

FreddieAkeroyd:
I pruned `requirements.txt` and it still seems to work on our local build server

prjemian (approved):
LGTM

Issue #590, opened 2017-08-14
NXdetector seem to require 4 dimensional data


zjttoefs:
The [manual](http://download.nexusformat.org/sphinx/classes/base_classes/NXdetector.html) renders as:

`data[np, i, j, tof]: (optional) NX_NUMBER {units=NX_ANY}`

To the newcomer it is not obvious how this should be interpreted. The x and y pixel offsets also require 2 dimensions according to the XML, but in the text they say **can** be multidimensional, if needed. 

There could be a whole host of XSD magic with `<choice>`, but getting any references right will be a mess. 

What to do?

mkoennecke:
Telco comment: improve documentation

prjemian:
@zjttoefs @mkoennecke Who will do this work?

zjttoefs:
I don't have a good solution. Otherwise I would do something.

prjemian:
This may be closed as "unresolved" due to lack of a sponsor

zjttoefs:
That seems a bit unfair, given that we have 47 tickets in the backlog, some untouched for years.

prjemian:
We should add words that assert that the definition is more flexible and more than one rank is allowed.  This could be presented in the description and *also* in the documentation for the **data** field.

prjemian:
Needs English wordsmithing to make this clear.  Perhaps other definitions  could benefit also.

zjttoefs:
As mentioned in the referenced ticket this can also have validation effects.

prjemian:
Could this be coupled with #412?

phyy-nx:
Yup, @yayahjb and I will work on this.

woutdenolf:
The new rendering of the dimensions field (PR https://github.com/nexusformat/definitions/pull/1086) may help. For this particular field the old rending looks like this

![image](https://user-images.githubusercontent.com/7264703/173606578-c32eb33e-720a-4c32-9862-9742d1abf552.png)

the new rendering (still in review) looks like this

![image](https://user-images.githubusercontent.com/7264703/173606909-16fed969-3d88-4bc1-b67a-9c23031d65c7.png)

PR #903, opened 2021-03-23
Updated the documentation of NXdetector:data Refs #590

mkoennecke:


base_classes/NXdetector.nxdl.xml
```diff
@@ -89,7 +89,28 @@
   </field>
 
   <field name="data" type="NX_NUMBER" units="NX_ANY">
-    <doc>Data values from the detector.</doc>
+    <doc>
+      Data values from the detector. The base dimensionality of this data is detector dependent.
+      If it is only a single value, it is just one value. If it is a linear detector, it is one dimensional
+      with the  number of pixels in the linear detector as dimension, [npixel].  
+      
+      If it is an image, then it is 2D, [xdim, ydim]. If the detector is three dimensional, then the
```
phyy-nx, line 97: Should we use fast/slow language instead of xdim, ydim?

mkoennecke: xdim and ydim sound more familiar to me; fast and slow is another abstraction level which does help when talking about storage order but not here. My 2c worth.

phyy-nx: It's not really xdim/ydim as names, it's the order.  In many datasets the order is [ydim,xdim], assuming a single panel detector aligned to the lab axes.  If instead it were [slow,fast] it would be unambiguous.  I think x/y should be concepts for NXtransformations, and fast/slow should be concepts for data.

phyy-nx:
I think this and #1055 conflict?

prjemian:
> I think this and #1055 conflict?

#1055 is the PR for this issue.

benajamin:
Changes moved to #1055 so this PR is no longer needed.

Issue #912, opened 2021-03-29
Validators should warn (but accept) same spelling but wrong case of NXDL specifications.


prjemian:
To help with #910 and #911, we could allow any representation (UPPER/lower case) of the NXDL class name given in the `definition` field to pass validation.  Those names that do not match exactly (matching the value given in the `name` attribute of the `definition` element) should be flagged as a warning but not marked as invalid.

The NIAC must ensure we do not allow multiple *different* NXDL classes that have the same name, differing *only* in case.

_Originally posted by @prjemian in https://github.com/nexusformat/definitions/issues/910#issuecomment-809520601_

prjemian:
## `definition` element:

https://github.com/nexusformat/definitions/blob/a232ead018f66d3cf973a2398785c25451f93406/applications/NXspe.nxdl.xml#L24-L31

## `definition` field 

https://github.com/nexusformat/definitions/blob/a232ead018f66d3cf973a2398785c25451f93406/applications/NXspe.nxdl.xml#L35-L42

prjemian:
The manual should include this advice about data file validation.

PR #913, opened 2021-03-29
removed NX_SPE from the definition fields enumeration list

### LINKED ISSUE #911: NXspe: specifies an invalid enumeration option of NX_SPE for the definition field
in order for a string to be valid as an enumeration in the **definition** field it must (as described) be a name that points to a valid NXDL schema file, there is no such application definition file _NXSPE.nxdl.xml_ therefor **NXSPE** needs to be removed from the definition enumeration list.
The current enumeration list for the field **definition**
```
            <enumeration>
                <item value="NXSPE"/> <--- to be removed
                <item value="NXspe"/>                
            </enumeration>
```

RussBerg:
removed the invalid NX_SPE from the definition field enumeration list because it does not point to a valid NXDL schema, fixes #911 :

```
            <enumeration>
                <item value="NXSPE"/>
                <item value="NXspe"/>                
            </enumeration>
```
is now 
```
            <enumeration>
                <item value="NXspe"/>                
            </enumeration>
```

### DIFF
--- a/applications/NXspe.nxdl.xml
+++ b/applications/NXspe.nxdl.xml
@@ -36,8 +36,7 @@
             <doc>Official NeXus NXDL schema to which this file conforms.</doc>
             <attribute name="version"/>
             <enumeration>
-                <item value="NXSPE"/>
-                <item value="NXspe"/>                
+                <item value="NXspe"/>
             </enumeration>
         </field>
         <group type="NXcollection" name="NXSPE_info">


stuartcampbell:
I think I added both of these because the Mantid writer for `nxspe` files writes the definition in uppercase (https://github.com/mantidproject/mantid/blob/master/Framework/DataHandling/src/SaveNXSPE.cpp#L103).

FreddieAkeroyd:
Shall I ask Mantid if they can change their code to write `NXspe` ?

prjemian (approved):
Exactly.

prjemian:
> Shall I ask Mantid if they can change their code to write `NXspe` ?

Please.

prjemian:
No NIAC review needed, just a bug fix.

PeterC-DLS:
Fixes #912

PR #927, opened 2021-05-05
Add geometry information to NXbeam for #925

### LINKED ISSUE #925: Add geometry information to NXbeam
NeXus has a [default coordinate frame](https://manual.nexusformat.org/design.html#nexus-geometry): its origin is at the centre of the scattering volume in the sample with the beam direction along positive `z` and gravity along negative `y`.

For diffraction and scattering experiments, there are many choices of coordinate frame: for example, the 'You' geometry in [diffcalc](https://diffcalc.readthedocs.io/en/latest/youmanual.html#introduction) or the six-circle diffractometer in [PyMCA](https://github.com/vasole/pymca/blob/master/PyMca5/PyMcaPhysics/SixCircle.py).

Adding a `direction` field to NXbeam will allow the `q` vector to be computed for any pixel on a detector from a properly configured (conforming to the `NXmx` application definition) `NXdetector` group  for any other choice of coordinate frame subject to the `NXdetector_module` group having consistent `vector` attributes in its `fast_pixel_direction`, `slow_pixel_direction`, and `module_offset` fields.

PeterC-DLS:
Fixes #925. Add optional depends_on field and NXtransformations group to override the default choice of the beam direction along positive z and passing through the origin

Also closes https://github.com/nexusformat/NIAC/issues/115

### DIFF
--- a/base_classes/NXbeam.nxdl.xml
+++ b/base_classes/NXbeam.nxdl.xml
@@ -142,4 +142,44 @@
             for a summary of the discussion.
         </doc>
     </attribute>
+
+    <field name="depends_on" minOccurs="0" type="NX_CHAR">
+        <doc>
+            NeXus path to the beam direction (and location) at the sample.
+        </doc>
+    </field>
+
+    <group type="NXtransformations" minOccurs="0">
+        <doc>
+            Direction (and location) for the beam. The location of the beam can be given by
+            any point which it passes through as its offset attribute.
+        </doc>
+        <field name="DIRECTION" nameType="any" units="NX_TRANSFORMATION" type="NX_NUMBER">
+            <doc>
+                Direction of beam vector, its value is ignored. If missing, then the beam direction is defined as [0,0,1]
+                and passes through the origin
+            </doc>
+            <attribute name="transformation_type">
+                <enumeration>
+                    <item value="translation" />
+                </enumeration>
+            </attribute>
+            <attribute name="vector" type="NX_NUMBER">
+                <doc>
+                    Three values that define the direction of beam vector
+                </doc>
+            </attribute>
+            <attribute name="offset" type="NX_NUMBER">
+                <doc>
+                    Three values that define the location of a point through which the beam passes
+                </doc>
+            </attribute>
+            <attribute name="depends_on" type="NX_CHAR">
+                <doc>
+                    Points to the path to a field defining the location on which this
+                    depends or the string "." for origin.
+                </doc>
+            </attribute>
+        </field>
+    </group>
 </definition>
--- a/base_classes/NXbeam.nxdl.xml
+++ b/base_classes/NXbeam.nxdl.xml
@@ -181,5 +181,33 @@
                 </doc>
             </attribute>
         </field>
+        <field name="reference_plane" nameType="any" units="NX_TRANSFORMATION" type="NX_NUMBER">
+            <doc>
+                Direction of normal to reference plane used to measure azimuth relative to the beam, its value is ignored.
+                This also defines the parallel and perpendicular components of the beam's polarization.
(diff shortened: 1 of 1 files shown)


base_classes/NXbeam.nxdl.xml
```diff
@@ -142,4 +142,72 @@
...
+                    depends or the string "." for origin.
+                </doc>
+            </attribute>
+        </field>
+        <field name="reference_plane" nameType="any" units="NX_TRANSFORMATION" type="NX_NUMBER">
+            <doc>
+                Direction of normal to reference plane used to measure azimuth relative to the beam, its value is ignored.
+                This also defines the parallel and perpendicular components of the beam's polarization.
+                If missing, then the reference plane normal is defined as [0,1,0] and passes through the origin
+            </doc>
```
phyy-nx, line 189: Is this redundant with the polarization stuff already in NXbeam, or proposed to be added in #858?  Or is it an alternate way to specify polarization?

benajamin:
The text seems to assume that NXbeam will only be used to describe the beam at the sample, but my understanding is that it can be used to describe the beam at any location (making it useful for recording ray tracing simulations of beamlines). Much of the added detail doesn't make sense to me, but maybe I'm just missing the point?

base_classes/NXbeam.nxdl.xml
PeterC-DLS: This is needed if you are not using the NeXus coordinate frame or beam is not along `z`-axis as the Q/U components are measured transverse to the beam.

PeterC-DLS:
Not soley at the sample so I will remove that...

yayahjb:
After a long discussion with Peter Chang and a side message to Aaron, out
of respect
for a request by Pete Jemian to put my comments in the record, here are my
comments.

The beam direction is a critical component of the McStas coordinate system.
Gravity is the other.  To have a coordinate system, you either need to
specify
the handedness or give a third axis.

Peter Chang has asserted that just changing the beam direction by itself is
sufficient and that the remaining essential metadata can be inferred from
elsewhere in the file.  I believe that would be a mistake and that in _any_
place in a NeXus files where a change to the specification or either
the beam direction or the direction of gravity is given the corresponding
change to the other should be given along with the handedness to
avoid serious mistakes, such as getting beam center locations on detector
faces wrong.  You really should always be clear on what coordinate frame
you are using.

P.S. Yes,  I am literally a belt and suspenders type!!!

base_classes/NXbeam.nxdl.xml
```diff
@@ -142,4 +142,72 @@
             for a summary of the discussion.
         </doc>
     </attribute>
+
+    <field name="depends_on" minOccurs="0" type="NX_CHAR">
+        <doc>
+            NeXus path to the beam direction (and location) at beamline component or sample
+        </doc>
```
phyy-nx, line 149: Proposed revision, workshopped in the code camp.  Instead of "NeXus path to the beam direction (and location) at beamline component or sample", we could use the following:

The NeXus coordinate system defines the Z axis to be along the nominal beam direction. This is the same as the McStas coordinate system (see :ref:`Design-CoordinateSystem`). However, the additional transformations needed to represent an altered beam direction can be provided using this depends_on field that contains the path to a NXtransformations group. This could represent redirection of the beam, or a refined beam direction.

benajamin: Waiting for reply from @PeterC-DLS

PeterC-DLS: Adding a transformation matrix back to McStats (somewhere in `NXentry`) is fine by me and a reminder note here.

benajamin (approved):

phyy-nx (approved):

PeterC-DLS:
Fixes #925

Issue #934, opened 2021-07-01
A new type of NXsource


timsnow:
In recent years it has become possible to purchase MetalJet x-ray sources, such as those made by Excillum.

The current list of source types, as given in the enumerated list in NXsource, doesn't include an entry for MetalJet sources.

Would it be possible to add MetalJet to the list of source types for NXsource please, I shall submit a pull request shortly with the change in place.

benajamin:
We just discussed this in the July NeXus Telco. It is certainly possible to add a new `NXsource` type, but I question whether "metal jet" is too specific. A ["nitrogen jet" type X-ray source](https://doi.org/10.1063/1.1884186) is also existing. Is this similar enough to combine them into one category (e.g. "liquid jet") that are further differentiated by the value of `target_material`?

PR #935, opened 2021-07-01
Suggested change to NXsource - Issue #934

timsnow:
Addition of X-ray MetalJet to the enumerated list of source types within NXsource


PeterC-DLS:
Can I suggest "Metal Jet X-ray" to follow other types?

yayahjb:
Good idea.

PeterC-DLS (approved):

PR #951, opened 2021-12-09
repair workflow and make Python code compliant with py3

### LINKED ISSUE #946: documentation build requires python2
Building the docs with Python 3.9.5 results in an error early on in nxdl_desc2rst.py:

>     raise "Something wrong with indentation on this line:\n" + line

This results in the following error:

>    TypeError: exceptions must derive from BaseException

It runs fine in python2, so I suspect that to be the issue. I might be wrong, though.
Would be nice to this migrate to supported python in the near future. 

Replacing the `raise` statement above with a `pass` builds successful with the newer python. So at least for the success path, the work required will be low.

### LINKED ISSUE #950: The workflow is failing, even when restarted.
The workflow is failing, even when [restarted](https://github.com/nexusformat/definitions/runs/4460007843?check_suite_focus=true).  This error is reported:

    AttributeError: 'FileNotFoundError' object has no attribute 'read'

Same is found when trying locally now (to create a custom conda environment to apply syntax checks, same as worflow).  The problem seems to be: https://github.com/nexusformat/definitions/blob/493adbeaa6a66f24b15c0882ca4c6594b0692f5f/environment.yml#L8-L9

_Originally posted by @prjemian in https://github.com/nexusformat/definitions/issues/948#issuecomment-988982515_

prjemian:
- FIX #946
- FIX #950

### DIFF
--- a/.github/env-workflow.yml
+++ b/.github/env-workflow.yml
@@ -1,0 +1,12 @@
+name: nexusformat
+
+channels:
+  - defaults
+  - conda-forge
+  - aps-anl-tag
+
+dependencies:
+  - lxml
+  - pyRestTable
+  - Sphinx
+  - six
--- a/.github/workflows/syntax-checks.yml
+++ b/.github/workflows/syntax-checks.yml
@@ -40,8 +40,7 @@ jobs:
 
     - name: Install dependencies
       run: |
-        $CONDA/bin/conda env update --file environment.yml --name $ENV_NAME
-        # $CONDA/bin/conda install pytest --name $ENV_NAME
+        $CONDA/bin/conda env update --file .github/env-workflow.yml --name $ENV_NAME
         $CONDA/bin/conda list -r --name $ENV_NAME
       env:
         PY_VER: ${{ matrix.python-version }}
--- a/nxdl.xsd
+++ b/nxdl.xsd
@@ -1182,7 +1182,7 @@
 			<xs:documentation>
 				An ``enumeration`` restricts the values allowed for a specification.
 				Each value is specified using an ``item`` element, such as:
-				``<item value="Synchrotron X-ray Source" />``.
+				``&lt;item value="Synchrotron X-ray Source" />``.
 				Could contain these elements:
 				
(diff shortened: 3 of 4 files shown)


prjemian:
Now, with #946 and #950 repaired, the [workflow log](https://github.com/nexusformat/definitions/runs/4475492301?check_suite_focus=true) shows there is an indentation error in the 51 use XML entity`nxdl.xsd` file.

```
IndentationError: Something wrong with indentation on this line:
			</xs:documentation>
```

The code is now having troubles resolving and removing the indentation of the embedded rst documentation.  Time to refactor to [`textwrap.dedent()`](https://stackoverflow.com/a/1412728/1046449) from the Python standard library.

prjemian:
Also, this code `lxml.etree.tostring(docnodes[0], pretty_print=True)` now returns the formatted text starting from `docnodes[0]` to the *end of the `nxdl.xsd` file!*  That needs a fix (add `method="text"` kwarg) described in the [lxml tutorial](https://lxml.de/tutorial.html).

prjemian:
@PeterC-DLS : please review

The workflow is now successful, after fixing the environment file and replacing our/my complicated (and failing) Python code that removes common indentation with similar capability from the Python standard library package [`textwrap`](https://docs.python.org/3/library/textwrap.html#textwrap.dedent).

@zjttoefs : please review
To make this workflow successful, it was necessary to resolve #946 as part of this PR.  But that issue is not completely resolved since it is not possible (as [demonstrated](https://github.com/nexusformat/definitions/issues/946#issuecomment-991099115)) to add a CI workflow to test with py2.7.  So, I proposed in #952 to stop trying to support Py2.7 (actually, all Python versions <3.7).

PeterC-DLS (approved):

zjttoefs (approved):
Looks good. Tried it locally and it was able to generate everything using python 3.9. Dropping support for 2.x Python makes sense.

prjemian:
@PeterC-DLS @zjttoefs : Thanks for the quick response!

Issue #944, opened 2021-11-15
Allow regions of interest to be recorded


PeterC-DLS:
In investigating #939, it became apparent that the sentence from [NXdetector](https://github.com/nexusformat/definitions/blob/493adbeaa6a66f24b15c0882ca4c6594b0692f5f/base_classes/NXdetector.nxdl.xml#L774):
> Use one or more instances of the NXdetector_module group to declare regions of interest or some other subdivision of a detector.

does not make sense as intended as a `NXdetector_module` specifies how a subset of a detector's data array maps to a physical rectangle with given size and orientation. A region of interest would specify a region of a detectors's data array and optionally associate some data derived from it.

To this end, a new NeXus base class is proposed as a child (or one of several children) to an `NXdetector` group:

1. @NX_class = NXregion
2. @type = "rectangular" (initially sole allowed value, other values could be proposed at later date)
3. start = starting position in data array
4. count = number of blocks or items in hyperslab selection
5. (stride) = optional striding used to downsample data (if omitted then use array of ones)
6. (block) = optional binning used to downsample data (if omitted then use array of ones)
7. (scale) = optional scaling of binned data  (if omitted then use arrays of ones)
8. (downsampled_data:NXdata)/ = optional group containing data sliced/downsampled from parent group’s data
9. (region_statistics:NXdata))/ = optional group containing any statistics derived from the region in parent group’s data
                                                 such as sum, minimum, maximum, mean, mode, median, rms, variance, etc. Where more than one
                                                 statistical dataset is written (specified with @signal) then add @auxiliary_signals listing the others.

Note, the fields for the rectangular region specifiers follow HDF5’s dataspace hyperslab parameters

PeterC-DLS:
Final draft looks like this:
![NXregion](https://user-images.githubusercontent.com/1381719/146429994-6714bd02-ded2-4c89-968b-d2f15a913c37.png)

benajamin:
The idea looks good, but some examples would be valuable to see more clearly the use-cases you are considering. It would also be good to allow a mask to define the ROI.

PR #957, opened 2021-12-15
Add a new base class to define a region for #944

PeterC-DLS:
A region specifies a subset of a dataset and can include downsampled or statistics data groups

### DIFF
--- a/base_classes/NXdetector.nxdl.xml
+++ b/base_classes/NXdetector.nxdl.xml
@@ -770,10 +770,6 @@
     <doc>
       For use in special cases where the data in NXdetector
       is represented in several parts, each with a separate geometry.
-
-      Use one or more instances of the NXdetector_module
-      group to declare regions of interest or some other
-      subdivision of a detector.
     </doc>
   </group>
   <choice name="pixel_shape">
--- a/base_classes/NXregion.nxdl.xml
+++ b/base_classes/NXregion.nxdl.xml
@@ -1,0 +1,105 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?xml-stylesheet type="text/xsl" href="nxdlformat.xsl" ?>
+<!--
+# NeXus - Neutron and X-ray Common Data Format
+# 
+# Copyright (C) 2013-2021 NeXus International Advisory Committee (NIAC)
+# 
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 3 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+#
+# For further information, see http://www.nexusformat.org
+-->
+<definition 
+	category="base" 
+	xmlns="http://definition.nexusformat.org/nxdl/3.1"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://definition.nexusformat.org/nxdl/3.1 ../nxdl.xsd"
+	name="NXregion" 
+	type="group" extends="NXobject">
+	<symbols>
+		<doc>
+			These symbols will denote the shape of the parent's data array which has rank ``S+D`` for
+			``S`` scan dimensions and ``D`` detector dimensions
+		</doc>
+		<symbol name="S"><doc>Scan rank</doc></symbol>
+		<symbol name="D"><doc>Detector rank</doc></symbol>
+	</symbols>
+	<doc>
+		Geometry and logical description of a region of detector data. When used, child group to NXdetector.
+		
+		This can be used to describe a subset of data used to create downsampled data or to derive
+		some data from that subset.
+
+		Note, the fields for the rectangular region specifiers follow HDF5’s dataspace hyperslab parameters
+		(see https://portal.hdfgroup.org/display/HDF5/H5S_SELECT_HYPERSLAB).
+	</doc>
+	<attribute name="type" optional="false">
+		<doc>
+			This is ``rectangular`` to describe the region as a hyper-rectangle in
+			the index space of its parent group's data array
+		</doc>
+		<enumeration>
+			<item value="rectangular" />
+		</enumeration>
+	</attribute>
+	<field name="start" type="NX_NUMBER" optional="false">
+		<doc>
+			The starting position for region in detector data array. The first ``S`` dimensions are omitted
+		</doc>
+		<dimensions rank="D">
+		</dimensions>
+	</field>
+	<field name="count" type="NX_INT" optional="false">
+		<doc>
+			The number of blocks or items in hyperslab selection. The first ``S`` dimensions are omitted
+		</doc>
+		<dimensions rank="D">
+		</dimensions>
+	</field>
+	<field name="stride" type="NX_INT" optional="true">
+		<doc>
+			An optional field to define striding used to downsample data. If omitted then defined as an array of ones
+		</doc>
+		<dimensions rank="D">
+		</dimensions>
+	</field>
+	<field name="block" type="NX_INT" optional="true">
+		<doc>
+			An optional field to define binning used to downsample data. If omitted then defined as an array of ones
+		</doc>
+		<dimensions rank="D">
+		</dimensions>
+	</field>
+	<field name="scale" type="NX_NUMBER" optional="true">
+		<doc>
+			An optional field to define scaling of binned data. If omitted then defined as an array of ones
+		</doc>
+		<dimensions rank="D">
+		</dimensions>
+	</field>
+	<group name="downsampled_data" type="NXdata" optional="true">
+		<doc>
+			An optional group containing data sliced/downsampled from parent group’s data
+		</doc>
+	</group>
+	<group name="region_statistics" type="NXdata" optional="true">
+		<doc>
+			An optional group containing any statistics derived from the region in parent group’s data
+			such as sum, minimum, maximum, mean, mode, median, rms, variance, etc. Where more than one
+			statistical dataset is written (specified with ``@signal``) then add ``@auxiliary_signals``
+			listing the others
+		</doc>
+	</group>
+</definition>


prjemian:
Since it has not been ratified by the NIAC, this must be relocated to contributed definitions.

prjemian:
New base classes must be ratified by the NIAC.

prjemian:
@PeterC-DLS Is this ready for review?

woutdenolf:
image
The description here gets cut off

contributed_definitions/NXregion.nxdl.xml
```diff
@@ -0,0 +1,191 @@
...
+		</doc>
+		<enumeration>
+			<item value="rectangular" />
+		</enumeration>
+	</attribute>
+	<field name="parent">
+		<doc>
+			The name of the data field or :ref:`NXdata` group in the parent group. A
+			non-signal data field in a group can be chosen by appending a ``/`` character
+			and that field's name to the group name
```
woutdenolf, line 103: Do we have other examples where a field value refers to another field more than 1 level deep? I think we have always avoided this (see for example the discussion on the step-wise @default attribute for finding default plot https://github.com/nexusformat/definitions/issues/1003).

On the other hand `depends_on` from NXtransformations allows a field reference more than 1 level deep I think. But this refers to children. For example

```xml
sample:NXsample
  depends_on="diffr/phi"
   diffr:NXtransformations
     phi[...]
     @transformation_type=rotation
     @vector=[0,1,0]
     @depends_on=...
```

while in this case the reference is first up and then down

```xml
data:NXdata
   @signal="data1"
   @auxiliary_signals=["data2"]
   data1
   data2
region:NXregion
   parent = "data/data2"
```
I think this is getting complicated. I'm not sure we should allow a reference more than one level deep. But if we do, why restrict `data` to be an `NXdata`? And do we allow more than two levels deep?

contributed_definitions/NXregion.nxdl.xml
```diff
@@ -0,0 +1,191 @@
...
+		</doc>
+	</field>
+	<field name="parent_mask">
+		<doc>
+			The name of an optional mask field in the parent group
```
woutdenolf, line 108: In contrast to `parent` we don't allow a reference more than 1 level deep?

Needs a better description of what a `mask` is. Is this an array with the same dimensions as `parent`?

contributed_definitions/NXregion.nxdl.xml
```diff
@@ -0,0 +1,191 @@
...
+	</symbols>
+	<doc>
+		Geometry and logical description of a region of data in a parent group. When used, it could
+		be a child group to, say, :ref:`NXdetector`.
+
+		This can be used to describe a subset of data used to create downsampled data or to derive
+		some data from that subset.
+
+		Note, the fields for the rectangular region specifiers follow HDF5’s dataspace hyperslab parameters
+		(see https://portal.hdfgroup.org/display/HDF5/H5S_SELECT_HYPERSLAB). Note when **block** :math:`= 1`,
```
woutdenolf, line 52: HDF5 hyperslab slicing is complicated. Isn't the [python slicing (`start`, `stop`, `step`)](https://docs.python.org/3/library/functions.html#slice) enough?

contributed_definitions/NXregion.nxdl.xml
```diff
@@ -0,0 +1,191 @@
...
+			then the blocks are contiguous; otherwise the blocks overlap.
+			If omitted then defined as an array of ones
+		</doc>
+		<dimensions rank="1">
+			<dim index="1" value="R"/>
+		</dimensions>
+	</field>
+	<field name="scale" type="NX_NUMBER">
+		<doc>
+			An optional field to define multiplication factor for scaling of blocked data.
```
woutdenolf, line 154: I don't understand what this means. Perhaps an example will help.

woutdenolf (commented):
I like it. Some things were not clear or need to be discussed (see comments). Nice work!

contributed_definitions/NXregion.nxdl.xml
PeterC-DLS: I think we can allow it to be relative to the parent group in general

contributed_definitions/NXregion.nxdl.xml
PeterC-DLS: `mask` would be like the `pixel_mask` of `NXdetector`

contributed_definitions/NXregion.nxdl.xml
PeterC-DLS: No, we need `block` as well for downsampling

contributed_definitions/NXregion.nxdl.xml
PeterC-DLS: I will add it in one of the examples. I think EPICS uses it to scale values to the type domain.

contributed_definitions/NXregion.nxdl.xml
woutdenolf: Kind of weird to have it in `NXregion`. What is the motivation?

contributed_definitions/NXregion.nxdl.xml
```diff
@@ -0,0 +1,193 @@
...
+		<dimensions rank="1">
+			<dim index="1" value="R"/>
+		</dimensions>
+	</field>
+	<field name="scale" type="NX_NUMBER">
+		<doc>
+			An optional field to define a divisor for scaling of reduced data. For example, in a
+			downsampled sum, it can reduce the maximum values to fit in the domain of the result
+			data type. In an image that is downsampled by summing 2x2 blocks, using
+			:math:`\mathrm{scale}=4' allows the result to fit in the same integer type dataset as
```
woutdenolf, line 157: The math doesn't render for me
image

contributed_definitions/NXregion.nxdl.xml
```diff
@@ -0,0 +1,193 @@
...
+			:math:`i`-th dimension, if :math:`\mathbf{block}[i] \lt \mathbf{stride}[i]`
+			then the downsampling blocks have gaps between them; when ``block`` matches ``stride``
+			then the blocks are contiguous; otherwise the blocks overlap.
+			If omitted then defined as an array of ones
+		</doc>
+		<dimensions rank="1">
+			<dim index="1" value="R"/>
+		</dimensions>
+	</field>
+	<field name="scale" type="NX_NUMBER">
```
woutdenolf, line 152: Ok so for example suppose we have an 8-bit image, we do 2x2 summing and we want the result to still be 8 bit. In that case we need `scale=4` to not have any value overflow. Is that the use case? Then why not doing 2x2 averaging instead of 2x2 summing?

Anyway, I think I understand now. Thanks!

contributed_definitions/NXregion.nxdl.xml
```diff
@@ -0,0 +1,193 @@
...
+		<doc>
+			The name of data field in the parent group or the path of a data field relative
+			to the parent group (so it could be a field in subgroup)
+		</doc>
+	</field>
+	<field name="parent_mask">
+		<doc>
+			The name of an optional mask field in the parent group whose shape matches the
+			region dimensions of the parent data field, i.e. :math:`\boldsymbol{d}`. For example,
+			this could be ``pixel_mask`` of an :ref:`NXdetector`
```
woutdenolf, line 108: An alternative way if saying this:

```
The name of an optional mask field in the parent group with
rank :math:`\boldsymbol{R}` and dimensions :math:`\boldsymbol{d}`.
For example, this could be the ``pixel_mask`` field of an :ref:`NXdetector` group.
```

As you prefer.

contributed_definitions/NXregion.nxdl.xml
```diff
@@ -0,0 +1,193 @@
...
+			the index space of its parent group's data field
+		</doc>
+		<enumeration>
+			<item value="rectangular" />
+		</enumeration>
+	</attribute>
+	<field name="parent">
+		<doc>
+			The name of data field in the parent group or the path of a data field relative
+			to the parent group (so it could be a field in subgroup)
```
woutdenolf, line 101: May add some additional clarification?

```
The name of data field in the parent group or the path of a data field relative
to the parent group (so it could be a field in a subgroup of the parent group).
The rank of this data field is **O** + **R**.
```

As you prefer.

woutdenolf (commented):
Looks good. Only one issue left: :math:`\mathrm{scale}=4' does not render.

I also made some suggestions for additional clarification. You can use them or not, as you prefer.

contributed_definitions/NXregion.nxdl.xml
```diff
@@ -0,0 +1,193 @@
...
+			This is recommended as it also defines the region rank.
+			If omitted then defined as an array of zeros
+		</doc>
+		<dimensions rank="1">
+			<dim index="1" value="R"/>
+		</dimensions>
+	</field>
+	<field name="count" type="NX_INT" recommended="true">
+		<doc>
+			The number of blocks or items in hyperslab selection.
```
prjemian, line 123: insert `the` before ` hyperslab`

contributed_definitions/NXregion.nxdl.xml
prjemian: This is the error `=4'`, should end with a backtick.

contributed_definitions/NXregion.nxdl.xml
```diff
@@ -0,0 +1,193 @@
...
+						@signal = "maximum"
+						@auxiliary_signals = "copy"
+						maximum[128,128,20]
+						copy[128,128,320]
+
+		the ``copy`` dataset selects 20 16-channel blocks that start 32 channels apart,
+		the ``maximum`` dataset will show maximum values in each 16-channel block
+		in every spectra.
+	</doc>
+	<attribute name="type" optional="false">
```
prjemian, line 89: The name of this attribute will **collide** with a different widespread use of the `<group type="...">` attribute (names a NeXus base class):

```bash
(bluesky_2022_2) prjemian@zap:~/.../NeXus/definitions$ git grep type= | grep .nxdl.xml | grep "<group " | wc -l
502
```

the first few examples:

```bash
applications/NXarchive.nxdl.xml:    <group type="NXentry" name="entry">
applications/NXarchive.nxdl.xml:      <group type="NXuser" name="user">
applications/NXarchive.nxdl.xml:      <group type="NXinstrument" name="instrument">
applications/NXarchive.nxdl.xml:        <group type="NXsource">
applications/NXarchive.nxdl.xml:      <group type="NXsample" name="sample">
applications/NXarpes.nxdl.xml:  <group type="NXentry">
applications/NXarpes.nxdl.xml:    <group type="NXinstrument">
applications/NXarpes.nxdl.xml:      <group type="NXsource">
applications/NXarpes.nxdl.xml:      <group type="NXmonochromator" name="monochromator">
applications/NXarpes.nxdl.xml:      <group type="NXdetector" name="analyser">
applications/NXarpes.nxdl.xml:    <group type="NXsample">
```

benajamin: So would `region_type` or `shape` be appropriate?

benajamin (commented):
LGTM

contributed_definitions/NXregion.nxdl.xml
PeterC-DLS: Because it could be another number and not match the size of each block. For example, see areaDetector's [NDPluginROIStat](https://areadetector.github.io/master/ADCore/NDPluginROIStat.html).

contributed_definitions/NXregion.nxdl.xml
PeterC-DLS: Thanks!

contributed_definitions/NXregion.nxdl.xml
woutdenolf: +1 for `region_type`

contributed_definitions/NXregion.nxdl.xml
```diff
@@ -151,12 +151,11 @@
...
-			divided by a different value. For example, in a downsampled sum, it can reduce
-			the maximum values to fit in the domain of the result data type. In an image that
-			is downsampled by summing 2x2 blocks, using :math:`\mathrm{scale}=4' allows the result
-			to fit in the same integer type dataset as the parent dataset.
+			An optional field to define a divisor for scaling of reduced data. For example, in a
+			downsampled sum, it can reduce the maximum values to fit in the domain of the result
+			data type. In an image that is downsampled by summing 2x2 blocks, using
+			:math:`\mathrm{scale}=4' allows the result to fit in the same integer type dataset as
+			the parent dataset.
 			If omitted then defined as an array of ones
```
woutdenolf, line 159: I believe this should be: `If omitted the scale is assumed to be 1`. You change the scale to be a scalar right?

woutdenolf (commented):
Only one question left for me.

prjemian (approved):
That resolves my comment.  OK by me.  @woutdenolf has a question remaining.

contributed_definitions/NXregion.nxdl.xml
```diff
@@ -156,7 +156,7 @@
 			data type. In an image that is downsampled by summing 2x2 blocks, using
 			:math:`\mathrm{scale}=4` allows the result to fit in the same integer type dataset as
 			the parent dataset.
-			If omitted then defined as an array of ones
+			If omitted then no scaling occurs
```
prjemian, line 159: add `.` at end of sentence.

woutdenolf (approved):
LGTM

PeterC-DLS:
Needed to rebase according to GitHub

PR #959, opened 2021-12-15
958 remove six usage

PeterC-DLS:


utils/build_preparation.py
```diff
@@ -23,9 +21,8 @@
 # Re-run this code to bring in any changed files (for incremental build)
 # Be sure to properly specify the source and target directories.
 
-from __future__ import print_function
 import os, sys
-import local_utilities
+from local_utilities import printf, replicate
```
prjemian, line 25: Is this change (and related changes below) necessary?  It's more a matter of style.  IMO, the revised code reflects what users may want to use, but in a utility or library module such as this, the previous style makes it easier to diagnose.

utils/build_preparation.py
```diff
@@ -135,22 +132,22 @@ def update(source_path, target_path):
     qualify_inputs(source_path, target_path)
     
     paths, files = get_source_items(REPLICATED_RESOURCES, source_path)
-    local_utilities.printf('source has  %d directories   and   %d files\n', len(paths), len(files))
+    printf('source has  %d directories   and   %d files\n', len(paths), len(files))
```
prjemian, line 135: Since this format ends with `\n`, could this line be replaced with:

```python
    print('source has %d directories   and   %d files', len(paths), len(files))
```

utils/build_preparation.py
```diff
@@ -135,22 +132,22 @@ def update(source_path, target_path):
...
     
     # create all the directories / subdirectories
     for source in sorted(paths):
         relative_name = source[len(source_path):].lstrip(os.sep)
         target = standardize_name(target_path, relative_name)
         if not os.path.exists(target):
-            local_utilities.printf('create directory %s\n', target)
+            printf('create directory %s\n', target)
```
prjemian, line 142: same

utils/build_preparation.py
```diff
@@ -135,22 +132,22 @@ def update(source_path, target_path):
...
             os.mkdir(target, os.stat(source_path).st_mode)
     # check if the files need to be updated
     for source in sorted(files):
         relative_name = source[len(source_path):].lstrip(os.sep)
         target = standardize_name(target_path, relative_name)
         if not identical(source, target):
-            local_utilities.printf('update file %s\n', target)
-            local_utilities.replicate(source, target)
+            printf('update file %s\n', target)
```
prjemian, line 149: same

utils/local_utilities.py
```diff
@@ -1,28 +1,26 @@
...
-# Coded for both python2 and python3.
-
 '''
 Common code for NeXus definitions Python tools
 
 ======================  ==================================
 tool                    description
 ======================  ==================================
-:meth:`printf`          wrapper for Python 2.x and 3.x
+:meth:`printf`          formatted print without newline
```
prjemian, line 9: Will changes indicated, is `printf()` still used?  (and below)

utils/build_preparation.py
prjemian: And, I see I wrote `nxdl2rst.py` in the new way.  So much for my own consistency.

utils/local_utilities.py
prjemian: Still used in `nxdl2rst.py`

utils/build_preparation.py
PeterC-DLS: I think you meant with string interpolation too

prjemian (approved):
LGTM

PR #974, opened 2022-02-02
Use `NX_NUMBER` for saturation/underload values

kalcutter:
Change the type of `saturation_value` and `underload_value` to be
`NX_NUMBER`. These values should have the same type as the pixel data
they describe. Data in `NXdata` has the type `NX_NUMBER`.


yayahjb (approved):

soph-dec:
The changes in `applications/NXmx.nxdl.xls`got lost in the merge, the same happened in #975.

Issue #943, opened 2021-09-15
NXxpcs - proposed application definition


prjemian:
The US Dept of Energy light sources have been preparing a new application definition for X-ray photon correlation spectroscopy (XPCS) for contribution to NeXus.  Progress can be viewed from the group's [project page](https://github.com/lightsources/BES-XPCS-Pilot/projects/5) on GitHub.

The NXDL file is expected to reach contribution status this month.

Related to #82

The definition is requesting (issue #932) that [`NX_COUNT`](https://github.com/lightsources/BES-XPCS-Pilot/issues/44#issuecomment-917044251) be added as a new type of units.

The definition is requesting (issue #921) that [`NX_ARBITRARY_UNITS`](https://github.com/lightsources/BES-XPCS-Pilot/issues/37) be added as a new type of units.

prjemian:
The draft NXDL specification is available now: https://github.com/lightsources/BES-XPCS-Pilot/blob/master/NeXus/NXxpcs.nxdl.xml

There is work in branches yet to merged at this time, so expect this is not the final draft.

prjemian:
NXcanSAS avoids a new unit type by use of a [named `arbitrary` units](https://github.com/nexusformat/definitions/blob/4e48fcf6a6d8e01bf995bd0ced4b4b416ddb6e80/applications/NXcanSAS.nxdl.xml#L491) value.

prjemian:
[NeXus technical committee advises](https://www.nexusformat.org/Telco_20220225.html) to use [`units="unknown"`](https://github.com/nexusformat/definitions/pull/980#issuecomment-1050925713) instead of creating a new units type.

PR #980, opened 2022-02-05
add NX_ARBITRARY_UNITS

### LINKED ISSUE #921: How to specify a field will have arbitrary units?
Propose adding `NX_ARBITRARY_UNITS` to the list of possible unit types.

prjemian:
- FIX #921
- REF #943

### DIFF
--- a/nxdlTypes.xsd
+++ b/nxdlTypes.xsd
@@ -50,6 +50,7 @@
 		<xs:union memberTypes="
 			nxdl:NX_ANGLE
 			nxdl:NX_ANY
+			nxdl:NX_ARBITRARY_UNITS
 			nxdl:NX_AREA
 			nxdl:NX_CROSS_SECTION
 			nxdl:NX_CHARGE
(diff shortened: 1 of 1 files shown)


nxdlTypes.xsd
```diff
@@ -102,6 +103,34 @@
...
+				units to some engineering units.  This should be considered as a
+				possibility in the analysis process.
+
+				**NOTE**: not the same as NX_UNITLESS or NX_DIMENSIONLESS
+
+				**NOTE**: It is common (such as in :ref:`NXcanSAS`) for intensity to
+				be reported in arbitrary units.
+
+				<xs:element name="example">arbitrary</xs:element>
+				<xs:element name="example">a.u.</xs:element>
```
PeterC-DLS, line 128: Not to be confused with astronomical units(!)

phyy-nx:
Are these the same as ADU (analog digital units)? I always understood ADU as being an arbitrary unit.  If so, we could mark our detector data as in arbitrary units for systems where that's applicable.

prjemian:
ADU usually has a scaling factor to engineering units

nxdlTypes.xsd
prjemian: noted

prjemian:
@phyy-nx The example of arbitrary units is the relative or normalized scattered intensity.  Often intensity is measured in counts.  Normalizing by a constant source parameter (such as incident beam monitor) can involving mixing of units (counts divided by photocurrent if the monitor is an ioc chamber, for example) where the conversion factor is too cumbersome to represent or no use of such a factor is made in data processing.  Hence: _arbitrary units_ have become customary.

phyy-nx:
Great, thanks for the clarification

prjemian:
Adding this new unit type is seen as unnecessary.  NeXus technical committee discussed possibilities using [`NX_UNITLESS`](https://github.com/nexusformat/definitions/blob/4e48fcf6a6d8e01bf995bd0ced4b4b416ddb6e80/nxdlTypes.xsd#L78), [`NX_DIMENSIONLESS`](https://github.com/nexusformat/definitions/blob/4e48fcf6a6d8e01bf995bd0ced4b4b416ddb6e80/nxdlTypes.xsd#L57), if the units truly cancel, or omitting units entirely.

In the end, the tech committee recommended using `units="unknown"` (which is recognized by UDUNITS) in such cases, rather than creating a new type of units.  Application software will recognize this value when considering the units of a field.

PR #991, opened 2022-03-01
clarify that NeXus utilities are not directly code

benajamin:
fixes https://github.com/nexusformat/NIAC/issues/91

### DIFF
--- a/manual/source/examples/index.rst
+++ b/manual/source/examples/index.rst
@@ -46,7 +46,7 @@ These examples illustrate the use of the NAPI
 .. _Examples.readers:
 
 Code Projects that work with NeXus data files
-################################
+#############################################
 
 The number of tools that read NeXus data files,
 either for general use or to read a specific
--- a/manual/source/examples/index.rst
+++ b/manual/source/examples/index.rst
@@ -45,12 +45,13 @@ These examples illustrate the use of the NAPI
 
 .. _Examples.readers:
 
-Code that reads NeXus data files
+Code Projects that work with NeXus data files
 ################################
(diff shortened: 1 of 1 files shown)


manual/source/examples/index.rst
```diff
@@ -45,12 +45,13 @@ These examples illustrate the use of the NAPI
 
 .. _Examples.readers:
 
-Code that reads NeXus data files
+Code Projects that work with NeXus data files
 ################################
```
prjemian, line 49: needs more `#` to match length of previous line

manual/source/examples/index.rst
```diff
@@ -45,12 +45,13 @@ These examples illustrate the use of the NAPI
 
 .. _Examples.readers:
 
-Code that reads NeXus data files
+Code Projects that work with NeXus data files
```
prjemian, line 48: How about `Tools that work with NeXus data files` ?  Seems more consistent with https://github.com/nexusformat/NIAC/issues/91

prjemian (changes requested):
as noted

manual/source/examples/index.rst
benajamin: Normally, I'd agree, but I see this as a special case. My thinking is that people come to this page looking for code examples. "Tools" doesn't indicate that example code can be found, while "code projects" does imply that one can find example code with a bit of digging.

prjemian: Ok

prjemian (approved):
LGTM

PR #997, opened 2022-03-02
ellipsometry proposal

sanbrock:


sanbrock:
https://fairmat-experimental.github.io/nexus-fairmat-proposal/d122a69ce0c953805e60a662e9580ee2c4a6fae7/index.html

mkoennecke (commented):
I wonder if a NXprocess group would be better instead of the program text field. 

At entry level there is program field, at the instrument level there is software field which both seem to describe the software used to generate the file. I assume now that there is only one program used to generate the file. Without further explanation it looks as if there is a duplication which should be avoided. I suggest to describe the software used in a NXprocess group at entry level.  


light_source and other_light_source are fields: I wonder if this should not be NXsource groups. And  add the enumeration values to the types in NXsource.

I feel that focussing_probes_applied and data_correction_applied would be clearer field names. And these names would be similar to names we use in NXdetector for such things. 

The stage NXsubentry confuses me. IMHO, the data there belongs into NXsample. 

I found things like data_type and measured_data in the NXsample group. IMHO they belong into the NXdetector group. 

Wavelength, IMHO, belongs into a NXmonochromator group in the NXinstrument hierarchy. 

I found a varied_parameters field. Please read the section on how to describe scans in NeXus in the manual. IMHO, this applies for this use case. 

I am not sure that NXsubentry is the right class for optical_excitation. The field listed in there rather suggest to me that this a kind of sample environment. And NXenvironment being the more appropriate group to describe this. And a whole bunch of fields would need to be added to NXenvironment in order to fit the bill. 

There is a group derived_parameters of type NXcollection with only one field in it, depolarization. Will there be other fields in this group? Is it worth considering to pull this up one level?

tomio13:
to @mkoennecke about NXprocess group would be better instead of the program text field: This is a field only to record what software was used to generate the data. It is not a description of any data processing done.

contributed_definitions/NXellipsometry.nxdl.xml
```diff
@@ -0,0 +1,839 @@
...
+            <doc>
+                 Number of time points measured, the length of NXsample/time_points
+            </doc>
+        </symbol>
+    </symbols>
+    <doc>
+         Ellipsometry, complex systems, up to variable angle spectroscopy.
+         
+         Information on ellipsometry is provided, e.g. in:
+         - H. Fujiwara, Spectroscopic ellipsometry: principles and applications,
```
prjemian, line 46: This is a markdown-style bullet list.  Change to reST:

- add blank lines before and after the bullet list
- change `-` to `*`

contributed_definitions/NXellipsometry.nxdl.xml
```diff
@@ -0,0 +1,839 @@
...
+           John Wiley &amp; Sons, 2007.
+         - R. M. A. Azzam and N. M. Bashara, Ellipsometry and Polarized Light,
+           North-Holland Publishing Company, 1977.
+         - H. G. Tompkins and E. A. Irene, Handbook of Ellipsometry,
+           William Andrew, 2005.
+         
+         Open acces sources:
```
prjemian, line 53: `acces` -> `access`

contributed_definitions/NXellipsometry.nxdl.xml
```diff
@@ -0,0 +1,839 @@
...
+                             laser radiation, a value or a set of values may do but for other illumination
+                             types, such as pulsed lasers, or lamps, a range may describe the source better.
+                        </doc>
+                    </field>
+                    <field name="broadening" optional="true" type="NX_NUMBER" units="NX_LENGTH">
+                        <doc>
+                             Specify the FWHM of the excitation
+                        </doc>
+                    </field>
+                    <group name="duration" optional="true" type="NX_TIME">
```
prjemian, line 794: `field` not `group`

`type="NX_NUMBER" units="NX_TIME"`

prjemian (approved):
LGTM for contributed

cmmngr:
The requested changes were implemented

PR #998, opened 2022-03-02
mpes proposal

sanbrock:


sanbrock:
https://fairmat-experimental.github.io/nexus-fairmat-proposal/d122a69ce0c953805e60a662e9580ee2c4a6fae7/index.html

mkoennecke (commented):
NXdeflector.nxdl.xml

It is not needed to copy the content of NXtransformations into the definition. It is enough to state that a group of type NXtransformations is required.

But a depends_on field is missing which points to the start of the transformation chain.

Both NXcalibration and NXdistortion contain a NXcollection for describing the procedure employed. I feel that NXprocess would be more appropriate here. 

It is not clear to me if NXelectronanalyser is a base class or an application definition. With  a few more extras it would look like an application definition. 

It should be made clearer that NXmpes, NXmpes_ARPES are at the application definitions whereas the other things are suggested new base classes. 

You also suggest extensions to NXsource, mostly adding to the enums for probes and types etc. I would separate this out, that is an easy thing for the NIAC to agree too.

contributed_definitions/NXdistortion.nxdl.xml
```diff
@@ -0,0 +1,90 @@
...
+        </doc>
+    </field>
+    <field name="applied" type="NX_BOOLEAN">
+        <doc>
+             Has the distortion correction been applied?
+        </doc>
+    </field>
+    <field name="symmetry" type="NX_INT" units="NX_UNITLESS">
+        <doc>
+             For symmetry-guided distortion correction
```
prjemian, line 39: consider making this a link:

```rst
:ref:`symmetry-guided distortion correction <https://www.sciencedirect.com/science/article/abs/pii/S0304399118303474?via%3Dihub>`
```

contributed_definitions/NXspindispersion.nxdl.xml
```diff
@@ -0,0 +1,76 @@
+<?xml version="1.0" ?>
+<?xml-stylesheet type="text/xsl" href="nxdlformat.xsl"?>
+<definition category="base" extends="NXobject" name="NXspindispersion" type="group" xmlns="http://definition.nexusformat.org/nxdl/3.1" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.w3.org/2001/XMLSchema-instance">
+    <doc>
+         Subclass of NXelectronanalyser to describe the spin filters in photoemission
+         experiments.
+    </doc>
+    <field name="spin_filter_type" type="NX_CHAR">
```
prjemian, line 8: `spin_` is a common prefix for all these fields.  Consider if you might drop it (for shorter names).

prjemian: As written, it is easier to locate in a search.

contributed_definitions/NXmpes.nxdl.xml
```diff
@@ -0,0 +1,298 @@
...
+                    <group name="field_aperture" optional="true" type="NXaperture">
+                        <doc>
+                             The size and position of the field aperture inserted in the column. To add
+                             additional or other apertures use the APERTURE group of NXcollectioncolumn.
+                        </doc>
+                    </group>
+                    <group name="contrast_aperture" optional="true" type="NXaperture">
+                        <doc>
+                             The size and position of the contrast aperture inserted in the column. To add
+                             additional or other apertures use the APERTURE group of Nxcollectioncolumn.
```
prjemian, line 101: capitalize `NX` here in `NXcollectioncolumn`

prjemian (approved):
LGTM as contributed

contributed_definitions/NXmpes.nxdl.xml
domna: Typo is corrected

contributed_definitions/NXdistortion.nxdl.xml
domna: Changed it to a proper link

domna (commented):
All the requested points have been addressed in the later commits.

PR #999, opened 2022-03-02
Adding depends on

zjttoefs:
The `depends_on` way of defining the location of an instrument component works well. But in some cases the proper reference location that is placed by the `depends_on` is not well defined or obvious. 

This pull request mainly aims to rectify this.

`depends_on` has been added to all base classes that would benefit from a well defined location. (`NXnote` would not, for example.)

In cases where we felt we could define the reference location, we have done so. With drawings in most cases. 
Some base classes call for expert input what the reference location to place would be. TODO markers have been added.
If the preference would be to leave those classes untouched, that would mask an issue, but be acceptable to us.

OFF geometry shapes can be used to make the location unambiguous. For that reason they have been added. To keep the changeset small those edits could be deferred. Same for the explicit addition of NXtransformations.

In addition NXgeometry has been declared deprecated. Again, if this is not desired or does not reflect reality this does not have to happen as part of this PR.

The addition of the depends_on locations is a valuable tie breaker, though, and I would be thankful if that could be considered carefully.


zjttoefs:
According to https://www.nexusformat.org/NIAC2022_spring.html#minutes and https://github.com/nexusformat/NIAC/issues/111#issuecomment-1059295950 this is approved. 
How/when are we doing the merge?

prjemian:
Is this a duplicate of #897?

benajamin:
> Is this a duplicate of #897?

Yes, this does appear to be a duplicate, but with some cross-reference links and additional image files that #897 could profit from appropriating.

phyy-nx:
I am working on a unified PR for #897 and #999

PR #1020, opened 2022-03-07
Deprecate incr attribute of dimension type

### LINKED ISSUE #963: Deprecate incr tab in dimensionsType
Reading the description for [incr](https://manual.nexusformat.org/nxdl_desc.html#incr), it seems that it was accidentally left undeprecated in 2016 when the `ref` and `refindex` tags were deprecated

PeterC-DLS:
- Fixes #963 
- close nexusformat/NIAC#112. 

Also remove from NXfilter's coating_roughness field (which was add in 20090304 by change f86bdcb4591707db84346207dbbc49297c75c8a0)

### DIFF
--- a/base_classes/NXfilter.nxdl.xml
+++ b/base_classes/NXfilter.nxdl.xml
@@ -149,7 +149,7 @@
 	</field>
 	<field name="coating_roughness" type="NX_FLOAT" units="NX_LENGTH">
 		<doc>coating roughness (RMS) of supermirror filter</doc>
-		<dimensions rank="1"><dim index="1" incr="1" value="nsurf"/></dimensions>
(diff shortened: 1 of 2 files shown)


PeterC-DLS:
Needs reviewing @mkoennecke, @prjemian, @benajamin

benajamin (approved):
LTGM
I did a search and found that `@incr` is also used in [manual/source/examples/array-dimensions.xml.txt](https://github.com/nexusformat/definitions/blob/main/manual/source/examples/array-dimensions.xml.txt) but I don't think that file is actually referenced for inclusion in the manual. I will make a new issue about removing unreferenced files from the manual source.

prjemian (approved):
LGTM

PR #1024, opened 2022-03-08
syntax checks for pull requests

woutdenolf:
Closes #1023

PR #1016, opened 2022-03-05
Run CI only for the main branch. Run syntax check for pull requests.

woutdenolf:
Closes #1017

PR #1026, opened 2022-03-08
execute CI only for the main branch

### LINKED ISSUE #1025: Execute CI on main branch only

woutdenolf:
- Closes #1025

### DIFF
--- a/.github/workflows/publish-sphinx.yml
+++ b/.github/workflows/publish-sphinx.yml
@@ -1,5 +1,10 @@
 name: Publish Sphinx Docs to GitHub Pages
-on: [push]
+
+on:
+  # Triggers the workflow on push events but only for the main branch
+  push:
+    branches:
+      - main
 
 # see: https://sphinx-notes.github.io/pages/
 # see: https://github.com/marketplace/actions/sphinx-to-github-pages
--- a/.github/workflows/syntax-checks.yml
+++ b/.github/workflows/syntax-checks.yml
@@ -1,6 +1,13 @@
 name: Syntax Checks
 
-on: [push, pull_request]
+on:
+  # Triggers the workflow on push or pull request events but only for the main branch
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
 
 jobs:
   build-linux:


woutdenolf:
This is part 2 of the original pull request https://github.com/nexusformat/definitions/pull/1016: run CI only for the main branch

.github/workflows/publish-sphinx.yml
```diff
@@ -1,5 +1,10 @@
 name: Publish Sphinx Docs to GitHub Pages
-on: [push]
+
+on:
```
prjemian, line 3: This change is not necessary since documentation will not be published unless new tag is pushed to the repository: https://github.com/nexusformat/definitions/blob/af7ea8ba357cc0651eb435bf2a6236da84b2206e/.github/workflows/publish-sphinx.yml#L109

.github/workflows/syntax-checks.yml
```diff
@@ -1,6 +1,13 @@
 name: Syntax Checks
 
-on: [push]
+on:
```
prjemian, line 3: This is in conflict with https://github.com/nexusformat/definitions/pull/1024/files#diff-d679df740cd55f9498164266c52f7939b5b0fd15c83cece4a835a5473aee7640R3

In #1024, syntax checks will be run on branches that have a pull request to the repository.  This PR implements a change so that syntax checks are run only on the main branch.

.github/workflows/publish-sphinx.yml
woutdenolf: What if someone pushes a tag to another branch?

prjemian: Great question.  Looks possible right now.  Tags are not branch based, so the tag is on a commit hash in the repository.  Don't think we want any tags on branch commits.  What happens to the tag if the branch is closed without a merge?

.github/workflows/syntax-checks.yml
prjemian: I want this immediate feedback on any push.

.github/workflows/publish-sphinx.yml
woutdenolf: Yes I agree one should never push a tag to another branch than the default branch. But who knows, at some point someone might make a mistake.

.github/workflows/syntax-checks.yml
prjemian: For me, that is long-desired effect of a _continuous_ integration process.  The PR proposes an _occasional_ integration process.

.github/workflows/publish-sphinx.yml
woutdenolf: We actually never do it this way. One could always make mistakes when pushing tags.

We normally require someone to run the CI on the master manually with a flag that allows deployment explicitly.

https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow

prjemian: Probably needs its own issue.  When I get time, I want to read this page: https://git-scm.com/book/en/v2/Git-Basics-Tagging

.github/workflows/syntax-checks.yml
woutdenolf: I think it is good practice to always have a pull request.

woutdenolf:
@prjemian We can discuss this at some NIAC meeting. It's not super urgent, it's just that imo we may be wasting github action resources for no reason, that's all.

.github/workflows/publish-sphinx.yml
woutdenolf: Agreed, it's a different issue.

woutdenolf: https://github.com/nexusformat/definitions/issues/1029

.github/workflows/syntax-checks.yml
woutdenolf: If your branch has a PR you have as many PR events as push events.

.github/workflows/publish-sphinx.yml
PeterC-DLS: In other repos, I've used release branches (which, I guess should be protected too) and tag releases on them.

.github/workflows/syntax-checks.yml
```diff
@@ -1,6 +1,13 @@
...
 
-on: [push, pull_request]
+on:
+  # Triggers the workflow on push or pull request events but only for the main branch
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
```
prjemian, line 11: additionally, only for pushes to main branch and for PRs that will merge into main branch, correct?

prjemian:
The `.rst` file changes here are unrelated.

.github/workflows/syntax-checks.yml
woutdenolf: Indeed, it's just to avoid unnecessary CI usage. Basically you only want CI when you either push to the `main` branch or create/modify a pull request to the `main` branch.

prjemian (approved):
LGTM, thanks for explaining.

PR #1027, opened 2022-03-08
DOC doc string all on one line

prjemian:
For NXDL class summaries, only first line of doc block will be used.  Revise obvious content to fit.


prjemian:
Other contributed definitions also need revision:

- NXquadric

PeterC-DLS (approved):

prjemian:
No changes for NXquadric after all.

PR #1033, opened 2022-03-13
add a jupyter notebook on the topic of reference frames

woutdenolf:


.gitignore
```diff
@@ -10,6 +11,11 @@ manual/source/*.table
 manual/source/classes/*/NX*.rst
 manual/source/classes/*/*.nxdl.xml
 manual/source/classes/contributed_definitions/canSAS
+manual/source/classes/applications/canSAS
```
prjemian, line 14: These items to be ignored in `manual/source/classes` can appear in local directories when the documentation is built in the source directory and not in the `build` directory.  We could use improvements in the `/Makefile` sections to remove these files before a `git add`, improvements that might help to avoid expanding this growing set of `.gitignore` items.

.github/env-workflow.yml
```diff
@@ -9,4 +9,6 @@ dependencies:
   - lxml
   - pyRestTable
   - Sphinx
+  - nbsphinx
```
prjemian, line 12: Will there be a step added to the workflow (and also to the `/Makefile`) that uses `nbsphinx` to build documentation pages from the notebooks?  There is an [example `Makefile`](https://github.com/bluesky/hklpy/blob/main/examples/Makefile) in the *hklpy* project.

prjemian: That step would happen *before* sphinx is called to `make html`, and then the PDF version of the documentation.

manual/source/conf.py
```diff
@@ -46,6 +46,7 @@
     'sphinx.ext.ifconfig',
     'sphinx.ext.viewcode',
     'sphinx.ext.githubpages',
+    'nbsphinx'
```
prjemian, line 49: Is this the [*heavy lifter*](https://nbsphinx.readthedocs.io/en/0.8.8/) that brings notebooks into the documentation automagically (without need for a `Makefile`)?

.github/env-workflow.yml
woutdenolf: There is no need for adding another steps or modifying the makefile. The sphinxs extension `nbsphinx` takes care of everything (basically the .ipynb become equivalent to .rst files).

manual/source/conf.py
woutdenolf: Indeed, the sphinx extension `nbsphinx` basically turns all referenced jupyter notebooks in source files for documentation.

.gitignore
woutdenolf: Or we should not allow a build in the source directory.

prjemian: How would that be possible?

woutdenolf: By not adding that possibility to the Makefile? So `make local` becomes simply `make`.

woutdenolf: I'm actually surprised files are added to the source directory. That's not normal `sphinx` behaviour. I guess this is done by some custom code somewhere?

prjemian:
Right. Before sphinx is run, the nxdl.xml files are parsed, creating .rst
files for the documentation.

.gitignore
prjemian: `.rst` files are created (from documentation in NXDL files and the Schema) before sphinx is called.  Usually, the documentation is built out-of-source (outside the source tree).  An in-source build was always possible and that must be how these files were created locally.

prjemian:
I like the way this looks.  Makes it so easy to add documentation which becomes part of the manual_ from a Jupyter notebook.

Noticing that a single notebook creates 4 subsections.  Was expecting a single subsection per notebook, with sub-subsections.  Was that intentional?  Is it designed (by the notebook author)?  Looks like it.  Each of the sections there use `#`.  Thinking that a notebook should only have a single level-1 heading, then level-2 (`##`) or more below that heading.  The idea is that all this content belongs to this one document.

prjemian:
@woutdenolf Should we consider this PR at the [2022-06 Code Camp](https://github.com/orgs/nexusformat/projects/1)?

woutdenolf:
Not part of the code camp. I made this more to discuss https://github.com/nexusformat/NIAC/issues/102

Maybe after some cleanup this can be actually part of the documentation.

Not for [2022-06 Code Camp](https://github.com/orgs/nexusformat/projects/1) I would say.

vincefn:
Looks good to me. This will also be useful for 3D reciprocal space plots.
I would suggest to add a link in the doc to https://en.wikipedia.org/wiki/Active_and_passive_transformation to explain the notion of _passive/active_ transformations, which can be a bit obscure otherwise.

woutdenolf:
NXdata has a link to 

https://hdf5.gitlab-pages.esrf.fr/nexus/nxdata_transformation/notebooks/reference_frames.html#NXdata-reference-frame

The passive/active transformation is defined there. Perhaps not clear enough? I'd like to avoid links to wikipedia.

PR #1039, opened 2022-03-24
contribution for stress-strain analysis

woutdenolf:
Application definition for stress-strain analysis as defined by the member of https://easi-stress.eu/


phyy-nx:
From telco:

- Consider NXreflections instead of NXdata for `peaks` (might not be relevant)
- `center` and `center_errors` probably needs to be units `NX_ANY`, due to not having a mechanism for dereferences into an enum of possible units
- @hgoerzig to share an alternative for `calibration_data_file`
- Consider extending NXbeam to handle the beam profile more explicitly, like shown here, but this isn't a requirement for putting in this contributed definition.  Note, NXmx also has a profile field, but it's not as expansive, and considers different aspects of the profile.  Also look at NXcoordinate system contributed definition

Need one review from the committee and then this can go in.  Thanks!

woutdenolf:
Note to self: an example of using [NXcoordinate_system_set](https://fairmat-nfdi.github.io/nexus_definitions/classes/contributed_definitions/NXcoordinate_system_set.html#nxcoordinate-system-set): 

https://fairmat-nfdi.github.io/nexus_definitions/classes/contributed_definitions/NXapm_paraprobe_results_nanochem.html#nxapm-paraprobe-results-nanochem-entry-coordinate-system-set-group

woutdenolf:
* fixed NXsample transformations
* `qmagnitude` is redundant so remove it (if you can provide it you might as well provide center with type momentum transfer)

woutdenolf:
- rename diffraction_type to processing_type and add formulae
- use enumerations when appropriate
- rebase

contributed_definitions/NXstress.nxdl.xml
```diff
@@ -0,0 +1,1049 @@
...
+    However, when possible and technically feasible, the instrument using the NXstress application definition is expected to provide the type of information outlined below. 
+
+    Sample and detector positions can be defined with :ref:`NXtransformations`. If you don't specify the direction of gravity
+    and the direction of the beam then the standard NeXus Coordinate System is used.
+
+    It is highly recommended that in case certain parameters or values are the same for all the measurements (acquistions) in the same
+    file, that they are stored only in one location and then linked in the other instances. For example, if during an acquisition all
+    instrumental parameters but one stay the same and only the sample table moves in one direction (e.g. Xtranslation), then all the
+    static instrumental parameters should be saved just once (e.g. in just one NXentry or in a *Shared_Information group*) and their
+    vales linked to every *instrument group* under all the other acquistions. The value for the variable that changes, Xtranslation
```
phyy-nx, line 76:
```suggestion
    values linked to every *instrument group* under all the other acquisitions. The value for the variable that changes, Xtranslation
```

contributed_definitions/NXstress.nxdl.xml
```diff
@@ -0,0 +1,1049 @@
...
+                <field name="depends_on" type="NX_CHAR" optional="true">
+                    <doc>
+                        The axis on which the detector position depends may be stored
+                        anywhere, but is normally stored in the *NXtransformations
+                        group* within the *NXdetector group*.
+                    </doc>
+                </field>
+                <group type="NXtransformations" minOccurs="0">
+                    <doc>
+                        This is the recommended location for detector goniometer
```
phyy-nx, line 308: Do these experiments include goniometers?

phyy-nx (commented):
Just a quick scan before the Telco

prjemian:
[NXparameters](https://manual.nexusformat.org/classes/base_classes/NXparameters.html) should be used in many places where [NXdata](https://manual.nexusformat.org/classes/base_classes/NXdata.html) is used.  NXdata is intended to describe how to plot data.

contributed_definitions/NXstress.nxdl.xml
woutdenolf: They might. @aapaecklar can provide details.

woutdenolf:
Changes applied as response to Telco discussion:

* use NXparameters instead of NXdata and NXcollection, except for the DIFFRACTOGRAMS groups which plots the powder diffractogram fit(s) and therefore NXdata is appropriate
* added missing peak_parameter groups (center and errors)

aapaecklar:
Do we have any updates here?

contributed_definitions/NXstress.nxdl.xml
s-cabeza: Dear all, for the moment we only consider cradles for the eventual sin2psi method (1 ax rotation only)

mkuehbach:
n_Temp and n_sField symbols not defined in the symbol table

woutdenolf:
The n_Temp and n_sField symbols are defined in the symbol table of `NXsample`. How do we deal with that?

woutdenolf:
- We have `optional="false"` or `minOccurs="1"` in different cases where it can occur 0 or 1 time. What is the preferred way to denote this?
- We have several `optional="true"`. Should we drop those since it is the default?

woutdenolf:
Repeat the symbols imported from included classes.

aapaecklar:
Hello, are there any news on this topic?

woutdenolf:
> Hello, are there any news on this topic?

This one has been accepted by the NIAC. I just need these technicalities sorted out: https://github.com/nexusformat/definitions/pull/1039#issuecomment-2382590370

woutdenolf:
Moved NXstress to applications

To be done: 

- [ ] Repeat the symbols imported from included classes.
- [ ] Drop optional="true" since it is the default
- [ ] Consistent usage of optional="false" vs. minOccurs="1"

applications/NXstress.nxdl.xml
```diff
@@ -0,0 +1,1115 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?xml-stylesheet type="text/xsl" href="nxdlformat.xsl"?>
+<!--
+# NeXus - Neutron and X-ray Common Data Format
+# 
+# Copyright (C) 2008-2022 NeXus International Advisory Committee (NIAC)
```
PeterC-DLS, line 6: Update end year

applications/NXstress.nxdl.xml
```diff
@@ -0,0 +1,1115 @@
...
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+#
+# For further information, see http://www.nexusformat.org
+-->
+<definition name="NXstress" extends="NXobject" type="group" category="application"
+    xmlns="http://definition.nexusformat.org/nxdl/3.1"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://definition.nexusformat.org/nxdl/3.1 ../nxdl.xsd">
+    <symbols>
+        <symbol name="nX">
```
PeterC-DLS, line 28: Stylistically, maybe snake case all around: `n_X`, etc

applications/NXstress.nxdl.xml
```diff
@@ -0,0 +1,1115 @@
...
+    The analysis is typically some form of diffraction peak indexing and fitting.
+    The experiments are for example
+
+    - energy-dispersive X-ray powder diffraction
+    - angular-dispersive X-ray powder diffraction
+    - angular-dispersive neutron powder diffraction
+    - time-of-flight (TOF) neutron powder diffraction.
+
+    In addition, the application definition guarantees that the information about instrumental setups, measurement conditions, and data analysis workflows are described.
+    This ensures not only the reproducability and tracability of the measured data, but also the meta-data. Since not all participating beamlines or instruments can provide an input to all the NXfields listed in this application definition, not all of them are "required".
```
PeterC-DLS, line 72:
```suggestion
    This ensures not only the reproducability and tracability of the measured data, but also the metadata. Since not all participating beamlines or instruments can provide an input to all the NXfields listed in this application definition, not all of them are "required".

```

applications/NXstress.nxdl.xml
```diff
@@ -0,0 +1,1115 @@
...
+    However, when possible and technically feasible, the instrument using the NXstress application definition is expected to provide the type of information outlined below. 
+
+    Sample and detector positions can be defined with :ref:`NXtransformations`. If you don't specify the direction of gravity
+    and the direction of the beam then the standard NeXus Coordinate System is used.
+
+    It is highly recommended that in case certain parameters or values are the same for all the measurements (acquistions) in the same
```
PeterC-DLS, line 78:
```suggestion
    It is highly recommended when certain parameters or values are the same for all the measurements (acquistions) in the same

```

applications/NXstress.nxdl.xml
```diff
@@ -0,0 +1,1115 @@
...
+    file, that they are stored only in one location and then linked in the other instances. For example, if during an acquisition all
```
PeterC-DLS, line 79:
```suggestion
    file, they are stored only in one location and then linked in the other instances. For example, if during an acquisition all

```

applications/NXstress.nxdl.xml
```diff
@@ -0,0 +1,1115 @@
...
+    instrumental parameters but one stay the same and only the sample table moves in one direction (e.g. Xtranslation), then all the
+    static instrumental parameters should be saved just once (e.g. in just one NXentry or in a *Shared_Information group*) and their
+    values linked to every *instrument group* under all the other acquisitions. The value for the variable that changes, Xtranslation
+    in this example, is suggested to only be saved at every instrument group under each acquistion but not in the *Shared_Information group*.
```
PeterC-DLS, line 83:
```suggestion
    in this example, is suggested to be saved only at every instrument group under each acquistion but not in the *Shared_Information group*.

```

applications/NXstress.nxdl.xml
```diff
@@ -0,0 +1,1115 @@
...
+                </item>
+                <item value="time-of-flight">
+                    <doc>
+                        :math:`\varepsilon = \large \frac{\mathrm{TOF}}{\mathrm{TOF}_{0}}-1`
+                    </doc>
+                </item>
+                <item value="sin2psi">
+                    <doc>
+                        A description of the :math:`\mathrm{\sin}^{2}\psi` method can be found in the literature. Two examples are:
+                        `Fitzpatrick, M. E. et. al.; Determination of residual stresses by X-ray diffraction; National Physical Laboratory, Teddington, 2005`
```
PeterC-DLS, line 166: Why are these phrase references?

applications/NXstress.nxdl.xml
```diff
@@ -0,0 +1,1115 @@
...
+                    <doc>Short name for the instrument, perhaps the acronym, which would be for the the example above ``SALSA``.</doc>
+                </attribute>
+            </field>
+            <group name="CALIBRATION" nameType="any" type="NXnote" minOccurs="0" maxOccurs="1">
+                <doc>This group contains information about the geometry and/or efficiency measurement(s).</doc>
+                <field name="calibration_type" type="NX_CHAR" minOccurs="0" maxOccurs="1">
+                    <doc>Describe the type of calibration.</doc>
+                </field>
+                <field name="file_name" type="NX_CHAR" minOccurs="0" maxOccurs="1">
+                    <doc>File name(s) and/or path(s) (withing file(s)) containing data from the last calibration(s). This can be a list.</doc>
```
PeterC-DLS, line 206:
```suggestion
                    <doc>File name(s) and/or path(s) (within file(s)) containing data from the last calibration(s). This can be a list.</doc>

```

applications/NXstress.nxdl.xml
```diff
@@ -0,0 +1,1115 @@
...
+                <item value="tangential"/>
+                <item value="multiple"/>
+            </enumeration>
+        </field>
+        <group name="experiment_responsible" type="NXuser" minOccurs="0" maxOccurs="1">
+            <doc>Information about the person who performed the experiment.</doc>
+            <field name="name" type="NX_CHAR" minOccurs="0" maxOccurs="1">
+            </field>
+            <field name="role" type="NX_CHAR" minOccurs="0" maxOccurs="1">
+                <doc>Role of user responsible for this entry. Suggested roes are, for example, ``local contact``, ``beamline_scientist``, ``post_doc``,…</doc>
```
PeterC-DLS, line 190: Is the repetition of docs here and below necessary any more? They become a maintenance burden

applications/NXstress.nxdl.xml
```diff
@@ -0,0 +1,1115 @@
...
+                    </doc>
+                    <enumeration>
+                        <item value="neutron"/>
+                        <item value="X-ray"/>
+                    </enumeration>
+                </field>
+            </group>
+            <group name="DETECTOR" nameType="any" type="NXdetector" minOccurs="1" maxOccurs="unbounded">
+                <doc>Zero or more of these groups describe the detectors used in the experiment.</doc>
+                <field name="description" type="NX_CHAR" minOccurs="0" maxOccurs="1">
```
PeterC-DLS, line 248: Use `NXfabrication` instead?

applications/NXstress.nxdl.xml
```diff
@@ -0,0 +1,1115 @@
...
+                    <doc>
+                        This is the recommended location for detector goniometer
+                        and other related axes.
+                    </doc>
+                </group>
+            </group>
+            <group name="beam_intensity_profile" type="NXbeam" minOccurs="1" maxOccurs="1">
+                <doc>
+                 Defines the dimensions of the beam profile used for probing the sample which corresponds to or can be used to determine the instrumental gauge volume.  
+                 A description of the subsequent fields can be found in the folowing figure. The term "primary" in the subsequent fields refers to the beam path between the sample and the source. The term "secondary" refers to the beam path between the sample and the detector(s).
```
PeterC-DLS, line 322: This changes NXbeam to sample two locations instead of

applications/NXstress.nxdl.xml
```diff
@@ -0,0 +1,1115 @@
...
+                <doc>Zero or more of these groups describe the detectors used in the experiment.</doc>
+                <field name="description" type="NX_CHAR" minOccurs="0" maxOccurs="1">
+                    <doc>name/manufacturer/model/etc. information</doc>
+                </field>
+                <field name="type" type="NX_CHAR">
+		            <doc>
+			                Description of type such as \ :sup:`3`\ He gas cylinder, \ :sup:`3`\ He PSD, scintillator, fission chamber, proportion counter, ion chamber, CCD, pixel, image plate, CMOS, …
+		            </doc>
+	            </field>
+                <field name="distance" type="NX_NUMBER" units="NX_LENGTH" minOccurs="0" maxOccurs="1">
```
PeterC-DLS, line 256: This changes the meaning of the field from the base class

applications/NXstress.nxdl.xml
```diff
@@ -0,0 +1,1115 @@
...
+                    <doc>
+                        Length of the second diagonal normal to :ref:`x &lt;/NXstress/ENTRY/sample_description/gauge_volume/x-field&gt;`.
+                    </doc>
+                </field>
+                <field name="z" type="NX_FLOAT" units="NX_LENGTH" minOccurs="0" maxOccurs="1">
+                    <doc>
+                        Height of the gauge volume.
+                    </doc>
+                </field>            
+                <field name="depends_on" type="NX_CHAR" minOccurs="0" maxOccurs="1">
```
PeterC-DLS, line 488: `NXmx` has both `depends_on` and `NXtransformations` in the `NXsample` group level

applications/NXstress.nxdl.xml
```diff
@@ -0,0 +1,1115 @@
...
+                <doc> The gauge volume can be described with the following parameters:
+                 .. figure:: stress/GV2.jpg
+                  :width: 70%
+                  :alt: Dimensions of the gauge volume.
+
+                  Parameters describing the gauge volume.
+                </doc>
+                <field name="x" type="NX_FLOAT" units="NX_LENGTH" minOccurs="0" maxOccurs="1">
+                    <doc>
+                        Length of the first diagonal.
```
PeterC-DLS, line 475: It isn't clear how these diagonals relate to the local coordinate frame

applications/NXstress.nxdl.xml
```diff
@@ -0,0 +1,1115 @@
...
+            </group>
+            <group type="NXtransformations" minOccurs="0" maxOccurs="1">
+                <doc>
+                This is the recommended location for sample goniometer
+                and other related axes.
+                </doc>
+            </group>
+        </group>
+
+        <group name="FIT" nameType="any" type="NXprocess" minOccurs="1" maxOccurs="unbounded">
```
PeterC-DLS, line 504: The new MPES application definition has `NXfit`, etc. Perhaps this could be migrated to use those new base classes.

applications/NXstress.nxdl.xml
```diff
@@ -0,0 +1,1115 @@
...
+                    </attribute>
+                </field>
+            </group>
+        </group>
+
+        <group name="NOTES" nameType="any" type="NXnote" minOccurs="0" maxOccurs="1">
+            <doc>
+                User description of the data acquisitions.
+                A description of data analysis goes in the
+                *fig_log group*.
```
PeterC-DLS, line 933: Where is this?

applications/NXstress.nxdl.xml
```diff
@@ -0,0 +1,1115 @@
...
+                    <dim index="1" value="n_Peaks" />
+                </dimensions>
+            </field>
+            <field name="lattice" type="NX_CHAR" minOccurs="0" maxOccurs="1">
+                <doc>Crystal lattice systems (*cubic*, *hexagonal*, ...)</doc>
+                <dimensions rank="1">
+                    <dim index="1" value="n_Peaks" />
+                </dimensions>
+            </field>
+            <field name="space_group" type="NX_CHAR" minOccurs="0" maxOccurs="1">
```
PeterC-DLS, line 966: Similar fields in `NXsample_component`

applications/NXstress.nxdl.xml
woutdenolf: Replaced with RST links. I'm assuming that is what you meant.

applications/NXstress.nxdl.xml
woutdenolf: NXstress has indeed lots of repetition from base classes. The consortium wanted to highlight the fields they deemed most relevant for this application definition. If we can find a way of "highlighting fields/groups to be shown in the application definition" without copying them, that would be great!

applications/NXstress.nxdl.xml
woutdenolf: Sorry, can you clarify? You mean this NXbeam should be under NXsample?

applications/NXstress.nxdl.xml
woutdenolf: The `dimensions` was missing indeed. I added them + the symbols.

applications/NXstress.nxdl.xml
woutdenolf: Indeed.

applications/NXstress.nxdl.xml
woutdenolf: Indeed. I'm guessing the volume was important, not really the orientation. Nevertheless it might be best to be precise. I'll make a drawing with the XYZ coordinate system in the bottom-left corner. It is inconvenient that the diagonals and the height is also called x, y and z ...

applications/NXstress.nxdl.xml
woutdenolf: This must have been its original name. It is now /NXstress/ENTRY/FIT/description group.

woutdenolf: I see that /NXstress/ENTRY/FIT/description has this in the docstring "The name will be numbered to allow for ordering of steps."

Note sure what this means. There is only one description.

applications/NXstress.nxdl.xml
woutdenolf: For reference:

https://github.com/nexusformat/definitions/pull/1424/files#diff-0bd69f0a4c1bf5a303bd1c85a4fbf265bc64416513d6cec9af6212934b1274d8

https://github.com/nexusformat/definitions/pull/1424/files#diff-3baad8533e4f0fa71dcd336841c1ed902c77d00e3f63cdba7b8a83298d59ba6b

That would require work and time and I don't have any.

applications/NXstress.nxdl.xml
woutdenolf: Previous reviews resulted in the choice of NXreflections are class for the group. Fields like `space_group` can be found in NXcrytal as well but this does not help us here.

applications/NXstress.nxdl.xml
woutdenolf: I think you mean that primary and secondary should be two NXbeam instances?

applications/NXstress.nxdl.xml
woutdenolf: Also it seems like NXfit decided to have one group per peak. That is very inconvenient. Here we decided to have an NXparameters group fit a 1D field per fit parameters.

applications/NXstress.nxdl.xml
woutdenolf: Ah, they forgot nameType="any"

applications/NXstress.nxdl.xml
woutdenolf: I renamed the fields x, y and z to a, b and c. The gauge volume now has a local XYZ coordinate system with a depends_on. The base class `NXparameters` does not have it but I can add it in an application definition no?

applications/NXstress.nxdl.xml
PeterC-DLS: I think you could just omit the doc - this makes the node optional/requirement/recommend but shown in the application definition. There may be a change by FAIRmat to bubble up the base class definitions' docs in the rendered html/pdf...

applications/NXstress.nxdl.xml
PeterC-DLS: Yep, that's the standard practice

applications/NXstress.nxdl.xml
PeterC-DLS: Maybe it will be good to work with the FAIRmat people (@lukaspie) to see if you can make common base classes for fitting - they have multiple per-peak groups whereas you have the multiple peaks per group as you are using the same function (peak/profile type) for each peak.

applications/NXstress.nxdl.xml
PeterC-DLS: Too late to change...

applications/NXstress.nxdl.xml
PeterC-DLS: Too late to change...

phyy-nx:
From Telco: no more big restructuring.  Time to go ahead and get this in :)

PR #1055, opened 2022-06-13
590 nxdetector dimensions

### LINKED ISSUE #590: NXdetector seem to require 4 dimensional data

benajamin:
fixes #590 by expanding doc strings to explain that the displayed dataset dimensions are only illustrative. The symbols used have also been made more consistent.

### DIFF
--- a/base_classes/NXdetector.nxdl.xml
+++ b/base_classes/NXdetector.nxdl.xml
@@ -31,13 +31,15 @@
             >
 
   <symbols>
-    <doc>These symbols will be used below to illustrate the coordination of the shapes of datasets and the 
-	    preferred ordering of the dimensions. Each of these are optional (so the rank of the datasets 
-	    will vary according to the situation) and the general ordering principle is slowest to fastest.</doc>
-    <symbol name="np"><doc>number of scan points (only present in scanning measurements)</doc></symbol>
+    <doc>These symbols will be used below to illustrate the coordination of the rank and sizes of datasets and the 
+	  preferred ordering of the dimensions. Each of these are optional (so the rank of the datasets 
+	  will vary according to the situation) and the general ordering principle is slowest to fastest.
+	  The type of each dimension should follow the order of scan points, detector pixels,
+	  then time-of-flight (i.e. spectroscopy, spectrometry).</doc>
+	  
+    <symbol name="nP"><doc>number of scan points (only present in scanning measurements)</doc></symbol>
     <symbol name="i"><doc>number of detector pixels in the first (slowest) direction</doc></symbol>
     <symbol name="j"><doc>number of detector pixels in the second (faster) direction</doc></symbol>
-    <symbol name="k"><doc>number of detector pixels in the third (if necessary, fastest) direction</doc></symbol>
     <symbol name="tof"><doc>number of bins in the time-of-flight histogram</doc></symbol>
   </symbols>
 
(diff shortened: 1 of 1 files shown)


PeterC-DLS (approved):
LGTM

benajamin:
@RussBerg do you see any problems with the use of symbols here?

base_classes/NXdetector.nxdl.xml
```diff
@@ -31,11 +31,17 @@
...
+      preferred ordering of the dimensions. Each of these are optional (so the rank of the datasets 
+      will vary according to the situation) and the general ordering principle is slowest to fastest.
+      The type of each dimension should follow the order of scan points, detector pixels,
+      then time-of-flight (i.e. spectroscopy, spectrometry).
+    </doc>
+    
+    <symbol name="nP"><doc>number of scan points (only present in scanning measurements)</doc></symbol>
     <symbol name="i"><doc>number of detector pixels in the first (slowest) direction</doc></symbol>
     <symbol name="j"><doc>number of detector pixels in the second (faster) direction</doc></symbol>
-    <symbol name="k"><doc>number of detector pixels in the third (if necessary, fastest) direction</doc></symbol>
```
phyy-nx, line 38: I like the new documentation, but I'm worried about dropping `k` here, since it is [used](https://manual.nexusformat.org/classes/applications/NXmx.html#nxmx-entry-instrument-detector-data-field) in NxMX.    Is the idea behind dropping `k` that it's not used below in the NXdetector base class?  Perhaps in addition to dropping it, a note could be added about how some detectors have a 3rd dimension `k` of pixels that often corresponds to module number?

benajamin: I dropped `k` because it was not used anywhere. else below. The extra text in the doc string points out that the symbols and dimensions ordering, sizes, etc. are simple illustrative, rather than restrictive. I agree that a detector can produce data with more than 2 dimensions, but I don't feel it is so important to say so explicitly in this piece of reference documentation. If we keep `k`, then we should use it, but using it consistently is going to add lots of "noise" so I don't see it being worthwhile.

phyy-nx: 👍

yayahjb:
I would prefer to leave mention of k in.  It is hard to break people to
break free of thinking
of detectors as just 0, 1 and 2 dimensional.  Once they make it to three
they then
seem willing to accept more.

benajamin:
The new text specifically mentions higher dimensional detectors in the doc string of the data array:
> The next fastest axes are typically the readout of the detector. A point detector will not add any dimensions
      (as it is just a single value per scan point) to the data array, a strip detector will add one dimension, an 
      imaging detector will add two dimensions (e.g. X, Y axes) and detectors outputting higher dimensional data 
      will add the corresponding number of dimensions. Note that the detector dimensions don't necessarily have to
      be written in order of the actual readout speeds - the slowest to fastest rule principle is only a guide.

I really dislike having a symbol that isn't used (@RussBerg will consider it a bug). I also think that adding a an extra dimension of size `k` to everything looks cluttered and want to resist doing so. I will try to think up some further compromises.

benajamin:
I have further added the following text to the doc string of the symbol table:
> Note that the output of a detector is not limited to single values (0D), lists (1D) and images (2), but three or higher dimensional arrays can be produced by a detector at each trigger.

I fear that if we beat people about the head over higher dimensional detectors any further then they will be concussed beyond the ability to read!

phyy-nx:
I can accept this compromise.  @yayahjb we have `k` in the NXmx spec anyway.

yayahjb:
Sounds reasonable to me

PR #1071, opened 2022-06-13
Use a make variable to define common python for #1072

PeterC-DLS:


woutdenolf (approved):
If `python3` also works for virtual environment (pyenv, conda, ...) then I'm ok with it. t's ok for `pyenv` and `conda`.

prjemian (approved):
Great!

PR #1079, opened 2022-06-14
use unbuffered stdout for python in make files

woutdenolf:
Closes #1078 

It seems that doing `python ... > file.ext` in make files could result in `file.ext` to be incomplete.

The solution seems to be unbuffered stdout: `python -u ... > file.ext`


prjemian (approved):
Did not know about this option.  Nice!

prjemian:
@woutdenolf Merge when you are ready.

PeterC-DLS (commented):
LGTM

PR #1082, opened 2022-06-14
fix flux units and improve docs

woutdenolf:
@phyy-nx This is a PR to your branch, not the main branch. It implements the fixes I proposed.

image


phyy-nx (approved):
Excellent.  Thank you.

PR #1090, opened 2022-06-14
added more specific language for the definition fields required value

### LINKED ISSUE #1046: Is it a rule that an appdef/entry/definition must be the name of the given appdef?

RussBerg:
fixes #1046, 
added more specific language for what the **definition** field in the NXentry group must be along with an example.


prjemian (approved):
LGTM

PR #1091, opened 2022-06-15
Add depends_on and NXtransformations to more base classes

### LINKED ISSUE #681: Explicitly add depends_on and shape fields to all component type classes?

### LINKED ISSUE #1050: NXgeometry in NXaperture
I understand NXgeometry is depreceated to describe the geometry of an experiment, and NXtransformations is supposed to be used. However, this does not carry through the whole documentation, e.g. in NXaperture an NXgeometry group is suggested to describe the geometry of the aperture. In this place, I find NXgeometry much more appropriate than NXtransformations. Was this thought through when NXgeometry fell out of favour, or is this rather an unintended consequence?

phyy-nx:
This is a combination of #897 and #999.  I took language from both and spliced them together into this PR.

- closes #681 
- closes #897
- closes #999
- closes #1050

Quoting @zjttoefs from #999:
> The `depends_on` way of defining the location of an instrument component works well. But in some cases the proper reference location that is placed by the `depends_on` is not well defined or obvious.
> 
> This pull request mainly aims to rectify this.
> 
> `depends_on` has been added to all base classes that would benefit from a well defined location. (`NXnote` would not, for example.)
> 
> In cases where we felt we could define the reference location, we have done so. With drawings in most cases. Some base classes call for expert input what the reference location to place would be. TODO markers have been added. If the preference would be to leave those classes untouched, that would mask an issue, but be acceptable to us.
> 
> OFF geometry shapes can be used to make the location unambiguous. For that reason they have been added. To keep the changeset small those edits could be deferred. Same for the explicit addition of NXtransformations.
> 
> In addition NXgeometry has been declared deprecated. Again, if this is not desired or does not reflect reality this does not have to happen as part of this PR.
> 
> The addition of the depends_on locations is a valuable tie breaker, though, and I would be thankful if that could be considered carefully.

This touches a lot of files so I think it needs a careful review :)

Doesn't need a NIAC vote though.  From @zjttoefs:
> According to https://www.nexusformat.org/NIAC2022_spring.html#minutes and [nexusformat/NIAC#109 (comment)](https://github.com/nexusformat/NIAC/issues/109#issuecomment-1059283549) this is approved.


phyy-nx:
TODO:

These files without NXoff_geometry have references to NXoff_geometry: NXdisk_chopper.nxdl.xml, NXinsertion_device.nxdl.xml, NXmonochromator.nxdl.xml, NXsource.nxdl.xml, and NXvelocity_selector.nxdl.xml.

yayahjb (approved):
Nice job.  A few todos to do, but not a reason to hold this up. -- Herbert

woutdenolf:
I believe this also closes #1050?

PR #1095, opened 2022-06-15
adjust string for examples so they appear on new lines

### LINKED ISSUE #979: BUG: display of example units is confusing
When additional information is specified with a NeXus unit category, such as `NX_DIMENSIONLESS`: https://github.com/nexusformat/definitions/blob/5649a991793ff3f8a0555a0d3a0df5f7b2c7fc01/nxdlTypes.xsd#L146-L155

the examples appear as appended to the last text with a comma prefix added.

Clipboard03

Too confusing when additional text is added, such as the proposed `NX_ARBITRARY_UNITS` (#921).

RussBerg:
- fixes #979 

a single newline character was ignored during conversion to html, new line with a tab worked and rendered it looks ok

### DIFF
--- a/utils/units2rst.py
+++ b/utils/units2rst.py
@@ -58,7 +58,7 @@ def worker(nodeMatchString, section = 'units'):
                     examples.append("``"+example.text+"``")
             a = words.text
             if len(examples) > 0:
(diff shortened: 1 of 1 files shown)


prjemian (approved):
LGTM, thanks!

PR #1097, opened 2022-06-16
Fix Makefiles to clean up properly

PeterC-DLS:
Propagate cleaning to all sub-directories, remove generated index.rst, remove copied definitions sub-directories


prjemian (approved):
Just one question about generalizing, otherwise all good.

woutdenolf (approved):
LGTM

I checked with

```
rm .gitignore
make
make clean
```

The only artifact left is the directory `manual/build/`.

PR #1102, opened 2022-06-17
Fix rank bug + add checks

### LINKED ISSUE #1101: Bug in dimensions tag to rst
https://github.com/nexusformat/definitions/blob/ea7d5c4a81001eaf5d7baa536fea888188b0d796/utils/nxdl2rst.py#L270-L290

It should be `if dims and not optional:`.

woutdenolf:
Closes #1101 

* Fix bug: when rank was missing, it was set to the number of dimensions, even when some dimensions were optional.
* Change: a variable rank without symbol was shown as `Rank: .`, now we simply omit the rank
* Add rank and dimension checks. They all pass for the existing classes.

### DIFF
--- a/utils/nxdl2rst.py
+++ b/utils/nxdl2rst.py
@@ -248,9 +248,9 @@ def analyzeDimensions(ns, parent):
     optional = False
     for subnode in node_list:
         # Dimension index (starts from index 1)
-        index = subnode.get('index')
-        if not index or not index.isdigit():
-            continue
+        index = subnode.get('index', '')
+        if not index.isdigit():
+            raise RuntimeError("A dimension must have an index")
         index = int(index)
         if index == 0:
             # No longer needed: legacy way to specify that the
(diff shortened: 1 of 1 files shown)


utils/nxdl2rst.py
```diff
@@ -269,31 +269,43 @@ def analyzeDimensions(ns, parent):
...
+    if rank and dims:
         dims = ', '.join(dims)
         return ' (Rank: %s, Dimensions: [%s])' % (rank, dims)
-    else:
+    elif rank:
         return ' (Rank: %s)' % rank
+    elif dims:
+        dims = ', '.join(dims)
+        return ' (Dimensions: [%s])' % dims
+    else:
```
PeterC-DLS, line 307: Can drop this `else`

woutdenolf: You can get in the `else` by doing this

```xml
<dimensions>
</dimensions>
```

PeterC-DLS: I meant remove `else` and dedent `return ''`

woutdenolf: ah ok, sure why not.

woutdenolf: done

PeterC-DLS (approved):
Thanks

PR #1136, opened 2022-06-26
update gitignore and remove all editor settings from the repo

woutdenolf:


.gitignore
```diff
@@ -1,21 +1,24 @@
...
-manual/makelog.txt
-/build/
+
+# Unknown
 /python/
-impatient-guide/_build
-.DS_Store
 __github_creds__.txt
-
-# Microsoft VSCode editor configuration
```
prjemian, line 20: This is still a good protection to keep.  It prevents unwanted content from creeping in by accident.

prjemian (commented):
I like it.  One detail to keep, as noted.

.gitignore
woutdenolf: That's taken care of by the more general

```
# Hidden files
.*
```

Should stop all editors like eclipse, vscode, pycharm, ...

PeterC-DLS:
Just wondering why you want to be rid of IDE project files?

prjemian (approved):
OK

PR #1133, opened 2022-06-25
Introduce developer tools to replace the util scripts

woutdenolf:
Closes #1144 #1131

Currently: apart from the actual `sphinx` building we have some scripts in `utils` which in combination with `make` files and some custom stuff in the github workflows prepare the files for a  `sphinx` build.

In this PR I propose a python based CLI to do all the non-`sphinx` stuff

https://github.com/woutdenolf/definitions/blob/dev_tools/dev_tools/README.md

Benefits:

* the CLI has a help and has a README so you don't have to dig through make files and python code to understand what needs to be executed to build the documentation
* better / explicit directory management
* constants are defined only once
* building is always out-of-source
* rst files are directly generated by python instead of piping the stdout to a file
* all NXDL rst files are generated by one command, before it was one command per NXDL file
* speed-up of the anchor list files (saved only once, see previous point)
* we no longer need to copy the nxdl files and `utils` for an out-of-source build
* uses `pytest` to minimize test code (`unittest` has lots of boilerplate)
* easier to make sure CI and a local repo do the same thing

The python scripts in `utils` directory become obsolete:

- [x] `build_preparation.py`: `python3 -m dev_tools manual --prepare` (prepare sphinx build)
- [x] `nxdl2rst.py`: `python3 -m dev_tools manual --prepare` (prepare sphinx build)
- [x] `dev_nxdl2rst.py`: `python3 -m dev_tools nxclass nxfluo --print` (for development)
- [x] `test_nxdl.py`: `pytest dev_tools` (for testing)
- [x] `test_nxdl2rst.py`: `pytest dev_tools` (for testing)
- [x] `test_suite.py`: `pytest dev_tools` (for testing)
- [x] `nxdl_desc2rst.py`:  `python3 -m dev_tools manual --prepare` (prepare sphinx build)
- [x] `nxdl_summary.py`: `python3 -m dev_tools manual --prepare` (prepare sphinx build)
- [x] `types2rst.py`: `python3 -m dev_tools manual --prepare` (prepare sphinx build)
- [x] `units2rst.py`: `python3 -m dev_tools manual --prepare` (prepare sphinx build)
- [x] ~~`dev_units2rst.py`:~~
- [x] ~~`local_utilities.py`:~~

Not included (could be but not in this PR)
- `create_release_notes.py`
- `dev_create_release_notes.py`
- `update_copyright_date.py`

To show the difference with `make local`:

```bash
make local

python3 -m dev_tools manual --prepare --build-root build2
[...]

PR #1138, opened 2022-06-26
_punx target already exists

woutdenolf:
https://github.com/nexusformat/definitions/blob/2afc7f8a536251f0645db06611a4a64918da15c5/manual/source/validation.rst#L57

https://github.com/nexusformat/definitions/blob/918df5e60130e8c1d8c5d3a6b2c2eec0128533c1/manual/source/utilities.rst#L163

### DIFF
--- a/manual/source/utilities.rst
+++ b/manual/source/utilities.rst
@@ -160,7 +160,7 @@ The list is not intended to be a complete list of all available packages.
    file; validate
    punx (utility)
 
-.. _punx:
+.. _punx_util:
 
 **punx**
    Python Utilities for NeXus HDF5 files


prjemian (approved):
LGTM

woutdenolf:
I noticed when running the doc build with the -W option. Another thing I introduced in https://github.com/nexusformat/definitions/pull/1133. 

I also found https://github.com/nexusformat/definitions/pull/1139. For that one I did not let failing latex build go through with `($(MAKE) latexpdf LATEXOPTS="--interaction=nonstopmode -f" -C manual || exit 0)`.

Warnings are there for a reason. The `sphinx>=5` requirement in https://github.com/nexusformat/definitions/pull/1139 makes sure we can actually do this. Also there is no reason anymore to run the pdf build twice with a manual in index build in between.

PR #1147, opened 2022-06-29
add CNAME automatically

sanbrock:


prjemian (approved):
Well, now.  This avoids such clutter in the repo.  I'm fine with this!

.github/workflows/publish-sphinx.yml
```diff
@@ -76,6 +76,10 @@ jobs:
         wget https://github.com/nexusformat/code/raw/master/doc/api/NeXusIntern.pdf -O ./build/manual/source/_static/NeXusIntern.pdf
         wget https://github.com/nexusformat/code/raw/master/applications/NXtranslate/docs/NXtranslate.pdf -O ./build/manual/source/_static/NXtranslate.pdf
 
+    - name: Add CNAME to set custom domain
+      run: |
+        echo manual.nexusformat.org > ./build/manual/source/CNAME
```
woutdenolf, line 81: I guess this should be

`echo manual.nexusformat.org > ./build/manual/build/html/CNAME`

woutdenolf (changes requested):
CNAME file is created in the wrong directory I think.

.github/workflows/publish-sphinx.yml
woutdenolf: In fact this is cleaner: https://www.sphinx-doc.org/en/master/usage/extensions/githubpages.html?highlight=githubpages#module-sphinx.ext.githubpages See #1133

woutdenolf: The manual CNAME creation is better than specifying `html_baseurl` because local HTML references resolve locally. So we can stick to

```
echo manual.nexusformat.org > ./build/manual/build/html/CNAME
```

woutdenolf:
No follow up. I'll do it myself.

woutdenolf:
The fix is done here: https://github.com/nexusformat/definitions/pull/1150

PR #1151, opened 2022-06-30
Fix URL's in anchor list and NeXus class rst file

woutdenolf:
Closes  1144


PeterC-DLS (approved):
Looks more correct than mine!

PR #1156, opened 2022-06-30
Define use_application_defaults only once

### LINKED ISSUE #1155: use_application_defaults asignment is wrong
This

https://github.com/nexusformat/definitions/blob/1ac17ef55af511113458699889691d28052a8d58/utils/nxdl2rst.py#L559

should be like this

https://github.com/nexusformat/definitions/blob/1ac17ef55af511113458699889691d28052a8d58/utils/nxdl2rst.py#L681

woutdenolf:
- Closes #1155

### DIFF
--- a/utils/nxdl2rst.py
+++ b/utils/nxdl2rst.py
@@ -22,7 +22,7 @@ from local_utilities import replicate
 
 
 INDENTATION_UNIT = "  "
-listing_category = None
+use_application_defaults = None
 repo_root_path = pathlib.Path(__file__).parent.parent
 WRITE_ANCHOR_REGISTRY = False
 HTML_ROOT = "https://github.com/nexusformat/definitions/blob/main"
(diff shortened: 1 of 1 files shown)


PeterC-DLS (approved):
LGTM

PeterC-DLS:
I have to check `NXcsg` for required logic as it's different in https://raw.githubusercontent.com/nexusformat/definitions/main/legacy_docs/nexus-v2020.10-2020-12-08-0649b8a4.pdf

woutdenolf:
> I have to check `NXcsg` for required logic as it's different in https://raw.githubusercontent.com/nexusformat/definitions/main/legacy_docs/nexus-v2020.10-2020-12-08-0649b8a4.pdf

You need to check before I merge or can I merge?

PeterC-DLS:
Yes, I will double-check a selection of definitions

PeterC-DLS:
Ok, `NXcsg` has no `required` or `minOccurs=1` set so okay with generation. There's a separate issue with whether it should; also `NXsnsevent` and `NXsnshisto` has wrong category. I will fixed latter for release.

PeterC-DLS (approved):
LGTM (though new global is also passed around too - one for the refactor)

woutdenolf:
> LGTM (though new global is also passed around too - one for the refactor)

Yes indeed. I missed that. Not important. `dev_tools` replaces all that.

PR #1163, opened 2022-06-30
Listing category is either "base class" or "application definition"

### LINKED ISSUE #1162: Another contributions bug in the main branch
@prjemian @PeterC-DLS What should the "status" of a contributed definition be? At this point I'm using `dev_tools` to verify the main branch.

Current documentation

image

`main` branch

image

`dev_tools` branch

image

woutdenolf:
Closes #1162 

The main branch had this after #1151

image

That is wrong. The status should mention the category (base or application definition). This PR fixes it back to how it was

image

### Summary of what happened

Originally we had this

```python
     listing_category = {
        "base": "base class",
        "application": "application definition",
        "contributed": "contributed definition",
    }[subdir]
```

which then changed in #1151 to this

```python
     listing_category = {
        "base_classes": "base class",
        "applications": "application definition",
        "contributed_definitions": "contributed definition",
    }[nxdl_subdir]
 ```

Seems legit. The problem however is the `subdir` was in fact NOT the directory name, it was the category! So the name of that variable threw us off. Note that the category is never `contributed` so we always had either `base class` or `application definition` as listing category. In this PR we put back this behavior.

Now in this PR we have

```
listing_category = {
        "base": "base class",
        "application": "application definition",
    }[category]
```

### DIFF
--- a/utils/nxdl2rst.py
+++ b/utils/nxdl2rst.py
@@ -668,10 +668,9 @@ def print_rst_from_nxdl(nxdl_file):
     anchor_registry.nxdl_subdir = nxdl_subdir
 
     listing_category = {
-        "base_classes": "base class",
-        "applications": "application definition",
-        "contributed_definitions": "contributed definition",
-    }[nxdl_subdir]
+        "base": "base class",
+        "application": "application definition",
+    }[category]
 
     use_application_defaults = category == "application"
 
--- a/utils/nxdl2rst.py
+++ b/utils/nxdl2rst.py
@@ -304,12 +304,11 @@ def getDocLine(ns, node):
     return re.sub(r"\n", " ", blocks[0])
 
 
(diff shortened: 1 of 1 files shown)


woutdenolf:
Note: the `dev_tools` branch does things correctly.

PeterC-DLS:
I don't know if the status should say both that it's contributed and which category as well.

woutdenolf:
> I don't know if the status should say both that it's contributed and which category as well.

Yes I like that, but lets keep that for another PR.

prjemian:
- base class, extends [NXobject]
- base class (contributed), extends [NXobject]

informs me that one is different than the other

prjemian:
I agree we can push this idea (identifying contributed status) to the next milestone.

PeterC-DLS (approved):
LGTM

PR #1171, opened 2022-07-07
Remove sub-section headings to fix table of content to fix #1168

PeterC-DLS:
Add custom role to customize rubric directives with font size to match sub-section (H2) headings. Note the font size is not changed in the PDF output

PR #1174, opened 2022-07-08
Make incident_polarisation_stokes optional

### LINKED ISSUE #1173: incident_polarisation_stokes should be optional
#858 makes "incident_polarisation_stokes" required (see [here](https://nexusformat.github.io/definitions/classes/applications/NXmx.html#nxmx-entry-instrument-beam-incident-polarisation-stokes-field)). I guess it should have been changed to `optional="true"`.

_Originally posted by @soph-dec in https://github.com/nexusformat/definitions/pull/858#discussion_r908271132_

phyy-nx:
Fixes #1173

### DIFF
--- a/applications/NXmx.nxdl.xml
+++ b/applications/NXmx.nxdl.xml
@@ -858,7 +858,7 @@
                             <item value="rectangular"/>
                         </enumeration>
                     </field>
-                    <field name="incident_polarisation_stokes" type="NX_NUMBER" recommended="false"
+                    <field name="incident_polarisation_stokes" type="NX_NUMBER" optional="true"
                         deprecated="use incident_polarization_stokes, see https://github.com/nexusformat/definitions/issues/708">
                         <doc>
                             Polarization vector on entering beamline


prjemian (approved):
OK

prjemian:
> This should have been in the recent release, [v2022.06](https://github.com/nexusformat/definitions/releases/tag/v2022.06). @PeterC-DLS, @prjemian, too late?

Per https://github.com/nexusformat/definitions/pull/1171#issuecomment-1178929620:

> I'm thinking that a new release, v2022.07, is the right thing. I'm averse to changing a release afterwards. The new release would gather the work since the v2022.06 release was tagged and also resolve the https://github.com/nexusformat/definitions/pull/858#discussion_r916092837 by @phyy-nx.

@PeterC-DLS : Agree?

phyy-nx:
A bug fix release would be fine.

@yayahjb does this PR resolve your [comment](https://github.com/nexusformat/definitions/issues/1173#issuecomment-1179123242)?  Instead of optional=True we could use minoccurs=0.

prjemian:
> Instead of optional=True we could use minoccurs=0.

Only `optional="true"` is correct for an application definition.  It overrides the default `optional="false"` for application definitions.

PeterC-DLS:
> > This should have been in the recent release, [v2022.06](https://github.com/nexusformat/definitions/releases/tag/v2022.06). @PeterC-DLS, @prjemian, too late?
> 
> Per [#1171 (comment)](https://github.com/nexusformat/definitions/pull/1171#issuecomment-1178929620):
> 
> > I'm thinking that a new release, v2022.07, is the right thing. I'm averse to changing a release afterwards. The new release would gather the work since the v2022.06 release was tagged and also resolve the [#858 (comment)](https://github.com/nexusformat/definitions/pull/858#discussion_r916092837) by @phyy-nx.
> 
> @PeterC-DLS : Agree?

Happy to do a quick 2020.07 release.

prjemian:
> Happy to do a quick 2020.07 release.

Great.  Let's get this PR fixed and merged, then update CHANGES.rst, then tag and release.

phyy-nx:
Ok, `optional=True` is what I proposed here then, so that is good.  I think I'd just like an approval from @yayahjb before this is merged.

yayahjb:
It is a reasonable thing to do until people get more focused on reporting
polarization properly.  Check back in a decade or two. -- Herbert

PeterC-DLS:
> Only optional="true" is correct for an application definition. It overrides the default optional="false" for application definitions.

The logic in [` _get_required_or_optional_text`](https://github.com/nexusformat/definitions/blob/main/dev_tools/docs/nxdl.py#L301) says that `minOccurs = 0` works for application definitions. Does this need amending?

PeterC-DLS:
> > Happy to do a quick 2020.07 release.
> 
> Great. Let's get this PR fixed and merged, then update CHANGES.rst, then tag and release.

Sure, I need to check if there are other deprecated items in all application definitions are correct too: #1175

prjemian:
Here are the definitions in `nxdl.xsd` in this branch:

### group
https://github.com/nexusformat/definitions/blob/4ba70ece79a1f991368d4d4643196074ef06493c/nxdl.xsd#L471-L477

### field

https://github.com/nexusformat/definitions/blob/4ba70ece79a1f991368d4d4643196074ef06493c/nxdl.xsd#L789-L795

### attribute

https://github.com/nexusformat/definitions/blob/4ba70ece79a1f991368d4d4643196074ef06493c/nxdl.xsd#L992-L999

prjemian:
For `group` and `field` in application definitions, the default changes to `true`.

At present, this rule is not represented in `nxdl.xsd`.  Hint: It would be fantastic if someone could describe this rule in `nxdl.xsd`.

soph-dec:
Thank you for the fix!

PeterC-DLS (approved):
LGTM

PR #1177, opened 2022-08-01
Update NXDL version for 2022.07 release

PeterC-DLS:


NXDL_VERSION
```diff
@@ -1,2 +1 @@
-v2022.06
-
+v2022.07
```
prjemian, line 1: Why not v2022.08?

PeterC-DLS: No work has been done since July

prjemian (approved):
Ok

PR #1182, opened 2022-08-25
Add equipment_component to NXtransformations

phyy-nx:
This is an arbitrary identifier of a component of the equipment to which the transformation belongs, such as 'detector_arm' or 'detector_module'. NXtransformations with the same equipment_component label form a logical grouping which can be combined together into a single change-of-basis operation.

Ported in from ImageCIF:
http://www.bernstein-plus-sons.com/software/CBF/doc/cif_img_1.8.4.html#_axis.equipment_component

### DIFF
--- a/base_classes/NXtransformations.nxdl.xml
+++ b/base_classes/NXtransformations.nxdl.xml
@@ -173,6 +173,15 @@
 				depends or the string ".".
 			</doc>
 		</attribute>
+		<attribute name="equipment_component" type="NX_CHAR">
+			<doc>
+				An arbitrary identifier of a component of the equipment to which
+				the transformation belongs, such as 'detector_arm' or 'detector_module'.
+				NXtransformations with the same equipment_component label form a logical
+				grouping which can be combined together into a single change-of-basis
+				operation.
+			</doc>
+		</attribute>
 	</field>
 	<field name="AXISNAME_end" units="NX_TRANSFORMATION" nameType="any" type="NX_NUMBER" minOccurs="0">
 		<doc>


phyy-nx:
A bit of context.  `equipment_component` is an ImageCIF field @yayahjb and I added a [while back](http://cci.lbl.gov/publications/download/CCN_2014_p19.pdf) to support grouping of axes together to form a single change-of-basis operation for a component of a detector.  I had thought this was already in NeXus as I've been using it in DIALS for some time, both when [writing](https://github.com/cctbx/cctbx_project/blob/master/xfel/swissfel/jf16m_cxigeom2nexus.py#L344) and [reading](https://github.com/cctbx/dxtbx/blob/main/src/dxtbx/format/nexus.py#L721-L745) NeXus files.  I only recently discovered that it wasn't actually in NXtransformations.

Also note that during our most recent [Telco](https://www.nexusformat.org/Telco_20220824.html#minutes), @prjemian agreed this would be useful for grouping transformations in multiple-axis goniometers together as well.

phyy-nx:
Here is an example of how we used `equipment_component` for [SwissFEL Jungfrau 16M data](https://doi.org/10.5281/zenodo.3352357).  Given the group `/entry/instrument/ELE_D0/transformations/`, the following NXtransformations are defined:

```
AXIS_RAIL
AXIS_D0 
AXIS_D0Q0 
AXIS_D0Q0M0 
AXIS_D0Q0M0A0 
```
The transformations depend on each other, in bottom to top order.  The first two, `RAIL` and `D0`, have `@equipment_component=detector_arm`, and the rest have it set to `detector_quad`, `detector_module`, and `detector_asic`, respectively.  Thus the first two axes can be grouped together as a single change-of-basis operation.

yayahjb (approved):

mkoennecke:
Why is it not enough to have NXtransformations groups on those components themselves? Or is the NeXus description not fine grained enough to support this?

yayahjb:
If we were talking about robot arms this is the equivalent of providing
descriptors for the various wrists
and shoulders along the arms.  In the case of complex multi-nodule xfel and
synchrotron detectors
the nest of components is just as complex as a robot arm, and providing
meaningful organized
names for those components is very helpful in managing the precise
refinement of their
orientations against the data.

PR #1183, opened 2022-08-29
Allowing the recommended xml attribute for nexus attributes

domna:
It is allowed for fields and groups to be recommended.
However, this is not allowed for attributes.

This PR allows recommended setting for attributes as well.

- close https://github.com/nexusformat/NIAC/issues/140

### DIFF
--- a/nxdl.xsd
+++ b/nxdl.xsd
@@ -989,6 +989,14 @@
 				</xs:documentation>
 			</xs:annotation>
 		</xs:attribute>
+		<xs:attribute name="recommended" use="optional" type="nx:NX_BOOLEAN" default="false" >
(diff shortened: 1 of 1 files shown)


prjemian:
Is there a NXDL file where this is/would be used?  I want to verify the Python code renders this in the documentation.

prjemian (approved):
LGTM

domna:
Here is a file which uses this: https://github.com/FAIRmat-Experimental/nexus_definitions/blob/d92992b279400c169b2f4df9e15063d003041607/contributed_definitions/NXtransmission.nxdl.xml#L81 and it gets rendered in the generated html docs, too.

PR #1200, opened 2022-09-20
publish the docs as-is

prjemian:
- #1119

Issue #1199, opened 2022-09-20
publish current documentation


prjemian:
Publish the manual revision with the new tabbed examples:

- #1117
- #1145

prjemian:
Solved by #1201

Issue #1029, opened 2022-03-08
Deploy documentation on pushing a tag is error prone


woutdenolf:
One can always make mistakes when pushing [tags](https://git-scm.com/book/en/v2/Git-Basics-Tagging) (typo, wrong tag, etc.). Tags could also be pushed to any branch. 

For these reasons I think deploying documentation upon pushing a tag is rather dangerous. I would advocate to limit the `publish-sphinx` workflow to the `main` branch and in addition use an environment variable (e.g. called `DEPLOY`) which is `False` by default. In that way the only way to deploy documentation is by running the `publish-sphinx` workflow manually from [the actions tab](https://github.com/nexusformat/definitions/actions) while setting `DEPLOY=True`.

https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow

https://docs.github.com/en/actions/learn-github-actions/environment-variables

prjemian:
No work on this issue at 2022-06 Code Camp. Is it necessary to resolve this for release of NXDL now?

woutdenolf:
We can keep it for the next release

Issue #1021, opened 2022-03-07
provide HTML documentation for any version of the NeXus standard


prjemian:
It would be nice to see the HTML documentation for any version of the NeXus standard (as much as may be made available).  An example is the documentation of Python as shown:
![image](https://user-images.githubusercontent.com/2279984/157106989-5ea417fa-71aa-4ebf-bc84-a7eeec952ab4.png)

This would be in keeping with the [current release procedure (step 11)](https://github.com/nexusformat/definitions/wiki/Release-Procedure) and benefit the community.

Presently, a PDF copy of a tag is committed to the repository: https://github.com/nexusformat/definitions/tree/main/legacy_docs

_Originally posted by @prjemian in https://github.com/nexusformat/definitions/issues/1016#issuecomment-1061073169_

PeterC-DLS:
So there's a sphinx extension called [sphinx-multiversion](https://github.com/Holzhaus/sphinx-multiversion) and there's an example of a PR from [another project](https://github.com/anexia-it/k8s-anexia-ccm/pull/89/files)

prjemian:
Work will not be complete for this milestone, moving to next milestone.

prjemian:
The [PyData Sphinx theme](https://pydata-sphinx-theme.readthedocs.io) documentation provides this feature with their Sphinx-built docs, using [`switcher.json`](https://github.com/pydata/pydata-sphinx-theme/blob/5d3520540062f231ab1f72cee1a308e27a054466/docs/_static/switcher.json).  A [search](https://github.com/pydata/pydata-sphinx-theme/search?q=switcher.json) of the source repository provides indications how the feature is provided.

PeterC-DLS:
You mean the [version switcher](https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/version-dropdown.html)

prjemian:
What changes are needed to our [publishing workflow](https://github.com/nexusformat/definitions/blob/main/.github/workflows/ci.yaml) to retain different versions of the documentation?  Now, the current version is published to the root of the `gh-pages` branch.  Perhaps we keep that but make a subdirectory for older versions.  That would be additional task(s) for the workflow.  But maybe there is already automation for this.

prjemian:
hint: save a build as as an artifact as [described](https://github.com/pydata/pydata-sphinx-theme/issues/663#issuecomment-1126528235)

prjemian:
[...]

PR #1201, opened 2022-09-21
Deploy by running CI manually

woutdenolf:
Closes 
- #1029
- #1199
- #1200

After this MR

1. the CI runs **without deployment** for every pull request to the main branch and push (tags or commits) to the master branch
2. the CI can only run **with deployment** when triggered manually, with "Deploy documentation" enabled, similar to [this](https://www.softwaretester.blog/detecting-github-workflow-job-run-status-changes).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -4,12 +4,16 @@ on:
   push:
     branches:
       - main   # push commit to the main branch
-    tags:
-      - 'v2*'  # push tag starting with "v2" to the main branch
   pull_request:
     branches:
       - main   # pull request to the main branch
   workflow_dispatch:   # allow manual triggering
+    inputs:
+      deploy:
+        description: 'Deploy documentation'
+        type: boolean
+        required: true
+        default: false
 
 defaults:
   run:
@@ -33,7 +37,7 @@ jobs:
           fetch-depth: 0
 
       - name: Deploy Information
-        if: ${{ startsWith(github.ref, 'refs/tags') && env.python_version == '3.7' }}
+        if: ${{ github.event.inputs.deploy && env.python_version == '3.7' }}
         run: |
           echo "The HTML NeXus User Manual will be pushed to"
           echo " https://github.com/nexusformat/definitions/tree/gh-pages"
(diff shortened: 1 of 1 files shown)


prjemian:
@woutdenolf - The workflow completes now with no errors.  This will resolve the problem I had with #1200 (where the matrix build created contention with writing the `gh-pages` branch from the different jobs).  What work remains to change this from draft status?

woutdenolf:
> What work remains to change this from draft status?

I was trying to test manual triggering but I think we need to merge it before we can actually test it.

prjemian:
Change it from draft and I will review.

prjemian:
Note that manual run [187](https://github.com/nexusformat/definitions/actions/runs/3097598978) was successful because it did not try to push the new `gh-pages` branch from each of the jobs at the same time.

prjemian (approved):
LGTM.  Let's give it a try after merge.

woutdenolf:
Yes but normally you should see something like this when triggering the workflow

image

Instead of "Fail the job?" we should have "Deploy documentation".

woutdenolf:
Ah, now we do

image

prjemian:
_Publish documentation_ only runs from the py3.7 job, not the others, right?

woutdenolf:
Ok I can confirm it works. I triggered the workflow with "Deploy decomentation" enabled and you see that the "Deploy information" step is executed, which means deployment will happen at the end. I cancelled the workflow as I'm not sure we actually want to deploy or not.

https://github.com/nexusformat/definitions/actions/runs/3097668463/jobs/5014607903

woutdenolf:
> Publish documentation only runs from the py3.7 job, not the others, right?

Yes

```
if: ${{ github.event.inputs.deploy && env.python_version == '3.7' }}
```

prjemian:
Also, related to #1021, I wonder if we can create an additional variable and modify the workflow to publish a different version of the manual to a different location?  Possibly a subdirectory of `gh-pages`?

prjemian:
@woutdenolf - Thanks!  @rayosborn - Now I believe we can push the latest documentation.  I'll do this next.

prjemian:
Success!  See the new [tabbed examples](https://manual.nexusformat.org/examples/python/simple_example_write1/index.html).

PR #1206, opened 2022-09-28
Remove obsolete index=0 usage

PeterC-DLS:
Remove from documatation in xsd, add code in doc generator to raise error if index<=0


woutdenolf (approved):
Run `black` but looks good otherwise.

PR #1210, opened 2022-10-04
NXdata: @signal, @auxiliary_signals and @axes refer to direct NXdata children

### LINKED ISSUE #1209: NXdata - @signal doc confusing
https://github.com/nexusformat/definitions/blob/6015a47a1afe792926c4c011f3a089d4e110bb8a/base_classes/NXdata.nxdl.xml#L65-L74

1. So the sentence `The value names an existing child of this group.` means that `The value is equal to the name of field which is the direct child of this group.`? (The fact that this child must exist is clear enough imo)
2. This also applies to @auxiliary_signals and @axes?
3. What does `The child group must itself be a NeXus group` mean?

I'm asking because some people interpret the docs as allowing `@signal` to contain a relative path to a field. We had an h5web [bug report](https://github.com/silx-kit/h5web/issues/1212) because of it. As far as I understand, this is not allowed:

```
data:NXdata
  @signal = "mygroup/mydata"
  mygroup
    mydata: float[100]
```

I suspect `The child group must itself be a NeXus group` may cause the confusion.

woutdenolf:
Closes #1209 

Improved docs to clearly state that the strings in @signal, @auxiliary_signals, @axes refer to

* fields or links
* direct children of the NXdata group
* must exist

### DIFF
--- a/base_classes/NXdata.nxdl.xml
+++ b/base_classes/NXdata.nxdl.xml
@@ -53,10 +53,12 @@
 		<doc>
 			.. index:: plotting
 			
-			Array of strings holding the names of additional signals to
-			be plotted with the default signal (specified by the
-			``signal`` attribute).  Each auxiliary signal needs to be of
-			the same shape as the default signal.
+			Array of strings holding the :ref:`names &lt;validItemName>` of additional
+			signals to be plotted with the default signal (specified by the
+			``signal`` attribute). These field(s) *must* exist and be direct children
+			of this NXdata group.
+			
+			Each auxiliary signal needs to be of the same shape as the default signal.
 			
 			..  NIAC2018:
 			    https://www.nexusformat.org/NIAC2018Minutes.html
@@ -69,9 +71,8 @@
 			.. index:: signal attribute value
 			
 			Declares which NeXus field is the default. 
-			The value is the name of the NeXus field to be plotted.
-			(The value :ref:`names &lt;validItemName>` an existing child
-			of this group. The child group must itself be a NeXus group.)
+			The value is the :ref:`name &lt;validItemName>` of the data field to be plotted.
+			This field must *exist* and be a direct child of this NXdata group.
 			
 			It is recommended (as of NIAC2014) to use this attribute
 			rather than adding a signal attribute to the field.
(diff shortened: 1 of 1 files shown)


sanbrock (approved):

base_classes/NXdata.nxdl.xml
```diff
@@ -83,25 +83,28 @@
...
 			
-			String array that defines the independent data fields used in 
-			the default plot for all of the dimensions of the *signal* field
-			(the *signal* field is the field in this group that is named by
-			the ``signal`` attribute of this group). 
-			One entry is provided for every dimension in the *signal* field.
+			Array of strings holding the :ref:`names &lt;validItemName>` of
+			the independent data fields used in the default plot for all of
+			the dimensions of the :ref:`signal &lt;/NXdata@signal-attribute&gt;`
+			or :ref:`auxiliary signals &lt;/NXdata@auxiliary_signals-attribute&gt;`.
```
PeterC-DLS, line 89: Maybe, "as well as any" instead of "or".

base_classes/NXdata.nxdl.xml
```diff
@@ -83,25 +83,28 @@
...
+			One name is provided for every dimension in the *signal* or *auxiliary signal* fields.
 			
-			The field(s) named as values (known as "axes") of this attribute 
-			*must* exist. An axis slice is specified using a field named 
-			``AXISNAME_indices`` as described below (where the text shown here
-			as ``AXISNAME`` is to be replaced by the actual field name).
+			The *axes* values are the names of fields or links that *must* exist and be direct
+			children of this NXdata group.
+
+			An axis slice is specified using a field named ``AXISNAME_indices``
```
PeterC-DLS, line 96: Do we define what we need by an axis slice? Simpler to say "An axis's dimensional mapping to the signal data field...".

base_classes/NXdata.nxdl.xml
woutdenolf: done

base_classes/NXdata.nxdl.xml
woutdenolf: So if I understand correctly: for each signal dimension you have exactly one axis, but each axis can be used for one or more dimensions?

woutdenolf: Actually I don't understand why `AXISNAME_indices` exists. The `axes` attribute is enough.

woutdenolf: Anyway, `AXISNAME_indices` is not related to this issue. I'' make another issue for this.

PeterC-DLS: If your axis dataset is multi-dimensional, then you need to index into that from the position in your signal dataset. I.e. `signal` has shape `(I,J,K,L)` for `axes=[pressure,temperature,.,.]`, where `pressure` has shape `(I,K,L)` and `temperature` has shape `(J,)`, then `pressure_indices=[0,2,3]` and `temperature_indices=[1]`

woutdenolf: https://github.com/nexusformat/definitions/issues/1212

woutdenolf: Ok I had absolutely no idea that was a possibility. The docs definitely need improving but not in this PR.

PeterC-DLS (approved):

PR #1243, opened 2023-02-08
Typo: NXtomoproc nX -> nY

samtygier-stfc:
Typo in the definition. This is a 3D volume so should be x,y,z.

Spotted by Paul Millar.


FreddieAkeroyd (approved):
I agree, looks like a typo that has been there for a long time

Issue #940, opened 2021-08-17
NXmx: Definitions for multi-channel (thresholds) data


soph-dec:
It is currently undefined how to write data with images from more than one channel (for example multiple thresholds or difference images).

A possible solution we see is to extend the [data field](https://manual.nexusformat.org/classes/applications/NXmx.html#nxmx-entry-data-data-field) by one dimension, i.e. data[nP, ch, i, j, k], where "ch" denotes the channel. However, there are still some open questions regarding this approach:

**How can the "ch" be identified?**
If for example two threshold images are taken and one difference image, how can they be identified? 
  
**How can the "ch" index be connected with associated metadata like flatfield, pixel mask or threshold energy?**
One problem is that this metadata is not given/defined for every channel, so there will likely not be a 1:1 relation. For example, taking an image with difference mode means two threshold images are taken, one with low and one with high threshold energy, corrections like flatfield and pixel masks are applied for each of them and finally the difference image is caclulated from the corrected threshold images. That means not one, but two threshold energies are associated with that data, but no pixel mask or flatfield - but having that metadata might still be useful for the user.

Regarding the first question, we were wondering if perhaps the [dimension scale definition](https://manual.nexusformat.org/classes/base_classes/NXdata.html#nxdata-variable-field) could be extended to also take a list of strings, not only numbers - then we could annotate the channels there. However, we are unsure if this is the right place to store this kind of information since the [axes attribute](https://manual.nexusformat.org/classes/base_classes/NXdata.html#nxdata-data-axes-attribute) is tied to the default plot. Furthermore, this does not solve the second question.

A more general solution for both questions could be the introduction of a new group "channel", where the corresponding indices of the data and also the associated metadata (either directly or also via indices and links to the proper fields) could be stored. A clear advantage of this solution is that it can be easily extended for future purposes.
For example, this could look like this:
```
/data/channel/th_1/@index = 0
/data/channel/th_1/threshold_energy
/data/channel/th_1/flatfield
/data/channel/th_1/pixel_mask
/data/channel/th_2/@index = 1
/data/channel/th_2/threshold_energy
/data/channel/th_2/flatfield
/data/channel/th_2/pixel_mask
/data/channel/diff/@index = 2
/data/channel/diff/threshold_energy_low
/data/channel/diff/threshold_energy_high
```

What are your thoughts regarding these solutions? Do you maybe have other suggestions how to handle these cases?

soph-dec:
[...]

PR #1246, opened 2023-03-13
Allow dimensions in NXdata to be lists of strings

### LINKED ISSUE #945: Allow NXdata dimension variables to contain a list of strings
[NXdata](https://manual.nexusformat.org/classes/base_classes/NXdata.html) groups contain a multidimensional data array with the @axes attribute pointing to fields within the group that hold lists of values to be used as labels for the corresponding dimensions of the dataset. Currently, the values of these "variable" datasets are restricted to NX_NUMBER. I propose that this be relaxed to allow a list of strings, with the number of list items being equal to the size of the corresponding dimension of the main data array (the one that @signal points to).
This proposal is motivated by Dectris, who would like to present data recorded by some of their detectors in terms of "channels". They would like one axis of the dataset to be labelled as "channel" and to label slices of the data array with strings like "low threshold", "high threshold" and "difference".
partly fixes #940

phyy-nx:
Two changes here:
- Add new datatype, NX_CHAR_OR_NUMBER, the union of NX_CHAR and NX_NUMBER.  As [proposed](https://github.com/nexusformat/definitions/issues/945#issuecomment-1247955937) by @prjemian.
- Change AXISNAME in NXdata to allow a list of strings using this new datatype

With this change, part of the channel use case proposed by @soph-dec in #940 can be supported.  Quoting the documentation for the current [FileWriter2 proposal](https://github.com/dectris/documentation/tree/main/filewriter_v2#user-content-multi-channel-data) by Dectris with slight modification (the dimensions of `data`) as an example:

```
data: NXdata
  @signal = "data"
  @axes = ["image_id", "channel", ".", "."]
  @image_id_indices = 0
  @channel_indices = 1
  image_id = [1, ..., nP]
  channel = ["threshold_1", "threshold_2", "difference"]
  data = uint[nP, nC, i, j]
```

Where `nP` is the number of images and `nC` is the number of channels.  Normally the `channel` field, aka [`AXISNAME`](https://manual.nexusformat.org/classes/base_classes/NXdata.html#nxdata-axisname-field), could only be of type NX_NUMBER, similar to the `image_id` field, but a set of English names is preferred to identify the channels.

Note, while the change in AXISNAME was approved by vote in nexusformat/NIAC#97, I do not know what effect this has on existing plotting software.  For example the DIALS image viewer would need to know what channel to display by default.  Also, this could affect the implementation of plot viewing in NeXpy.

We have a Telco tomorrow, so I hope we can discuss those issues then.

Closes #945 and nexusformat/NIAC#97.  Partially addresses #940.


phyy-nx:
Anyone know how to fix the warning that's being generated?  :)

prjemian:
> Anyone know how to fix the warning that's being generated? :)

It's a new data type included in the PR.

phyy-nx:
Hi all, I had a meeting today about this PR and we have some changes (thanks @graeme-winter, @yayahjb, @soph-dec, and @kalcutter).

First, for 2D images, we note that if the data has the dimensions listed below, then existing general case NeXus readers/plotters should work:
- [i,j]: single image, single panel
- [i,j,k]: single image, multi-panel 
- [nP,i,j]: multiple images, single panel
- [nP,i,j,k]: multiple images, multi-panel

Note the number of dimensions above: 2, 3, 3, and 4.  Distinguishing between the two degenerate cases is done by the presence of the NXdetector_module class, which has the fields `data_origin` and `data_size` that can be expanded from 2 to three dimensions to handle the multi-panel case.

OK!  So here are the four new use-cases that are not compatible with existing general-case plot making tools:
- [nC, i,j]: single image, multi-channel, single panel
- [nC,i,j,k]: single image, multi-channel, multi-panel 
- [nP,nC,i,j]: multiple images, multi-channel, single panel
- [nP,nC,i,j,k]: multiple images, multi-channel, multi-panel

With the number of dimensions 3,4,4,5.  Now we have another layer of degeneracy and no way to know how to plot the 2D image.

So the new proposal is to add an attribute `@default_slice`.  You can see the docstring I've proposed in this PR, but briefly, it allows the user to reduce the channel degeneracy, _or any degeneracy where there are more dimensions that the plotter can handle_, by specifying which subset of the data to show.

In the below example, we combine both using strings as array indices and default_slice:

```
data: NXdata
  @signal = "data"
  @axes = ["image_id", "channel", ".", "."]
  @image_id_indices = 0
  @channel_indices = 1
  @default_slice = [".", "difference", ".", "."]
  image_id = [1, ..., nP]
  channel = ["threshold_1", "threshold_2", "difference"]
  data = uint[nP, nC, i, j]
```

So, what do folks think of this alternative?

Note to @mkoennecke, some feedback we received during this meeting was that since the proposed use case of multi-channel data already breaks existing viewers, there's no harm in extending the data type of `AXISNAME_indices`.  Existing software will still read data that doesn't use this feature just fine, but any software that wants to use this feature will need to understand that indices can be strings and which slice to show by default anyway.

phyy-nx:
Any comments on this PR?  Including `@default_slice`?

Also, I'm afraid I don't know enough about the build system to fix the current build error.  Can someone lend a hand?  Thanks!

phyy-nx:
(updated above comment to indicate it's the data type of `AXISNAME` that is being extended to include strings, not `AXISNAME_indices`)

soph-dec:
> Also, I'm afraid I don't know enough about the build system to fix the current build error. Can someone lend a hand? Thanks!

Adding `nxdl:NX_CHAR_OR_NUMBER` to the members of [`primitiveType`](https://github.com/nexusformat/definitions/blob/6d2370324cc8aacd963be57796a8282388089098/nxdlTypes.xsd#L454) solved the build error for me locally.

base_classes/NXdata.nxdl.xml
```diff
@@ -115,6 +115,38 @@
...
+
+				data: NXdata
+					@signal = "data"
+					@axes = ["image_id", "channel", ".", "."]
+					@image_id_indices = 0
+					@channel_indices = 1
+					@default_slice = [".", "difference", ".", "."]
+					image_id = [1, ..., nP]
+					channel = ["threshold_1", "threshold_2", "difference"]
+					data = uint[nP, nC, i, j]
```
soph-dec, line 140:
```suggestion
			  data:NXdata
				@signal = "data"
				@axes = ["image_id", "channel", ".", "."]
				@image_id_indices = 0
				@channel_indices = 1
				@default_slice = [".", "difference", ".", "."]
				image_id = [1, ..., nP]
				channel = ["threshold_1", "threshold_2", "difference"]
				data = uint[nP, nC, i, j]
```
I would suggest changing the spacing like this so it matches the other two examples in NXdata.

phyy-nx:
Thanks @soph-dec!  Checks now pass.

base_classes/NXdata.nxdl.xml
```diff
@@ -115,6 +115,38 @@
...
 	</attribute>
+	<attribute name="default_slice">
+		<doc>
+			Which slice of data to show in a plot by default. This is useful especially for
+			datasets with more than 2 dimensions.
+
+			Should be an array of strings of length equal to the number of dimesions
+			in the data, with the following possible values:
+			* ".": All the data in this dimension should be included
+			* Integer: Only this slice should be used.
```
soph-dec, line 126: Reading this again, I don't think it is completely clear what is meant here. An actual integer or a string containing only an integer? Since the type is NX_CHAR, I suppose it is the latter, but I still think it is a bit confusing. Maybe you could add an example with an integer as well?

phyy-nx: Hi, sorry for the long delay here.  Attributes like the proposed default_slice don't have a NeXus type, but you are right that you can't mix strings and integers in an array.  I've added the following as an example:

```
Alternate version using an integer would look like this (note 2 is a string)::
    @default_slice = [".", "2", ".", "."]
```

A little weird but I think it covers the requirements.

base_classes/NXdata.nxdl.xml
```diff
@@ -115,6 +115,43 @@
...
+		<doc>
+			Which slice of data to show in a plot by default. This is useful especially for
+			datasets with more than 2 dimensions.
+
+			Should be an array of strings of length equal to the number of dimesions
+			in the data, with the following possible values:
+
+			* ".": All the data in this dimension should be included
+			* Integer: Only this slice should be used.
+			* String: Only this slice should be used. Use if AXISNAME_indices is a string
```
soph-dec, line 128: I don't know why I have not noticed this before, but this should be `AXISNAME` not `AXISNAME_indices`.

base_classes/NXdata.nxdl.xml
soph-dec: The rendered docs say it has type `NX_CHAR`, I think this is just the default when you don't specify anything else.
image

I think I would give a complete example or at least show that `channel` no longer contains strings.

woutdenolf: The length of the dimension over which you want to slice is `nC` which is a relatively small number. So why not use `auxiliary_signals` for this?

```
data:NXdata
   @signal = "difference"
   @auxiliary_signals = ["threshold_1", "threshold_2"]
   @axes = ["image_id", ".", "."]
   difference = unit[nP, i, j]
   threshold_1 = unit[nP, i, j]
   threshold_2 = unit[nP, i, j]
```

If for some reason you are forced to have a raw 4D dataset `data = uint[nP, nC, i, j]` then you can create three HDF5 virtual datasets 

```
threshold_1 = VDS(data[:, 0, :, :])
threshold_2 = VDS(data[:, 1, :, :])
difference = VDS(data[:, 2, :, :])
```

soph-dec: Sorry for my very late reply. For us, the channels are comparable to an image with RGB channels, since they belong to the exact same exposure. One wouldn't save a separate image for the red, green and blue channel as well. Therefore, we don't want to split the dataset in multiple three-dimensional datasets. Splitting them would mean their logical connection is no longer visible and it is also no longer a given that the dimensions will match.

Regarding your suggestion with adding multiple VDSs, that would be a possibility if it also makes sense from a processing point of view. Our 4D dataset is a VDS already (the raw data is stored in multiple files) and I do not know if that has any unwanted consequences (for example longer loading times etc.). Do you by chance have experience with this?

phyy-nx: @soph-dec discussed in Telco today:

- In HDF5 user meeting today, it was presented that VDS adds a 5% overhead
- @PeterC-DLS suggests in the past for a different use-case that VDS was a read-back performance problem with a round-robin data saving framework which saves virtually interlaced frames stitched together in a VDS dataset, but he suspects this was addressed in recent HDF5 versions.  Indeed, @yayahjb note that VDS performance is very sensitive to HDF5 version.
- One note is that the proposed NXdetector_channel base class (#1252) provides support not mentioned here of adding additional metadata from NXdetector (such as flatfield or pixelmask) on a per channel basis.
- One correction: "Splitting them would mean their logical connection is no longer visible and it is also no longer a given that the dimensions will match."  The first part is true, but FYI `@auxillary_signal` specifies the additional signals are required to be the shape of the original signal.

woutdenolf: > Our 4D dataset is a VDS already (the raw data is stored in multiple files)

If it is already a virtual dataset, would it work for you to create a virtual dataset for each channel separately instead of creating one virtual dataset for everything? This would also match better with the https://github.com/nexusformat/definitions/pull/1252 mentioned before.

woutdenolf: The reason I'm hesitant about `default_slice` is our previous experience with `AXISNAME_indices`. No Nexus viewer that I know of supports it. The `default_slice` attribute will likely go down the same road: writers will create it but reader will ignore it. Ping @rayosborn

soph-dec: I was looking at how we could add the 3D datasets without loosing the 4D dataset. That would then look something like this (note that we actually also want to use the `AXISNAME_indices` attribute for the `start_time` dataset which relates to the "image_id" axis):

```
data:NXdata
  @signal = "difference"
  @auxiliary_signals = ["threshold_1", "threshold_2"]
  @axes = ["image_id", ".", "."]
  @image_id_indices = 0
  @start_time_indices = 0
  difference = uint[nP, i, j]
  threshold_1 = uint[nP, i, j]
  threshold_2 = uint[nP, i, j]
  image_id = uint[nP]
  start_time = uint[nP]
  data = uint[nP, nC, i, j]
    @axes = ["image_id", "channel", ".", "."]
```

The problem I have now is that I can no longer specify the oder of the channels in the `data` dataset. The `channel` dataset in the NXdata group would vanish, but since the 4D data ist just another "DATA" field of the NXdata group, I can specify the axes attribute but no longer the corresponding `channel` dataset. Do you maybe have a suggestion how to solve that?
Also, I do not know how to specify that the `start_time` corresponds to the "image_id" axes of `data` as well.

I am open to other suggestions, but we really do not want to loose the 4D dataset.

soph-dec: > * One note is that the proposed NXdetector_channel base class ([Detector channels #1252](https://github.com/nexusformat/definitions/pull/1252)) provides support not mentioned here of adding additional metadata from NXdetector (such as flatfield or pixelmask) on a per channel basis.

I'm sorry, I don't quite understand what you mean with this comment. Could you elaborate?

woutdenolf: If you have 3D datasets, why do you still need to 4D dataset? You could just do this:

```
data:NXdata
  @signal = "difference"
  @auxiliary_signals = ["threshold_1", "threshold_2"]
  @axes = ["image_id", "start_time"]
  @image_id_indices = 0
  @start_time_indices = 0
  difference = uint[nP, i, j]
  threshold_1 = uint[nP, i, j]
  threshold_2 = uint[nP, i, j]
  image_id = uint[nP]
  start_time = uint[nP]
```

So this NXdata group has three 3D signals ("difference" is the default).

As for coordinates: the first data dimension has two alternative coordinates (`image_id` and `start_time`) while the other two dimensions do not have coordinates.

FYI see [here](https://hdf5.gitlab-pages.esrf.fr/nexus/nxdata_axes/classes/base_classes/NXdata.html)  for an improved documentation on NXdata (https://github.com/nexusformat/definitions/pull/1213). Long story short: when `AXISNAME_indices` is provided, the order of `@axes` becomes irrelevant (it will just be used by readers to know which datasets contain coordinates).

woutdenolf: If you still need to 4D dataset you can have it of course. It just won't be part of the NXdata definition for the 3D signals.

woutdenolf: If I understand correctly, the data has to be written in a 4D dataset. How about this:

```
+-- entry
    |
    +-- data@NXdata
         |
         +-- difference -> /entry/instrument/detector/difference_channel/data
         +-- threshold_1 -> /entry/instrument/detector/threshold_1_channel/data
         +--  threshold_2 -> /entry/instrument/detector/threshold_2_channel/data
    +-- instrument@NXsignal
        |
        +-- detector@NXdetector
            |
            +-- difference_channel
            |      |
            |     +-- data(3D) -> VDS(/entry/instrument/detector/data[0,...])
            +-- threshold_1_channel
            |      |
            |      +-- data(3D) -> VDS(/entry/instrument/detector/data[1,...])
            +-- threshold_2_channel
            |     |
            |      +-- data(3D) -> VDS(/entry/instrument/detector/data[2,...])
            |
            +-- data (4D: detector dumps the results in here)
```

`NXdetector` has a recommended `data` field so that matches nicely. The new `NXdetector_channel` class could have a recommended `data` field as well for consistency. And finally the NXdata group on the top provided a default plot for this data without the need to change NXdata.

soph-dec: I have thought about the possibility to put the data into the detector/detector_channels, but I still do not see how I can define the order of the channels in the 4D dataset or add the `start_time` axis to it with this solution.

soph-dec: Regarding the 4D dataset: We want to use that because for our use case it is the correct way to represent the data. Adding the 3D datasets would only be a workaround for the limitations of the current NXdata definition.

> FYI see [here](https://hdf5.gitlab-pages.esrf.fr/nexus/nxdata_axes/classes/base_classes/NXdata.html) for an improved documentation on NXdata (#1213).

But I looked at the new documentation and noticed this:
image

Ranks up to 32 are supported? What was the idea for plotting data with higher ranks than 3? Maybe all of this has already been solved before?

In addition, through a link in the new documentation I found [this](https://www.nexusformat.org/NIAC2018Minutes.html#nxdata-plottype--attribute), stating

> NeXus is about how to find and annotate the data to be plotted but not to describe how the data is to be plotted.

Does the proposed `default_slice` fall into "how the data has to be plotted"?

I feel like at this point we might come to a conclusion faster if we have a call and discuss this directly. Would you be open to that? I'll send you an email so we might find a time.

woutdenolf: > Does the proposed default_slice fall into "how the data has to be plotted"?

No I don't think so. It is more equivalent to what `@signal` vs. `@auxiliary_signals` does.

woutdenolf: > Ranks up to 32 are supported

The docs say

> The maximum rank is 32 for compatibility with backend file formats.

There is otherwise no reason to restrict the rank at all. A reader typically plots NXdata as a 1D plot (curve), 2D plot (image) or 3D plot (volume rendering). In addition it will provide slider to browse through the other dimensions,

woutdenolf: From the meeting with @phyy-nx @woutdenolf @soph-dec it is more clear why the request was made. Dectris could easily make 3D datasets or a 4D dataset with 3D virtual datasets. The request is more functional: it doesn't feel right to split up multi-channel images. This would be like creating a format for RGB images and each channel needs to be kept separately.

Still you could argue that the `signal-auxiliary_signals` approach works and it would not require a modification of `NXdata` and hence all `NXdata` reader software. However when you don't have 3 channels but douzens or even hundreds of channels, creating a dataset or virtual dataset for each channels will make the file structure unreadable. One example is the new proposal for the NXxas (https://github.com/nexusformat/definitions/pull/1293) where you essentially want to save 2D data with many columns and each column is a counter.

So for that reason I agree we need to support axes with string type. And since you don't have `@signal` vs. `@auxiliary_signals` we would need `@default_slice`.

base_classes/NXdata.nxdl.xml
```diff
@@ -115,6 +115,51 @@
 			the axes attribute can be omitted.
 		</doc>
 	</attribute>
+	<attribute name="default_slice">
+		<doc>
+			Which slice of data to show in a plot by default. This is useful especially for
+			datasets with more than 2 dimensions.
+
+			Should be an array of strings of length equal to the number of dimesions
```
woutdenolf, line 123: dimesions -> dimensions

woutdenolf:
Is this correct?

```
Should be an array of strings of length equal to the number of dimensions in the data, with the following possible values:

    * ".": All the data in this dimension should be included
    * Integer: Only this slice should be used.
    * String: Only this slice should be used. Use if ``AXISNAME`` is a string array.
```

Isn't the idea to always have integers but saved as NX_CHAR because we need to support `"."`? Why does it depend on the type of `AXISNAME`? That won't work btw because you can have multiple axes that span a single data dimension, and they could be of type NX_NUMBER or NX_CHAR?

phyy-nx:
@woutdenolf I've added the type `NXCHAR_OR_NUMBER` to `@default_slice` which I think resolves your concern.  This matches the type of of `AXISNAME`.   Note I think hdf5 not allow you to mix ints and chars in an attribute array, but I don't think we need to call that out here.

woutdenolf:
Ok thanks for the clarification. So basically `default_slice` has as many values as data dimensions and each value is either

 * `"."`
 * `"<str>"` (for example `"difference"`)
 * `"<int>"` (for example `"2"`)
 * `<int>` (for example `2`)

woutdenolf (approved):

phyy-nx:
> Ok thanks for the clarification. So basically `default_slice` has as many values as data dimensions and each value is either
> 
> * `"."`
> * `"<str>"` (for example `"difference"`)
> * `"<int>"` (for example `"2"`)
> * `<int>` (for example `2`)

Yup, that's my understanding.  Thanks!

phyy-nx:
Conflicts fixed, mostly trivial, with one re-word.  @woutdenolf can you re-check this?  Thanks!

phyy-nx:
This PR is ready for a NIAC vote.  Please vote with an emoji on this comment. Options include thumbs up for yes, thumbs down for no, and anything else to abstain. Thank you.  Voting will close in two weeks.

paulmillar:
Sorry, at the risk of exposing my ignorance, I have a few questions.

> Should be an array of length equal to the number of dimensions in the data

This statement seems somewhat ambiguous to me.

Does the "should" refer to an expectation that `@default_slice` is an array, or does "should" refer to an expectation that if `@default_slice` is an array then that array has a length equal to the number of dimensions in the data?

What does the word "should" actually mean?  Are we following RFC 2119 semantics, or is this meant as a more emphatic statement?
  * Is it valid for `@default_slice` to be something other than an array?  (e.g., a simple value, if data has only one dimension).
  * Is it valid for `@default_slice` to have length different from the number of dimensions in the data?

`@default_slice` is defined as having type `NX_CHAR_OR_NUMBER`.  Does this mean it is valid for `@default_slice` to be an array of integers (rather than an array of strings)?

Given `@default_slice` as an array of strings, how are numerical value distinguished from an axis names?  For example,
  * is `"01"` a slice index or an axis name?
  * is `"-1"` a slice index or an axis name?
  * is `"010"` a slice index with value 10 (decimal), a slice index with value 8 (octal), or is it an axis name?
  * is `"0xf"` a slice index with value 16 (hex) or an axis name?

What are the processing expectations if AXISNAME is an array of NX_CHAR that contains an element `"."` and the `@default_slice` array contains the element `"."`?

What are the processing expectations if AXISNAME is an array of NX_CHAR, the *first* element is `"2"` and the `@default_slice` array contains the element `"2"`?

What are the processing expectations if AXISNAME is an array of NX_CHAR with two elements having the same value and that value is specified in `@default_slice` ?

phyy-nx:
Thank you @paulmillar!  Note that at this point in we are in the middle of a vote so it is too late to make changes to this PR, otherwise it would be unfair to those that have already voted.  That has been the protocol at live NIAC meetings, where once a vote has been called for and seconded it has to finish.  I don't see why it wouldn't be the case for an online vote as well.  Indeed this PR has already been discussed and reviewed so I think that's fair.

THAT SAID, your questions are good, and I think you've identified one change I would have made in this PR.  However, I think the change is clarification, and not substance, so I think it would be fine as a separate PR that doesn't need a vote.  More below.

> What does the word "should" actually mean? Are we following RFC 2119 semantics, or is this meant as a more emphatic statement?

Should is a requirement.  I had never seen RFC 2119 before, thanks for pointing it out!  Just to check, I reviewed NXmx to find usages of the word "should" and found 10.  Each could equally be considered ambiguous based on RFC 2119.  I'd prefer a separate review of the usage of imperatives outside of this PR.

> Is it valid for `@default_slice` to be something other than an array? (e.g., a simple value, if data has only one dimension).

Yes to your specific example, a single value if the data has only one dimension.  My understanding of NeXus is that this doesn't generally need to be specified, but I could be wrong here.

> Is it valid for `@default_slice` to have length different from the number of dimensions in the data?

No.  See "should" above.

> `@default_slice` is defined as having type `NX_CHAR_OR_NUMBER`. Does this mean it is valid for `@default_slice` to be an array of integers (rather than an array of strings)?

`NX_CHAR_OR_NUMBER` is a new term defined here as "Any valid character string or NeXus number representation".  So a list of integers is inherently included.  Technically this should be NXCHAR_OR_INT I suppose...

> Given @default_slice as an array of strings, how are numerical value distinguished from an axis names?

This is a good question.  The second example lists `@default_slice = [".", "2", ".", "."]` as a possibility.  I had presumed someone writing software to support this would attempt a lookup for "2" in the `AXISNAME` `channel`, not find it, and fall back to see if "2" can be parsed as a number.  But this could be more clear in the documentation.  Note you cannot mix integers and strings in an HDF5 array, so using "." in an array with integers for `@default_slice` has to be done this way.

> What are the processing expectations if `AXISNAME` is an array of `NX_CHAR` that contains an element "." and the `@default_slice` array contains the element "."?

"." in `@default_slice` has a reserved meaning so the elements of `AXISNAME`don't matter.  If you have "." as a value in `AXISNAME`, you won't be able specify it in `@default_slice`.

> What are the processing expectations if `AXISNAME` is an array of `NX_CHAR`, the first element is "2" and the `@default_slice` array contains the element "2"?

Order of operations states that "2" will be searched for in `AXISNAME` and the index of "2" in `AXISNAME` will then be the slice used for `@default_slice`

> What are the processing expectations if `AXISNAME` is an array of `NX_CHAR` with two elements having the same value and that value is specified in `@default_slice`?

This is undefined.

I really appreciate these comments.  Of these, I'd like to add "order of operations" to `@default_slice`, to clarify that `AXISNAME` should be searched before a CHAR array is attempted to be parsed.  However, I'd like to do that in a separate PR, as I think it's clarification, and not really a change that needs a vote.  Do you agree?

For the rest, I think we have to be careful to avoid overdefining terms.  I don't think we need to add these clarifications to the NXdata spec unless someone asks for them.  Do you agree?

prjemian:
Since "2" is not a [valid name for a group or field](https://manual.nexusformat.org/datarules.html#index-0), that logic could be included and the search for such an HDF5 address avoided.

prjemian:
Same for ".".

phyy-nx:
Ah good stuff @prjemian.  Thanks.

phyy-nx:
Vote has passed, thanks all.

phyy-nx:
@paulmillar I believe @prjemian answered your questions best when he pointed out that "2" and "." are not legal `AXISNAME`s.  Is there anything in your comment you think should be made into an issue?

PR #1252, opened 2023-04-06
Detector channels

### LINKED ISSUE #940: NXmx: Definitions for multi-channel (thresholds) data

phyy-nx:
This PR adds NXdetector_channel as a base class and adds it as optional for NXmx.  @soph-dec and @kalcutter I put in placeholder documentation.  Can you review and comment?

See original Dectris proposal:
https://github.com/dectris/documentation/tree/main/filewriter_v2#user-content-multi-channel-data

Closes #940
Requires #1246 to be merged first

Co-authored-by: Sophie Hotz <sophie.hotz@dectris.com>


soph-dec:
Thank you for the PR Aaron! First glance is looking good, I will review everything in detail next week.

base_classes/NXdetector_channel.nxdl.xml
```diff
@@ -0,0 +1,103 @@
...
+    example in NXdata), the NXdetector will have a series of NXdetector_channel groups, one for each
+    channel, named CHANNELNAME_channel.
+
+    Example, given these axes in the NXdata group:
+      @axes = ["image_id", "channel", ".", "."]
+
+    And this list of channels in the NXdata group:
+      channel = ["threshold_1", "threshold_2", "difference"]
+
+    The NXdetector group would have three NXdetector_channel groups:
```
soph-dec, line 64:
```suggestion
    Example, given these axes in the NXdata group::
    
      @axes = ["image_id", "channel", ".", "."]

    And this list of channels in the NXdata group::
    
      channel = ["threshold_1", "threshold_2", "difference"]

    The NXdetector group would have three NXdetector_channel groups::
    
```
Putting these examples in a code block makes it easier to read in my opinion and it matches other parts of the documentation better (for example in NXdata).

The html version then looks like this
image

base_classes/NXdetector_channel.nxdl.xml
soph-dec, line 1: I think we should mention here somewhere that any field from NXdetector that makes sense on a channel-level might be used in NXdetector_channel. The question is, should this be allowed already or should people be encouraged to open an issue/a PR in case they want more fields for a channel? What do you think?

soph-dec (commented):
I think we should mention NXdetector_channel in NXdetector as well. I guess something like this might do?
```
  <group name="CHANNELNAME_channel" type="NXdetector_channel">
      <doc>
          Group containing the description and metadata for a single channel from a multi-channel
          detector.

          Given an :ref:`NXdata` group linked as part of an NXdetector group that has an axis with
          named channels (see the example in :ref:`NXdata &lt;/NXdata@default_slice-attribute&gt;`),
          the NXdetector will have a series of NXdetector_channel groups, one for each channel,
          named CHANNELNAME_channel.
      </doc>
  </group>
```

base_classes/NXdetector_channel.nxdl.xml
phyy-nx: I'd be more inclined to want to go through the PR process to add fields that are intended to be for general use if it were just a few of them.  But do you have some suggestions as to which specific fields could be useful?  We have `threshold_energy`, `flatfield` and `pixel_mask`.  Thinking forward, what could you imagine needing?

soph-dec: Hmm, I looked through the fields in NXdetector again and I guess the flatfield and pixelmask related ones (`flatfield_applied`, `flatfield_errors`, `pixel_mask_applied`) could also be part of this PR. Depending on whether `saturation_value` and `underload_value` should contain corrected values, this could also be per channel (since the flatfields are per channel). I am not sure if that is how those values should be interpreted or not though.

From our side, I don't see immediate need for something else. I also think it's the better approach to go through the PR process should need for more fields arise in the future.

phyy-nx: Done

yayahjb (approved):
Worth doing

base_classes/NXdetector_channel.nxdl.xml
```diff
@@ -0,0 +1,158 @@
...
+            >
+
+  <symbols>
+    <doc>
+      These symbols will be used below to illustrate the coordination of the rank and sizes of datasets and the 
+      preferred ordering of the dimensions. Each of these are optional (so the rank of the datasets 
+      will vary according to the situation) and the general ordering principle is slowest to fastest.
+      The type of each dimension should follow the order of scan points, detector output (e.g. pixels),
+      then time-of-flight (i.e. spectroscopy, spectrometry). Note that the output of a detector is not limited 
+      to single values (0D), lists (1D) and images (2), but three or higher dimensional arrays can be produced
```
soph-dec, line 40:
```suggestion
      to single values (0D), lists (1D) and images (2D), but three or higher dimensional arrays can be produced 
```

base_classes/NXdetector_channel.nxdl.xml
```diff
@@ -0,0 +1,158 @@
...
+        threshold_1_channel:NXdetector_channel
+          threshold_energy = float
+          flatfield = float[i, j]
+          pixel_mask = uint[i, j]
+        threshold_2_channel:NXdetector_channel
+          threshold_energy = float
+          flatfield = float[i, j]
+          pixel_mask = uint[i, j]
+        difference_channel:NXdetector_channel
+          threshold_energy = float[2]
```
soph-dec, line 79: Just a thought, you could add the new fields you just added (flatfield_applied, etc.) in the example here, but I don't think it is necessary.

phyy-nx:
Conflicts fixed, no changes to this branch.

Reviewers might find it easier to use this link to compare the `detector_channel` branch to the `dimension_as_char` branch until #1246 is merged:
https://github.com/nexusformat/definitions/compare/dimension_as_char...detector_channel

phyy-nx:
Vote has passed.  Thanks all.

PR #1255, opened 2023-04-25
Add Electron Source as one of the NXsource types

phyy-nx:
Part of work @dagewa has been doing on making some MicroED NeXus files.

### DIFF
--- a/base_classes/NXsource.nxdl.xml
+++ b/base_classes/NXsource.nxdl.xml
@@ -56,7 +56,7 @@
 			<item value="Ion Source" />
 			<item value="UV Plasma Source" />
 			<item value="Metal Jet X-ray" />
-			<item value="Electron Microscope" />
+			<item value="Electron Source" />
 		</enumeration>
 	</field>
 	<field name="probe">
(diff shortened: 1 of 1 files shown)


sanbrock:
An Electron Microscope is a complex device, maybe "electron beam" would be more appropriate. Note that details of an EM can be registered in [NXem](https://manual.nexusformat.org/classes/contributed_definitions/NXem.html)

phyy-nx:
Feedback from the Telco: we ask @sanbrock to meet with @dagewa to meet and come up with a reconciled proposal between NXem and this PR, specifically "Electron Microscope" vs. "Electron beam" vs. "Electron gun" vs. "Cathode ray", or similar.  The idea is to have a proposal in time for the June code camp NIAC meeting so it can be voted on.  Thanks!!

PeterC-DLS:
Also note that there are many types of electron guns...

phyy-nx:
Updated now with changes from #1268

dagewa:
I guess we should change the title? (I don't have edit access)

yayahjb (approved):
Worth doing

PeterC-DLS:
This has been duplicated by #1268

PR #1256, opened 2023-05-03
Add favicon to HTML pages of manual

PeterC-DLS:
Remove extra copy of icon


rayosborn:
I downloaded the favicon, and it appears to be striped. I also believe that it is now recommended to have larger (e.g., 128x128) images, whereas this is 32x32. This is the [favicon file](https://github.com/nexpy/nexpy/blob/main/doc/source/favicon.ico) we use for NeXpy documentation.

PeterC-DLS:
Thanks for that, Ray, or the one from our main site: https://github.com/nexusformat/wiki/blob/master/public/favicon.ico . Is the `www/` directory used anywhere these days?

prjemian:
@FreddieAkeroyd Can you answer this?

> Is the `www/` directory used anywhere these days?

rayosborn:
The current website favicon is certainly better than the striped version although it's still only 32x32. Apparently, larger favicons are now used to generate mobile browser shortcuts, so sizes of up to 180x180 (and even 512x512) are recommended on different websites. 

We also don't have consistent branding of our websites at the moment. The [main landing page](http://www.nexusformat.org) looks completely different from the [manual](https://manual.nexusformat.org/). Deciding which we prefer might govern the choice of icon. Not a high priority perhaps, but it would look more professional if we had more homogeneous styling. Future code camp?

woutdenolf (commented):
`manual` and `impatient-guide` are two independent documentations with their own sphinx configuration file. Can we put shared resources (currently only the flavicon) somewhere else (not in ./manual neither in ./impatient-guide)?

woutdenolf (approved):
LGTM

PR #1262, opened 2023-05-18
Add bit-depth-image

### LINKED ISSUE #1258: NXmx should (or should not?) mention bit_depth_image
Looking at 

https://manual.nexusformat.org/classes/applications/NXmx.html

I see that `bit_depth_image` is not included but

https://www.dectris.com/support/downloads/header-docs/nexus/

DECTRIS have this in their headers and are using it, and if you are looking at the data from the 0mq stream or file writer the _format_ for the data corresponds to this not the readout (which could be fewer bits)

Should this be in the standard? It does make a certain amount of sense e.g. `bit_depth_readout=16` would indicate that a pixel of value `0x10006` is invalid.

graeme-winter:
Fixes #1258 by including it, using effectively the same definition as is currently in the field from DECTRIS

### DIFF
--- a/applications/NXmx.nxdl.xml
+++ b/applications/NXmx.nxdl.xml
@@ -666,6 +666,12 @@
                         </doc>
                     </field>
 
+                    <field name="bit_depth_image" type="NX_INT" recommended="true">
+                        <doc>
+                            The number of bits per pixel are saved to the image data.
+                        </doc>
+                    </field>
+
                     <field name="detector_readout_time" type="NX_FLOAT" units="NX_TIME"
                         minOccurs="0">
                         <doc>


sanbrock:
It could also be considered to extend NXdetector, too.

graeme-winter:
It is _in use now_ in the field and was a specific proposal - fast vs. slow are encoded in the existing standard, endian is (I think?) already handled elsewhere, and the dtype is (I think) explicitly encoded in HDF5 so we don't need to have the signed-ness really?

Anyway, I look forward to the outcome of the vote on adopting a tag which is already in use.

ndevenish:
> Should we also add bit sign, fast vs. slow, endian, etc.? We think that discussion could be had outside of this issue though.

Certainly signedness has caused confusions in dealing with this issue, as when trying to handle these mismatches we've had to make educated guesses as to whether we need to convert or truncate when converting to/from the declared bit depth. Endian is, I suppose, strictly missing, but I don't think we've seen any live examples of this being important.

But yes, somewhat orthogonal to the issue of this specific field.

phyy-nx:
Discussion from Code Camp:

- I showed [DIALS code](https://github.com/cctbx/dxtbx/blob/main/src/dxtbx/nexus/__init__.py#L517-L519) that uses bit_depth_image to handle safe/unsafe integer conversions between NumPy and "flex", which is the internal DIALS/cctbx memory management implementation.
- Unofficial straw poll in the Code Camp indicated we were in favor of moving this to NXdetector as well
- Update data type from NX_INT to NX_POSINT (no change to Dectris or DIALS)
- Modify language in NXdetector [bit_depth_readout](https://manual.nexusformat.org/classes/base_classes/NXdetector.html#nxdetector-bit-depth-readout-field): change "must not" to "might not"

I'll get these changes ready.  Thanks again to @graeme-winter.

phyy-nx:
Changes made in new PR #1284

PR #1278, opened 2023-06-16
NXtransformations: clarify that these are active transformations + example

### LINKED ISSUE #1269: NXtransformations: chaining vs. active/passive transformations
I assume that the values in NXtransformations define [*active transformations*](https://en.wikipedia.org/wiki/Active_and_passive_transformation) (i.e. coordinate transformations) in analogy with all other transformations in the Nexus standard. This needs to be specified explicitly in the docs. However when looking at chaining, they might actually be passive transformations.

The [NXtransformations docs](https://manual.nexusformat.org/classes/base_classes/NXtransformations.html#nxtransformations) say

> when T1 depends on T2 and that in turn T2 depends on T3, the final transformation is T3.T2.T1

Suppose we have two objects, one with `R4.R3.R2.T1` and the other with `R5.R3.R2.T1`. We could define two NXtransformations groups which define the full stack of transformations

```
entry
    instrument
        object1
            depends_on=position1/T1
                position1:NXtransformations
                    T1
                        @depends_on=R2
                    R2
                        @depends_on=R3
                    R3
                        @depends_on=R4
                    R4
        object2
            depends_on=position2/T1
                position1:NXtransformations
                    T1
                        @depends_on=R2
                    R2
                        @depends_on=R3
                    R3
                        @depends_on=R5
                    R5
```

Note that the first transformation (T1) occurs first if you think about applying the transformations to the objects one by one in physical space. This is assuming NXtransformations are active transformation.

The [NXtransformations docs](https://manual.nexusformat.org/classes/base_classes/NXtransformations.html#nxtransformations) also say

> Following the chain may also require following depends_on links to transformations outside, for example to a common base table

Since our two objects have a common base transformation, lets try to use common NXtransformations group

```
entry
    instrument
        baseposition:NXtransformations
            T1
                @depends_on=R2
            R2
                @depends_on=R3
            R3
        object1
            depends_on=position1/R4
                position1:NXtransformations
                    R4
                        @depends_on=/entry/instrument/baseposition/T1
        object2
            depends_on=position2/R5
                position1:NXtransformations
                    R5
                        @depends_on=/entry/instrument/baseposition/T1
```

However this means I have `R3.R2.T1.R4` and `R3.R2.T1.R5` which is not what I want. In other words using a common base NXtransformations cannot work.

However when NXtransformations are passive, it would work

```
entry
    instrument
        baseposition:NXtransformations
            R3
                @depends_on=R2
            R2
                @depends_on=T1
            T1
        object1
            depends_on=position1/R4
                position1:NXtransformations
                    R4
                        @depends_on=../baseposition/R3
        object2
            depends_on=position2/R5
                position1:NXtransformations
                    R5
                        @depends_on=../baseposition/R3
```

which would give `R4.R3.R2.T1` and `R5.R3.R2.T1`. In other words using a common base NXtransformations works.

Note that the last transformation (T1) occurs first if you think about applying the transformation to the objects one by one in physical space. This is assuming NXtransformations are passive transformation.

As a side question: is `../baseposition/R3` valid? Or does it need to be `entry/instrument/baseposition/R3`? Or is both allowed?

woutdenolf:
Closes #1269 

Rendered version of this PR:

https://hdf5.gitlab-pages.esrf.fr/nexus/nxtransformations_active/classes/base_classes/NXtransformations.html


base_classes/NXtransformations.nxdl.xml
```diff
@@ -108,6 +111,87 @@
...
+		          polar=5
+		            @depends_on=distance
+		            @units=degrees
+		            @vector=[0 1 0]
+		          distance=-10
+		            @depends_on=../../coordinate_system/beam
+		            @units=cm
+		            @vector=[1 0 0]
+		      horizontal:NXdetector
+		        position:NXtransformations
```
woutdenolf, line 174: missing `depends_on=position/azimuth`

phyy-nx:
Update from code camp: we are saving the [active transformations](https://en.wikipedia.org/wiki/Active_and_passive_transformation).  @woutdenolf will update this PR to accommodate

Example given in code camp: imagine a point detector with only one transformation, call it a rail.  Here's the NXtransformations field:
```
rail = 10
@vector = 0, 0, 1
@offset = 0, 0, 0
@transformation_type = translation
```
NXdetector will depend on rail.  This transformation can be written in homogenous coordinates:
```
1	0	0	0
0	1	0	0
0	0	1	10
0	0	0	1
```

Call that matrix A.  The equation to move from pixel space to laboratory space (where pixel space is just 0,0,0,1 as a homogenous vector since it's a point detector):

```
lab space = A * pixel space = [0,0,10,1]
```

So the final coordinates of the detector in lab space are 0,0,10

base_classes/NXtransformations.nxdl.xml
woutdenolf: fixed

woutdenolf:
Changes: NXtransformations are *active* transformations

base_classes/NXtransformations.nxdl.xml
```diff
@@ -108,6 +116,92 @@
...
+			* positioned in the XY-plane above the XZ-plane
+
+		.. code-block::
+
+		  entry:NXentry
+		    coordinate_system:NXtransformations
+		      beam=NaN  # value is never used
+		        @depends_on=gravity
+		        @vector=[1 0 0]          # X-axis points in the beam direction
+		      gravity=NaN  # value is never used
```
woutdenolf: @PeterC-DLS @phyy-nx  the values of directional Nxtransformations fields is never used right? So it can be NaN.

base_classes/NXtransformations.nxdl.xml
```diff
@@ -66,12 +66,21 @@
...
+		and that in turn depends on :math:`T_3`, the final *active* transformation
+		matrix :math:`T_f` is
 
-		.. math:: T_f = T_3 T_2 T_1
+		.. math:: T_f = T_1 . T_2 . T_3
+
+		For example when positioning a flat detector, its pixel positions in this laboratory
+		reference frame can be calculated by
+
+		.. math:: X_{\text{lab}} = T_1 . T_2 . T_3 . X_{\text{pixel}}
```
woutdenolf: @PeterC-DLS @phyy-nx  this is now the chain of active transformations as encountered when an HDF5 reader follows the *depends-on* directives. Makes more sense now.

base_classes/NXtransformations.nxdl.xml
```diff
@@ -108,6 +116,93 @@
...
+		            @units=cm
+		            @vector=[1 0 0]
+		      transmission:NXdetector
+		        depends_on=position/azimuth
+		        position:NXtransformations
+		          distance=20            # move downstream from the sample
+		            @depends_on=../../coordinate_system/beam
+		            @units=cm
+		            @vector=[1 0 0]
+
```
PeterC-DLS, line 205: Add explicit matrices for calculating detector positions:
```
r_lab = T_x(dist) R_y(-pol) R_x(-azi) r_det
```
where T<sub>x</sub> is translation in x, R<sub>x</sub> is rotation about x, etc.

In fact, this shows that the `depends_on` chain is wrong and should be `detector` depends on `distance` (as the detector is mounted on an arm of that length) depends on `polar` depends on `azimuth` depends on `.` so that
```
r_lab =  R_x(-azi) R_y(-pol) T_x(dist) r_det
```

base_classes/NXtransformations.nxdl.xml
woutdenolf: See https://github.com/nexusformat/definitions/pull/1292

base_classes/NXtransformations.nxdl.xml
```diff
@@ -66,12 +66,21 @@
 		specification.
 
 		For a chain of three transformations, where :math:`T_1` depends on :math:`T_2`
-		and that in turn depends on :math:`T_3`, the final transformation :math:`T_f` is
+		which in turn depends on :math:`T_3`, the final *active* transformation
+		matrix :math:`T_f` is
 
-		.. math:: T_f = T_3 T_2 T_1
+		.. math:: T_f = T_1 . T_2 . T_3
```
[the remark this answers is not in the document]
woutdenolf: T_1 depends on T_2 depends on T_3. This means that when physically positioning the object, starting at the origin, T_3 needs to be applied **first** to the coordinates of the object. So `X'  = T_f . X = T_1 . T_2 . T_3 . X` is the correct order.

I finally thought things were clear ... This really shows how hard these NXtransformations are.

base_classes/NXtransformations.nxdl.xml
```diff
@@ -108,6 +116,92 @@
...
+		          @transformation_type=rotation
+		          @vector=[0 1 0]
+		          @units=degrees
+		        chi=90
+		          @depends_on=rotation_angle
+		          @transformation_type=rotation
+		          @vector=[0 0 1]
+		          @units=degrees
+		        rotation_angle=30
+		          @depends_on=../../coordinate_system/beam
```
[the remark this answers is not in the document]
woutdenolf: But I want to reference the frame defined in ../../coordinate_system

base_classes/NXtransformations.nxdl.xml
```diff
@@ -66,12 +66,21 @@
...
+		which in turn depends on :math:`T_3`, the final *active* transformation
+		matrix :math:`T_f` is
 
-		.. math:: T_f = T_3 T_2 T_1
+		.. math:: T_f = T_1 . T_2 . T_3
+
+		For example when positioning a flat detector, its pixel positions in the laboratory
+		reference frame (McSTAS coordinate frame by default) can be calculated by
+
+		.. math:: X_{\text{lab}} = T_1 . T_2 . T_3 . X_{\text{pixel}}
```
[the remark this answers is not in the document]
woutdenolf: The example of @phyy-nx during the code camp was

X_lab = T_rail . T_quadrant . T_module . X_pixel

where T_module is the *first* coordinate transformation

T_rail depends_on T_quadrant depends_on T_module

where T_module is the *last* in the depends_on chain.

base_classes/NXtransformations.nxdl.xml
woutdenolf: I will add the formulae.

woutdenolf: For a point detector we always have r_det=[0,0,0,1]. The first thing you need to do is a translation, rotating a point doesn't change anything. This is when may depend_on chain does I think.

woutdenolf: The first transformation you encounter in the _depend_on_ chain of the horizontal and vertical point detector is `R_x(-azi)` (active transformation matrix). However as I understand it, this is actually the **last** that should be applied to the coordinates r_det.

```
r_lab = R_x(-azi) R_y(-pol) T_x(dist) r_det
```

woutdenolf: The coordinates of the three detectors in the lab frame

```
+X: beam
-Z gravity
transmission (in the beam): [20.  0.  0.]
vertical (in XZ-plane): [9.96194698 0.         0.87155743]
horizontal (in XY-plane): [1.09397408e+01 1.14981310e+00 7.04057464e-17]
```

These are the formulae:

```python
import numpy


def trnx(t: float) -> numpy.ndarray:
    """Active transformation: translation along X"""
    M = numpy.eye(4)
    M[0, -1] = t
    return M


def trny(t: float) -> numpy.ndarray:
    """Active transformation: translation along X"""
    M = numpy.eye(4)
    M[1, -1] = t
    return M


def trnz(t: float) -> numpy.ndarray:
    """Active transformation: translation along X"""
    M = numpy.eye(4)
    M[2, -1] = t
    return M


def rotx(a: float) -> numpy.ndarray:
    """Active transformation: rotation around X"""
    a = numpy.radians(a)
    cosa = numpy.cos(a)
    sina = numpy.sin(a)
    return numpy.array(
        [[1, 0, 0, 0], [0, cosa, -sina, 0], [0, sina, cosa, 0], [0, 0, 0, 1]]
    )


def roty(a: float) -> numpy.ndarray:
    """Active transformation: rotation around Y"""
    a = numpy.radians(a)
    cosa = numpy.cos(a)
    sina = numpy.sin(a)
    return numpy.array(
        [[cosa, 0, sina, 0], [0, 1, 0, 0], [-sina, 0, cosa, 0], [0, 0, 0, 1]]
    )


def rotz(a: float) -> numpy.ndarray:
    """Active transformation: rotation around Z"""
    a = numpy.radians(a)
    cosa = numpy.cos(a)
    sina = numpy.sin(a)
    return numpy.array(
        [[cosa, -sina, 0, 0], [sina, cosa, 0, 0], [0, 0, 1, 0], [0, 0, 0, 1]]
    )


def to_homogeneous(coordinates: numpy.ndarray) -> numpy.ndarray:
    return numpy.append(coordinates, 1)


def from_homogeneous(coordinates: numpy.ndarray) -> numpy.ndarray:
    coordinates = numpy.asarray(coordinates)
    return coordinates[:-1] / coordinates[-1]


if __name__ == "__main__":
    origin = [0, 0, 0]

    print("+X: beam")
    print("-Z gravity")

    transmission = trnx(20) @ to_homogeneous(origin)
    print("transmission (in the beam):", from_homogeneous(transmission))

    vertical = rotx(-0) @ roty(-5) @ trnx(10) @ to_homogeneous(origin)
    print("vertical (in XZ-plane):", from_homogeneous(vertical))

    horizontal = rotx(-90) @ roty(-6) @ trnx(11) @ to_homogeneous(origin)
    print("horizontal (in XY-plane):", from_homogeneous(horizontal))
```

woutdenolf:
@PeterC-DLS We established so far that the vector+value of an NXtransformations field describe an active transformation matrix. What still needs to be established is the order in which a reader encounters the NXtransformations fields when he follows the `depends_on` directives.

My interpretation was this: when we say T_1 "depends on" T_2 it means T_2 needs to be applied first before you can apply T_1.

Our description says the opposite is true: when we say T_1 "depends on" T_2 it means T_1 needs to be applied first before you can apply T_2.

Maybe it becomes more clear when connecting two NXtransformation groups. The first group is called "base" and describes the overall active transformation as a result of a stack of _base_ motors. The second group is called "positioner" and describes the overall active transformation as a result of a stack of _positioning_ motors. Suppose the positioning motors are physically stacked on top of the _base_ motors. This means you first need to apply the _positioning_ transformations and then the _base_ transformations. In this case your description makes more sense: "positioner" depends on "base" and _positioning_ needs to be applied before _base_.

So the words "depends on" can be interpreted in two ways:

1. The mathematical sense (what I described): when transformation A depends an transformation B, B needs to happen before A can happen so B needs to be applied first.
2. The stack-of-motors sense (what you described): when motor A depends on motor B, motor A is stacked on top of B and A needs to  be applied first.

So @PeterC-DLS if we agree that the stack-of-motors sense (what you described) is what was intended, I will adapt the PR accordingly.

woutdenolf:
I will also add this sentence to make the intention clear.

> When we say transformation A "depends on" transformation B we mean that the physical motor the realizes A is _stacked on top of_ the motor that realizes B. So the activate coordinate transformation A needs to be applied to a coordinate before B.

base_classes/NXtransformations.nxdl.xml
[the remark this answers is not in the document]
woutdenolf: Correction, see https://github.com/nexusformat/definitions/pull/1278#issuecomment-1600286106

It probably was this:

X_lab = T_rail . T_quadrant . T_module . X_pixel

T_module depends on T_quadrant which in turn depends in T_rail.

base_classes/NXtransformations.nxdl.xml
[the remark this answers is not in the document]
woutdenolf: Correction, see https://github.com/nexusformat/definitions/pull/1278#issuecomment-1600286106

base_classes/NXtransformations.nxdl.xml
woutdenolf: Correction, see https://github.com/nexusformat/definitions/pull/1278#issuecomment-1600286106

The formulae are correct, but the order in which the transformation is saved in HDF5 is the reverse as what I though. For example:

```
horizontal = rotx(-90) @ roty(-6) @ trnx(11) @ to_homogeneous(origin)
```

which means trnx(11) comes first when following the `depends_on` chain.

PeterC-DLS:
Yes, stage _A_ `depends_on` stage _B_ means _A_ is physically and rigidly mounted on _B_. This is why the chain must end on `.` so the last stage is rigidly mounted to the laboratory.

woutdenolf:
@PeterC-DLS You can use the rendered version t help with the review:

https://hdf5.gitlab-pages.esrf.fr/nexus/nxtransformations_active/classes/base_classes/NXtransformations.html

PeterC-DLS:
We need to add an explanation of `depends_on="."` and also about using any value for a direction axis.

base_classes/NXtransformations.nxdl.xml
```diff
@@ -55,23 +55,42 @@
...
-		links to transformations outside, for example to a common base table.  If
-		a relative path is given, it is relative to the group enclosing the ``depends_on``
-		specification.
+		**Transformation chain**
+
+		The entry point of a chain of transformations is a field called ``depends_on``
+		will be outside of this class and points to a field in here. Following the chain may
+		also require following ``depends_on`` links to transformations outside, for example
+		to a common base table.  If a relative path is given, it is relative to the group
+		enclosing the ``depends_on`` specification.
```
PeterC-DLS, line 68: Add that the end of the chain is denoted by `"."` and means at the origin of the laboratory frame.

base_classes/NXtransformations.nxdl.xml
```diff
@@ -108,6 +126,108 @@
...
+		* *transmission*: :math:`X_\text{lab} = T_x(20) . X_d`
+		* *vertical*: :math:`X_\text{lab} = R_y(-5) . T_x(10) . X_d`
+		* *horizontal*: :math:`X_\text{lab} = R_x(-90) . R_y(-6) . T_x(11) . X_d`
+
+		where :math:`X_d` is a coordinate in the detector reference frame. As these are
+		point detectors we only have one coordinate :math:`X_d=[0,0,0,1]^T`.
+
+		The sample is oriented as follows
+
+		.. math:: X_\text{lab} = R_y(-90) . R_x(30) . R_z(-45) . X_s
```
PeterC-DLS, line 160: This does not match the dependency chain

woutdenolf: Fixed. The order was reversed.

base_classes/NXtransformations.nxdl.xml
woutdenolf: > the end of the chain is denoted by "."

Ok

> means at the origin of the laboratory frame

Not clear. If you take the example of a flat 2D detector with chain `T_3 -> T_2 -> T_1 -> "."`

```
X_lab = T_3 . T_2 . T_1 . X_pixel
```

I don't see how `"."` refers to the "origin".

yayahjb:
These definitions are based on robot arm kinematics.  Please just look at a
kappa goniometer a notice that omega is grounded and depends on nothing
(except a possible translation stage) while phi depends on everything else.

woutdenolf:
Code camp:

- In the example, remove the relative paths with ../ and use an absolute path
- First example must use the standard reference frame, using a kappa goniometer for example (https://doi.org/10.5281/zenodo.3484187)
- Second example uses an alternative reference frame

base_classes/NXtransformations.nxdl.xml
```diff
@@ -108,6 +126,143 @@
...
+		          distance=10            # move downstream from the sample
+		            @depends_on=polar
+		            @units=cm
+		            @vector=[1 0 0]
+		          polar=5                # title above the horizontal plane
+		            @depends_on=azimuth
+		            @units=degrees
+		            @vector=[0 -1 0]
+		          azimuth=0              # stay in the vertical plane
+		            @depends_on=/entry/coordinate_system/beam
```
PeterC-DLS, line 233: This should be `@depends_on="."` as it attached to the laboratory frame's origin.

base_classes/NXtransformations.nxdl.xml
```diff
@@ -108,6 +126,143 @@
...
+		            @units=degrees
+		            @vector=[-1 0 0]
+		      transmission:NXdetector
+		        depends_on=position/distance
+		        position:NXtransformations
+		          distance=20            # move downstream from the sample
+		            @depends_on=/entry/coordinate_system/beam
+		            @units=cm
+		            @vector=[1 0 0]
+		    coordinate_system:NXtransformations
```
PeterC-DLS, line 258: These documents your choice of frame. If you need to know where NXbeam points to at the sample then you must specify it in the appropriate location.

woutdenolf:
@PeterC-DLS Could you write out an actual example on how you would define `coordinate_system`? You mentioned NXbeam a few times, but I fail to see the point. I perhaps should have chosen `gravity` and `magnetic north` or something.

woutdenolf:
Or perhaps `gravity` and `central axis`, as in the central axis of the beamline. The direction of the beam could change in space (mirror, monochromator crystals, ...) so that's probably a bad way of defining the coordinate system.

woutdenolf:
It also seems strange to me that an NXtransformations field without transformation_type cannot be part of a chain. What is it doing under NXtransformations then?

PeterC-DLS:
> @PeterC-DLS Could you write out an actual example on how you would define `coordinate_system`? You mentioned NXbeam a few times, but I fail to see the point. I perhaps should have chosen `gravity` and `magnetic north` or something.

The laboratory coordinate system used is by default McStat, when you want to use another coordinate system for the lab frame, you have to make sure your components are defined consistent to override any default assumptions. The NXbeam is assumed to be along (0,0,1) then if you were to write an NXmx (or any diffraction/scattering) application definition then you must specify the beam direction to be able to work out momentum transfer, q, correctly for any spatially resolved detected event.

Giving reference directions is mainly for documentation. Your application does not need these directions.

sanbrock:
We support this clarification and especially the distinction between active and passive transformations, device motions/displacements and base changes for coordinate system transformations which we also discussed [in Feb telco](https://www.nexusformat.org/Telco_20240214.html). For the later, the documentation (and/or the NXtransofmrations itself) shall be further improved, because the example of describing base changes is not so clear as it only allows defining a single axis at a time, so a transformation for a proper base change of all axes seems to be possible confusingly only in multiple steps.

woutdenolf:
> Giving reference directions is mainly for documentation. Your application does not need these directions.

It doesn't need it (the math would be the same) but as you said, if someone wants to document certain directions they would do it this way no?

`@depends_on=.` means the https://manual.nexusformat.org/design.html#the-nexus-coordinate-system

- +Z is the direction of the beam
- -Y is the direction of gravity

But for example at another institute the convention is

- +X is the direction of the beam
- -Z is the direction of gravity

I was thinking in example 2 with `@depends_on=/entry/coordinate_system/beam` I could define it this way. Is that not correct @PeterC-DLS ?

woutdenolf:
Changes:

1. In the example about coordinate systems
   - in addition to the approach under discussion [in this comment](https://github.com/nexusformat/definitions/pull/1278#issuecomment-5606084292)
   - I also use NXcoordinate_system to define the lab frame in response to [this comment](https://github.com/nexusformat/definitions/pull/1278#issuecomment-2131350090)
2. Use sphinx-gallery for the examples in analogy to [NXdata](https://manual.nexusformat.org/classes/base_classes/NXdata.html). Expand the **Description** to see it.

Rendering: https://hdf5.gitlab-pages.esrf.fr/nexus/nxtransformations_active/classes/base_classes/NXtransformations.html

[image](https://hdf5.gitlab-pages.esrf.fr/nexus/nxtransformations_active/classes/base_classes/NXtransformations.html)

base_classes/NXtransformations.nxdl.xml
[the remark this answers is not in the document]
woutdenolf: Continue https://github.com/nexusformat/definitions/pull/1278#issuecomment-5606084292

base_classes/NXtransformations.nxdl.xml
woutdenolf: Continue in https://github.com/nexusformat/definitions/pull/1278#issuecomment-5606084292

base_classes/NXtransformations.nxdl.xml
woutdenolf: Continue in https://github.com/nexusformat/definitions/pull/1278#issuecomment-5606084292

PR #1284, opened 2023-06-19
Bit depth image

phyy-nx:
New PR superseding #1262 with additional edits from code camp review

From @graeme-winter:

> Fixes https://github.com/nexusformat/definitions/issues/1258 by including it, using effectively the same definition as is currently in the field from DECTRIS

### DIFF
--- a/applications/NXmx.nxdl.xml
+++ b/applications/NXmx.nxdl.xml
@@ -673,7 +673,7 @@
                     </field>
 
                      <field name="bit_depth_image" type="NX_POSINT"
-                        deprecated="use bit_depth_stored">
+                        deprecated="Use bit_depth_stored.">
                         <doc>
                             The number of bits per pixel saved to the image data.
                         </doc>


yayahjb (approved):
Worth doing

applications/NXmx.nxdl.xml
```diff
@@ -666,6 +666,12 @@
                         </doc>
                     </field>
 
+                    <field name="bit_depth_image" type="NX_POSINT" recommended="true">
```
tacaswell, line 669: `image_bit_depth` is clearer and better fits the naming patterns of the other fields.

base_classes/NXdetector.nxdl.xml
```diff
@@ -758,10 +758,15 @@
...
       How many bits the electronics reads per pixel.
       With CCD's and single photon counting detectors,
-      this must not align with traditional integer sizes.
+      this might not align with traditional integer sizes.
       This can be 4, 8, 12, 14, 16, ...
     </doc>
   </field>
+  <field name="bit_depth_image" type="NX_POSINT" recommended="true">
+    <doc>
+      The number of bits per pixel saved to the image data.
```
tacaswell, line 767: The data may be saved as the next higher standard int type (e.g. a 12bit ADC being saved as 16bit uints).

tacaswell:
The name is over-determined by DECTRIS already writing with this name so we have to accept it as is.

I am still confused by why this not redundant with asking the dataset what type it has.

Some of the linked discussions make the good point that there is not a way to note if the data is signed or unsigned.  Might be notionally better to have `image_datatype`, but as noted above we are being jammed by DECTRIS so we should just accept this and move on.

tacaswell (approved):

phyy-nx:
Proposal is instead to try `bit_depth_STORED` in the base class, but this convention requires #1436

benajamin:
I don't like the proposal to to add bit_depth_STORED. It is implementing a feature whose only current use case is to make the name flexible enough to legitimise the "bit_depth_image" that DECTRIS is already writing. We don't need to make it ambiguous with this feature. We should just give the field a simple name (e.g. "bit_depth_stored") and and if that name is not "bit_depth_image", then we should both define and deprecate "bit_depth_image".

phyy-nx:
How about this?  Use `bit_depth_stored` for NXdetector and `bit_depth_image` for NXmx, and just accept a redundant field in NXmx.  Awkward, but it's less awkward than `bit_depth_STORED` for sure.

phyy-nx:
Chatting more with @benajamin, go with the deprecated `bit_depth_image` instead.

tacaswell:
I think we should either bless exactly what dectris/MX is doing in the most minimal way or consider ways to more accurately specify the data types (signed int vs unsigned int vs float (and if you want to open up floats there are non-standard floats (that you can write to hdf5) that use varying fractions of available bits for the exponent and mantissa and rgb (is a color image 8 bits or 24 bits under `bit_depth_image`?) of both what was collected from the hardware and what is written to the disk.

The first is pragmatic but wrong, the second is a lot of work but likely the correct thing to do.

phyy-nx:
Would changing to `NX_NUMBER` suffice for all the edge cases?  Could do that for both `bit_depth_image` and `bit_depth_stored`.

tacaswell (approved):
still approve after recent changes.

Issue #886, opened 2021-03-22
Documentation review


hgoerzig:
Review existing documentation considering understandability for newbies.

gerrit-guenther:
**NXsubentry:** The use of NXsubentry could be fleshed out. Should it only be used when the instrument would apply to different techniques/application definitions or is it intended to be a general approach (e.g. using a NXfluo subentry in a NXfluo entry)? What about the entry/definition attribute which is mandatory - what to write here if there are two subentries for e.g. SAXS/WAXS?

gerrit-guenther:
**NX_ANGLE:** units of angle, example(s): m - what does m mean? https://manual.nexusformat.org/nxdl-types.html#nx-angle

PeterC-DLS:
> **NX_ANGLE:** units of angle, example(s): m - what does m mean? https://manual.nexusformat.org/nxdl-types.html#nx-angle

Looks like a mistake.

gerrit-guenther:
> 
> 
> **NXsubentry:** The use of NXsubentry could be fleshed out. Should it only be used when the instrument would apply to different techniques/application definitions or is it intended to be a general approach (e.g. using a NXfluo subentry in a NXfluo entry)? What about the entry/definition attribute which is mandatory - what to write here if there are two subentries for e.g. SAXS/WAXS?
> 
> Since the sample is defined for each subentry the subentry can apply to different measurements, different setups... Are there suggestions when to use NXentry or NXsubentry?

Heike refered to https://manual.nexusformat.org/rules.html#nxsubentry-or-multi-method-data which gives a clear example how to use NXentry (and I was not aware of) - maybe a link from NXentry and index could be helpful?

gerrit-guenther:
**Nxentry/definition:** "This field is provided so that NXentry can be the overlay position in a NeXus data file for an application definition and its set of groups, fields, and attributes.

It is advised to use NXsubentry, instead, as the overlay position."

That seems to suggest to use NXsubentry in general/excessively, also for non-Multi-Method data - contrary to  https://manual.nexusformat.org/rules.html#nxsubentry-or-multi-method-data ?

PR #1286, opened 2023-06-19
NXsubentry documentation updates

### LINKED ISSUE #1124: **NXsubentry:** The use of NXsubentry could be fleshed out.
**NXsubentry:** The use of NXsubentry could be fleshed out. Should it only be used when the instrument would apply to different techniques/application definitions or is it intended to be a general approach (e.g. using a NXfluo subentry in a NXfluo entry)? What about the entry/definition attribute which is mandatory - what to write here if there are two subentries for e.g. SAXS/WAXS?

Since the sample is defined for each subentry the subentry can apply to different measurements, different setups... Are there suggestions when to use NXentry or NXsubentry?

_Originally posted by @gerrit-guenther in https://github.com/nexusformat/definitions/issues/886#issuecomment-804191753_

phyy-nx:
Closes #1124

### DIFF
--- a/base_classes/NXsubentry.nxdl.xml
+++ b/base_classes/NXsubentry.nxdl.xml
@@ -73,6 +73,8 @@
 		multiple techniques for data collection (such as SAXS/WAXS instruments), 
 		it was recognized the application definitions must be entered in the NeXus
 		data file tree as children of :ref:`NXentry`. 
+
+		See rules for :ref:`NXsubentry or Multi-Method Data &lt;rules-subentry&gt;`
 	</doc>
 	
 	<attribute name="IDF_Version">
--- a/manual/source/examples/hierarchy-subentry.txt
+++ b/manual/source/examples/hierarchy-subentry.txt
@@ -12,17 +12,18 @@
 		SAS:NXsubentry
 			definition = "NXsas"
 			instrument:NXinstrument
-				detector:NXdetector
-					data --> /entry/instrument/SASdet/data
+				detector -> entry/instrument/SASdet
 			data:NXdata
 				data --> /entry/instrument/SASdet/data
+			sample --> /entry/sample
 		Fluo:NXsubentry
 			definition = "NXfluo"
 			instrument:NXinstrument
-				detector --> /entry/instrument/fluordet/data
-				detector2 --> /entry/instrument/large_area/data
+				detector --> /entry/instrument/fluordet
+				detector2 --> /entry/instrument/large_area
 			data:NXdata
 				@signal = detector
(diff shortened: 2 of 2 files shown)


phyy-nx:
@prjemian can you review? I don't know if this helps address all the concerns in #1124.

prjemian:
@gerrit-guenther: Can you comment?

prjemian:
I'm not clear how the one line of documentation added by [commit](https://github.com/nexusformat/definitions/pull/1286/files#diff-c07ce990b08010a7550cd2308e7c069ccbb8a43407bc0f6173ad1daad58ae5ddR77) satisfies the questions raised in #1124 or https://github.com/nexusformat/definitions/issues/886#issuecomment-804191753.

phyy-nx:
Ok, Ive re-read the original issue, and here are the specific questions, broken out and restated:

1. Should NXsubentry only be used when the instrument would apply to different techniques/application definitions?
2. Is NXsubentry intended to be a general approach (e.g. using a NXfluo subentry in a NXfluo entry)?
3. What about the entry/definition attribute which is mandatory - what to write here if there are two subentries for e.g. SAXS/WAXS?
4. Since the sample is defined for each subentry the subentry can apply to different measurements, different setups... Are there suggestions when to use NXentry or NXsubentry?

I've added new documentation to NXsubentry which I believe answers them.
1. The primary goal of NXsubentry is to enable multi-model experiments.  This is now stated directly: "Group to enable the combination of multiple application definitions for “multi-modal” measurements (e.g. multi-technique methods like SAXS/WAXS)."
2. The two supported use cases are either an NXentry with an application definition, or an NXentry without an application definition and multiple sub entries with application definitions.  This is now called out directly using a modified version of @prjemian's [comment](https://github.com/nexusformat/definitions/issues/1124#issuecomment-1165593996).

3 and 4 are hopefully answered by 2.  In addition I've noted that NXsubentries aren't designed to be contained within each other.

Ok, these are at least my interpretations of what is going on.  Feedback appreciated :)

phyy-nx:
From code camp:

I have included this note in the PR: "An NXsubentry should not contain another NXsubentry".  Is that correct?  @yayahjb says we shouldn't need the restriction.  @PeterC-DLS says it doesn't make sense for subentries to be nested.  This would open the possibility of an application specifying a subentry.  (Aaron: this might be useful for workflows?)

I based the assertion on my reading of the docs and @prjemian's comment showing the two use cases (with and without subentries) as meaning those are the only supported, but this might not be what was intended.

@mkoennecke what is the behavior of cnxvalidate if it finds a nested subentry?

Thoughts?

PR #1294, opened 2023-06-21
Fix deployment in CI

### LINKED ISSUE #1288: Fix CI deployment
A previous change that updated the version of `sphinx-notes/pages` action and broke the deployment steps.

Use of this action actually duplicates sphinx builds in previous steps so should be replaced by steps to simply push built files to the correct branch.

PeterC-DLS:
Remove repeated sphinx build and deploy with single action to fix #1288


PeterC-DLS:
Tested in https://github.com/nexusformat/definitions/actions/runs/5332970947

PR #1305, opened 2023-07-18
Fixes automatic versioning

domna:
This fixes a typo which resulted in setuptools-scm not being able to resolve the version (was always 0.0.0).

It also enables `no-guess-dev`, i.e. no new versions are guessed from the current ones, as this makes no sense for date based versioning.


sanbrock (approved):
LGTM

PR #1314, opened 2023-10-11
add field `static_phi_list` to NXxpcs

### LINKED ISSUE #1313: NXxpcs: static_phi_list missing
On review for use at APS, note that field `static_phi_list` is missing.

Also, look for typos (such as `the the`) and some other minor formatting.

Thanks for pointing this out, @azjk, @sureshnaps.  Comment here if you find other items.

prjemian:
- close #1313

@sureshnaps, @AZjk, @ambarb, @AbbyGI: Please comment.

### DIFF
--- a/contributed_definitions/NXxpcs.nxdl.xml
+++ b/contributed_definitions/NXxpcs.nxdl.xml
@@ -59,7 +59,7 @@
       <doc>
         **Locally** unique identifier for the experiment (a.k.a. run or scan).
 
-        * For bluesky users, this is the run's `"scan_id"`.
+        * For bluesky users, this is the run's ``"scan_id"``. If ``scan_id`` is not available, use the run's ``uid``.
         * For SPEC users, this is the scan number (``SCAN_N``).
       </doc>
     </field>
@@ -134,10 +134,10 @@
             the data and its axes are self-describing as per NeXus recommendations. However, the data is preferred in one
             of the following two formats:
 
-            * iterable list of linked files (or keys) for each :math:`g_2` with 1 file (key) per :math:`q`, where `q` is called by the nth roi_map value
-            * 2D array [#]_ with shape (:math:`g_2`, :math:`q`), where `q` is represented by the nth roi_map value, not the value `q` value
+            * iterable list of linked files (or keys) for each :math:`g_2` with one file (key) per :math:`q`, where `q` is called by the nth roi_map value
+            * 2D array [#]_ with shape (:math:`g_2`, :math:`q`), where :math:`q` is represented by the nth roi_map value, not the value :math:`q` value
 
-            Note it is expected that "g2" and all quantities following it will be of the same length.
+            .. note:: It is expected that ``g2`` and all quantities following it will be of the same length.
 
             Other formats are acceptable with sufficient axes description.
 
@@ -152,8 +152,8 @@
 
             We encourage the documentation of other formats not represented here.
 
-            * one array representing entire data set ("one_array")
-            * data exchange format with each key representing one ``q`` by its corresponding roi_map value ("data_exchange_keys")
+            * one array representing entire data set (``"one_array"``)
+            * data exchange format with each key representing one :math:`q` by its corresponding roi_map value (``"data_exchange_keys"``).
           </doc>
           <enumeration>
             <item value="one_array"/>
@@ -194,14 +194,14 @@
       </field>
       <field name="delay_difference" type="NX_INT" units="NX_COUNT" minOccurs="0" maxOccurs="1">
         <doc>
-          delay_difference (also known as delay or lag step)
+          (Also known as delay or lag step.)
 
           This is quantized difference so that the "step" between two consecutive
           frames is one frame (or step ``= dt = 1 frame``)
 
           It is the "quantized" delay time corresponding to the ``g2`` values.
 
-          The unit of delay_differences is ``NX_INT`` for units of frames (i.e., integers) preferred,
+          The unit of delay_differences is ``NX_INT`` (i.e., integers) for units of frames preferred,
           refer to :ref:`NXdetector` for conversion to time units.
         </doc>
         <attribute name="storage_mode" type="NX_CHAR">
@@ -220,7 +220,7 @@
     </doc>
       <field name="two_time_corr_func" type="NX_NUMBER" units="NX_ANY" minOccurs="0" maxOccurs="1">
         <doc>
-          two-time correlation of speckle intensity for a given q-bin or roi (represented by the nth roi_map value)
+          two-time correlation of speckle intensity for a given :math:`q`-bin or roi (represented by the nth roi_map value)
 
           See Fluerasu, Phys Rev E (2007), Eq 1 and Sutton, Optics Express (2003) for an early
           description applied to X-ray scattering:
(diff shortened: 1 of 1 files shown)


AZjk:
static_phi_list is necessary for the result file. thank you for adding that.

contributed_definitions/NXxpcs.nxdl.xml
```diff
@@ -507,7 +507,7 @@
             roi index array or labeled array
 
             The values of this mask index (or map to) the :math:`Q` value from the
-            the ``dynamic_q_list`` field. Not that the value of ``0`` represents in-action. XPCS computations
+            ``dynamic_q_list`` field. Not that the value of ``0`` represents in-action. XPCS computations
```
padraic-shafer, line 510:
```suggestion
            ``dynamic_q_list`` field. Note that the value of ``0`` represents in-action. XPCS computations
```

Here, should "in-action" be "inaction"? Or does it mean something like "in motion"?  Would "uniform mode" or "unmodulated" be more appropriate?

padraic-shafer (commented):
A couple of minor comments here:
* There seems to be an inconsistent use of `q` and `Q`.  Is this meaningful or arbitrary?
* Find-replace for 'xcps' -> 'xpcs'

contributed_definitions/NXxpcs.nxdl.xml
```diff
@@ -523,8 +523,8 @@
             The only requirement for the list is that it may be iterable. Some expected formats are:
 
             * iterable list of floats (i.e., :math:`Q(r)`)
-            * iterable list of tuples (i.e., :math:`Q(r)`, :math:`\varphi`), but preferable use the seperate :math:`\varphi` field below
-            * iterable list of tuples (e.g., (H, K, L); (qx, qy, qz); (horizontal_pixel, vertical_pixel))
+            * iterable list of tuples (i.e., :math:`Q(r)`, :math:`\varphi`), but preferable use the separate :math:`\varphi` field below
```
padraic-shafer, line 526: Which version is the intention here"
* :math:`Q(r)`, :math:`\varphi`
* :math:`Q(r, \varphi)`

contributed_definitions/NXxpcs.nxdl.xml
```diff
@@ -543,15 +543,22 @@
...
+            ``static_q_list`` field.
 
             The ``units`` attribute should be set to ``"au"``
             indicating arbitrary units.
             </doc>
           </field>
           <field name="static_q_list" type="NX_NUMBER" units="NX_PER_LENGTH" minOccurs="0">
             <doc>
-            1-D list of :math:`|Q|` values, 1 for each roi.
+            1-D list of :math:`|Q|` values, one for each roi.
```
padraic-shafer, line 554: Should this `doc` field describe the same options as in its dynamic counterpart?
* Q(r)
* Q(r), phi
* H, K, L
* qx, qy, qz
* x, y pixels

prjemian: Good catch!  You are correct about the inconsistent use of `Q` and `q`.  In small-angle (which is used together sometimes with XPCS), $Q$ has a different meaning than reciprocal space vector; it represents the total scattered intensity, integrated over all $\vec{q}$.  I will suggest a revision as part of this PR.  Should that discussion become distracting to the intent of this PR (as stated in the title), I suggest moving such comment and revision to a separate issue.

prjemian: @sureshnaps, @AZjk, @qzhang234, @ambarb, @AbbyGi: Can you comment if the options and other documentation for `dynamic_q_list` should be copied to `static_q_list`?

padraic-shafer: @qzhang234 Thanks for the gathering input from APS folks and sharing the consensus here!

I think that makes sense to fold these into single lists--one for static and one for dynamic. @ambarb: Is this what you were thinking also?

padraic-shafer: Since many of those columns (`q`, `phi`, `qx`, `qy`, `qz`, `h`, `k`, `l`, etc.) would be empty if we insist that all of them be present...@ambarb had suggested that we could instead add a [@storage_mode](https://manual.nexusformat.org/classes/contributed_definitions/NXxpcs.html#nxxpcs-entry-twotime-two-time-corr-func-storage-mode-attribute) attribute. 

This would be similar to how it is used in the `NXxpcs::entry::twotime` group, and it could have defined values, such as `q_only`, `q_phi`, `qxyz`, `hkl`, `xy`, or whatever makes sense here.  Effectively this would act as a [tagged union](https://en.wikipedia.org/wiki/Tagged_union) discriminator for the q_lists.  I think this sounds useful.

This could be split out into a separate PR, but it seemed relevant to bring it up here.  What do others think?

ambarb: one thing not yet resolved is how the `static_q_list` is envisioned to be used.  I imagine that the list correlates an intensity list so that one may plot I(Q).  @qzhang234  can you confirm?  

The exact wording in our current NXxpcs definition for static and dynamic q-labels is:

**"Dynamic” represents quantities directly related to XPCS and NXxcps/entry/data and NXxpcs/entry/two_time.

“Static” refers to finer binning used for computation not strictly used for the final XPCS results. Implementation of _static_ binning is left for individual libraries to document. We encourage usage of [NXcanSAS](https://manual.nexusformat.org/classes/applications/NXcanSAS.html#nxcansas) to represent standard SAXS results or development of new NeXus definitions for GI-SAXS or other reciprocal space intensity mapping.**

I was wrong, there is a `static_roi_map` so this part is handled, but I see absolutely no data to plot against the x-variables (`static_q_list` or `static_phi_list`).  I believe this is because we left it to others to document as the preference was to use NXcanSAS.  I am fine with what the community wants to do in this regard (I(Q) fully inside custom NXxpcs  versus NXcanSAS), but if the choice is fully inside NXxcps, it seems better to fully document beyond just adding "static_phi_list" and the other discussed changes.

ambarb: https://github.com/nexusformat/definitions/pull/1314#discussion_r1376543362. should have been in this thread, not where it is at the bottom of this page

ambarb: > Since many of those columns (`q`, `phi`, `qx`, `qy`, `qz`, `h`, `k`, `l`, etc.) would be empty if we insist that all of them be present...@ambarb had suggested that we could instead add a [@storage_mode](https://manual.nexusformat.org/classes/contributed_definitions/NXxpcs.html#nxxpcs-entry-twotime-two-time-corr-func-storage-mode-attribute) attribute.
> 
> This would be similar to how it is used in the `NXxpcs::entry::twotime` group, and it could have defined values, such as `q_only`, `q_phi`, `qxyz`, `hkl`, `xy`, or whatever makes sense here. Effectively this would act as a [tagged union](https://en.wikipedia.org/wiki/Tagged_union) discriminator for the q_lists. I think this sounds useful.
> 
> This could be split out into a separate PR, but it seemed relevant to bring it up here. What do others think?

https://github.com/nexusformat/definitions/pull/1314#discussion_r1376543362. should have been in this thread, not where it is at the bottom of this page

contributed_definitions/NXxpcs.nxdl.xml
```diff
@@ -523,8 +523,8 @@
             The only requirement for the list is that it may be iterable. Some expected formats are:
 
             * iterable list of floats (i.e., :math:`Q(r)`)
-            * iterable list of tuples (i.e., :math:`Q(r)`, :math:`\varphi`), but preferable use the seperate :math:`\varphi` field below
-            * iterable list of tuples (e.g., (H, K, L); (qx, qy, qz); (horizontal_pixel, vertical_pixel))
+            * iterable list of tuples (i.e., :math:`Q(r)`, :math:`\varphi`), but preferable use the separate :math:`\varphi` field below
+            * iterable list of tuples (e.g., :math:`(H, K, L)`; :math:`(q_x, q_y, q_z)`; (horizontal_pixel, vertical_pixel))
```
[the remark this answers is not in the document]
ambarb: @padraic-shafer suggestions for attribute names I think are better than what I posted here.

PR #1316, opened 2023-10-31
Adds setup python step in CI/CD

### LINKED ISSUE #1317: The CI/CD does not actually set up the python version
The current CI/CD does not test for each python version as the setup python setup step is missing. Therefore, it's just run multiple times with the default python version in `ubunut-latest`.

domna:
The current CI/CD does not setup the different python version and just sets the `python-version` environment variable. I added a step to set the python version to the one of the current matrix element.

Fixes #1317

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -42,6 +42,11 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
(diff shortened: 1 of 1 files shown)


PeterC-DLS (approved):
LGTM, thanks for the fix.

domna:
> @domna the workflow is to merge your merge once it's been approved and you are happy with it.

I see. The problem is that I don't have the rights to merge in this repo, so I cannot do it myself.

PR #1333, opened 2023-11-21
Rewrite the NXdata scaling_factor and offset fields

### LINKED ISSUE #1332: Rework NXdata scaling_factor and offset
The NXdata fields [scaling_factor](https://manual.nexusformat.org/classes/base_classes/NXdata.html#nxdata-scaling-factor-field) and [offset](https://manual.nexusformat.org/classes/base_classes/NXdata.html#nxdata-offset-field) can both be applied to an NXdata group, but it's unclear in which order.  Rewrite the description of these terms to make it explicit that offset should be applied and then the scaling_factor, like this:

```
corrected_data = (data + offset) * scaling_factor
```

Note this is counter-intuitive to folks like me who want to apply both pedestal and gain corrections.  Here gain is in "counts per photon", so dividing by gain converts from counts to photons.  That equation would like this:

```
corrected_data = (data - pedestal) / gain
```

It would be good to include a note to this effect in the description of these terms, either in the base class itself or in NXmx.

Issue raised through the telco in response to question from @biochem-fan regarding how to supply gain values in NXmx.  Proposal is to use `scaling_factor` as inverse of gain.

phyy-nx:
Closes #1332

Also adds clarification in NXmx that these fields can be used as pedestal and gain correction fields, as well as elaborates on the possible rank options.  These rank options were implied (in my opinion) in the original wording, but in NXmx I made it more explicit.

Tagging @biochem-fan for reference

### DIFF
--- a/applications/NXmx.nxdl.xml
+++ b/applications/NXmx.nxdl.xml
@@ -174,7 +174,7 @@
                 <doc>
                         An optional offset to apply to the values in data.
 
-                        See :ref:`scaling_factor` for more information.
+                        See :ref:`scaling_factor &lt;/NXmx/ENTRY/DATA/scaling_factor-field&gt;` for more information.
                 </doc>
                 </field>
 
--- a/applications/NXmx.nxdl.xml
+++ b/applications/NXmx.nxdl.xml
@@ -143,13 +143,13 @@
                     </dimensions>
                 </field>
 
-                <field name="scaling_factor" type="NX_FLOAT" optional="true">
+                <field name="data_scaling_factor" type="NX_FLOAT" optional="true">
                     <doc>
                         An optional scaling factor to apply to the values in data.
 
                         The elements in data are often stored as integers for efficiency reasons and need
-                        further correction, generating floats.  The two fields scaling_factor and offset
-                        allow linear corrections using the following convention:
+                        further correction, generating floats.  The two fields data_scaling_factor and
+                        data_offset allow linear corrections using the following convention:
 
                         .. code-block::
 
@@ -160,9 +160,9 @@
                         Use these fields to specify gain and/or pedestal constants that need to be applied
                         to the data to correct it to physical values.  For example, if the detector gain
                         is 10 counts per photon and a constant background of 400 needs to be subtracted
-                        off the pixels, specify scaling_factor as 0.1 and offset as -400 to specifiy the
-                        required conversion from raw counts to pedestal-corrected photons. It is implied
-                        processing software will apply these corrections on-the-fly during processing.
+                        off the pixels, specify data_scaling_factor as 0.1 and data_offset as -400 to
+                        specifiy the required conversion from raw counts to pedestal-corrected photons. It
+                        is implied processing software will apply these corrections on-the-fly during processing.
 
                         The rank of these fields should either be a single value for the full dataset, a
                         single per-pixel array applied to every image (rank (i, j) or (i, j, k)),
@@ -170,11 +170,11 @@
                         (np, 1), (np, i, j) or (np, i, j, k)).
                 </doc>
                 </field>
-                <field name="offset" type="NX_FLOAT" optional="true">
+                <field name="data_offset" type="NX_FLOAT" optional="true">
                 <doc>
                         An optional offset to apply to the values in data.
 
-                        See :ref:`scaling_factor &lt;/NXmx/ENTRY/DATA/scaling_factor-field&gt;` for more information.
+                        See :ref:`data_scaling_factor &lt;/NXmx/ENTRY/DATA/data_scaling_factor-field&gt;` for more information.
                 </doc>
                 </field>
 
--- a/applications/NXmx.nxdl.xml
+++ b/applications/NXmx.nxdl.xml
@@ -142,6 +142,43 @@
                         <dim index="4" value="k" required="false"/>
                     </dimensions>
                 </field>
+
+                <field name="scaling_factor" type="NX_FLOAT" optional="true">
+                    <doc>
+                        An optional scaling factor to apply to the values in data.
+
+                        The elements in data are often stored as integers for efficiency reasons and need
+                        further correction, generating floats.  The two fields scaling_factor and offset
+                        allow linear corrections using the following convention:
+
+                        .. code-block::
+
+                                corrected_data = (data + offset) * scaling_factor
+
+                        This formula will derive the actual physical value, when necessary.
+
+                        Use these fields to specify gain and/or pedestal constants that need to be applied
+                        to the data to correct it to physical values.  For example, if the detector gain
+                        is 10 counts per photon and a constant background of 400 needs to be subtracted
+                        off the pixels, specify scaling_factor as 0.1 and offset as -400 to specifiy the
+                        required conversion from raw counts to pedestal-corrected photons. It is implied
+                        processing software will apply these corrections on-the-fly during processing.
+
+                        The rank of these fields should either be a single value for the full dataset, a
+                        single per-pixel array applied to every image (rank (i, j) or (i, j, k)),
+                        a per-image correction specified with an array whose slowest rank is nP (rank
+                        (np, 1), (np, i, j) or (np, i, j, k)).
+                </doc>
+                </field>
+                <field name="offset" type="NX_FLOAT" optional="true">
+                <doc>
+                        An optional offset to apply to the values in data.
+
+                        See :ref:`scaling_factor` for more information.
+                </doc>
+                </field>
+
+
             </group>
 
             <group type="NXsample">
--- a/applications/NXmx.nxdl.xml
+++ b/applications/NXmx.nxdl.xml
@@ -155,7 +155,7 @@
 
                                 corrected_data = (data + offset) * scaling_factor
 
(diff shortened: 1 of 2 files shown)


phyy-nx:
I purport this doesn't need a vote as it doesn't change functionality, and the change to NXmx only clarifies functionality that was already there.  Feedback welcome!

biochem-fan:
@phyy-nx

> The elements in data are usually float values really. For efficiency reasons these are usually stored as integers after scaling with a scale factor. This value is the scale factor. It is required to get the actual physical value, when necessary.

The original description for the gain sounds like `stored_integer_value = physical_value * the_scale_factor`, making the definition of the gain the same as in MX (i.e. value per photon). But your new description is the other way round. Did you check the interpretation at the NIAC?

phyy-nx:
Ack, I think you might be right and I misinterpreted this field!

> The elements in data are usually float values really. For efficiency reasons these are usually stored as integers after scaling with a scale factor. This value is the scale factor. It is required to get the actual physical value, when necessary.

So I think I now agree this is saying the stored value has already had the scaling factor applied.  But that reads differently from offset:

> An optional offset to apply to the values in data.

That implies the offset hasn't been implied yet!

So, @nexusformat/developers, how do you use scaling_factor and offset in NXdata, if at all?

phyy-nx:
Bump! @nexusformat/developers :)

How do you use scaling_factor and offset in NXdata, if at all?

PeterC-DLS:
I don't use these fields but it seems that if you want to document the stored values then `stored_value = (physical_value - offset) * scaling_factor` may be best for the pedestal use but other people want to use `offset` as a bias: `stored_value = physical_value * scaling_factor + offset`. So two more interpretations to choose from!

benajamin:
I have always interpreted these fields as things that need to be done to the `stored value` in order to get to the `physical value`. I would do this as:
`physical value = stored value * scaling_factor + offset`

phyy-nx:
Ok I like switching the discussion to "stored" vs. "plotted".  Based on that, just to recap without math, there are two general ways these can be interpreted:

1. You, the person plotting the data, must apply `scaling_factor` and `offset` before you plot the data
2. I, the data preparer, have already applied `scaling_factor` and `offset` for you, and I'm noting the values I used

I hope the answer is 1. to match @biochem-fan's use case, but it seems to line up with some of the comments above.

I do note that both @benajamin and @woutdenolf switched the order of operations to match the "bias" version that @PeterC-DLS suggested, compared to the pedestal version that I originally suggested:

a. `plotted_value = physical_value * scaling_factor + offset`
b. `plotted_value = (physical_value + offset) * scaling_factor`

a. makes more intuitive sense to me as it more closely matches the equation of a line `y=mx+b`.  But the consequence is that for our MX data that we need to store "gain-corrected" pedestals...  But that's not relevant for @biochem-fan's use case (no pedestal I believe), so maybe a. is fine?

phyy-nx:
I've made the change to use the "plotted" nomenclature for NXdata, but I've kept the pedestal formula for now, until we get a bit more discussion here (`plotted_value = (physical_value + offset) * scaling_factor`).  Mainly because it's I'm coming from an MX background where pedestal is applied first in the use cases I know about.  But I can be convinced otherwise! Especially if folks already have code with the other convention (`plotted_value = physical_value * scaling_factor + offset`)

PeterC-DLS:
LGTM. For clarity's sake in your comments, I think `stored_value` is better than `physical_value` as you may be plotting the "physical" value.

PeterC-DLS (approved):

phyy-nx:
Feedback from Telco addressed.  If there's no more comments we'll send this to a vote.

phyy-nx:
Hello, please vote by providing an emoji on this comment.  Thanks.

prjemian:
Not happy with the term `pedestal` in this context.  Is it the same as `offset`?

phyy-nx:
@prjemian `pedestal`, used only for NXmx, does mean offset, and I assumed it was a known term in x-ray diffraction land.  I could be convinced that gain and pedestal should be defined more clearly, but that could be done in a separate, not-voted on PR, since it would only be a documentation clarification.

prjemian:
image

phyy-nx:
Wait it's not being changed in the PR.  For NXdata it's `FIELDNAME_scaling_factor` and `FIELDNAME_offset`, and for NXmx is `data_scaling_factor` and `data_offset`.

Pedestal is just a documentation term.  Does that help?

applications/NXmx.nxdl.xml
```diff
@@ -142,6 +142,43 @@
...
+                        further correction, generating floats.  The two fields data_scaling_factor and
+                        data_offset allow linear corrections using the following convention:
+
+                        .. code-block::
+
+                                corrected_data = (data + offset) * scaling_factor
+
+                        This formula will derive the corrected value, when necessary.
+
+                        Use these fields to specify gain and/or pedestal constants that need to be applied
```
prjemian, line 160:
```suggestion
Use these fields to specify gain and/or offset constants that need to be applied
```

applications/NXmx.nxdl.xml
```diff
@@ -142,6 +142,43 @@
...
+                        to the data to correct it to physical values.  For example, if the detector gain
+                        is 10 counts per photon and a constant background of 400 needs to be subtracted
+                        off the pixels, specify data_scaling_factor as 0.1 and data_offset as -400 to
+                        specifiy the required conversion from raw counts to pedestal-corrected photons. It
```
prjemian, line 164:
```suggestion
specifiy the required conversion from raw counts to offset-corrected photons. It
```

biochem-fan:
Depending on the field, various terms are used: "offset", "pedestal" and "bias". For example, ThermoFisher's electron detectors use "bias".

Perhaps keeping "offset"  for NXdata is a good idea but I would prefer to have other term mentioned as well in the NXmx documentation for better searchability.

prjemian:
Yes

applications/NXmx.nxdl.xml
phyy-nx: Happy to add more clarification in a different PR after this vote.

applications/NXmx.nxdl.xml
phyy-nx: Happy to add more clarification and fix the typo in a different PR after this vote.

phyy-nx:
Happy to modify reserved suffixes in a different PR after this vote.

base_classes/NXdata.nxdl.xml
```diff
@@ -398,21 +398,43 @@
...
-			when necessary.
+			An optional scaling factor to apply to the values in FIELDNAME (usually the signal).
+
+			The elements in data are often stored as integers for efficiency reasons and need
+			further correction, generating floats.  The two fields FIELDNAME_scaling_factor and
+			FIELDNAME_offset allow linear corrections using the following convention:
+
+			.. code-block::
+
+				plotted_data = (data + offset) * scaling_factor
```
benajamin, line 411: I strongly disagree with the use of "plotted_data" and "data" in this equation. Firstly, "data" is ambiguous and should be something like "stored values", "dataset values", or "recorded values". Secondly "plotted data" makes no sense because nothing is plotted (past tense) at the time when one is considering this equation and it is also meaningless because one could plot any old values. The NeXus manual says that we strive to record physically meaningful values - this equation is to be used when we do not record physically meaningful values and so its purpose should be to convert to physically meaningful values. Therefore, I would argue that "physical values" should definitely be used instead of "plotted values".

base_classes/NXdata.nxdl.xml
```diff
@@ -398,21 +398,43 @@
...
-			An optional offset to apply to the values in data.
+			An optional offset to apply to the values in FIELDNAME (usually the signal).
+
+			See :ref:`FIELDNAME_scaling_factor &lt;/NXdata/FIELDNAME_scaling_factor-field&gt;` for more information.
 		</doc>
 	</field>
 
+	<field name="scaling_factor" type="NX_FLOAT" deprecated="Use FIELDNAME_scaling_factor instead">
+		<doc>
+			Due to scaling_factor being ambiguous in the case of multiple signals, use
```
paulmillar, line 426: A couple of points (both minor)

1. I suggest having some throw-away comment describing the intended semantics (e.g., "Had similar semantics to FIELDNAME_scaling_factor").  This is to allow someone reading the spec to understand how to interpret existing data, where `scaling_factor` has already been used.

    This comment also applies to the `offset` field.

2. Since the ambiguity comes in the case where multiple signals are present (per the proposed doc), I'm assuming the ambiguity doesn't exist if there is a single signal.  For single signal data, is `scaling_factor` still deprecated?  I would imagine so, but perhaps the wording could be made more explicit; for example, by making the statement "use FIELDNAME_scaling_factor instead" a separate sentence, perhaps qualifying it by saying something like "all future data should use ..." . 

    This comment also applies to the `offset` field.

I consider neither comment blocking

paulmillar:
LGTM :smile:

phyy-nx:
Vote did not pass (got 12 votes, needed 13 for quorum), which is fine given all the discussion.  For the sake of clarity I'm closing this PR, removing it and the associated issue from the milestone, and I will make a new PR that addresses the feedback here.  We'll try again then.

PR #1341, opened 2024-01-18
Fix documentation of NXaperture

### LINKED ISSUE #1231: Apparent advice loop: NXaperture and NXslit
I noted that the definition of `NXaperture` says:
> A beamline aperture. This group is deprecated, use NXslit instead

while the definition of `NXslit` says:
> A simple slit. For more complex geometries, `NXaperture` should be used.

This appears (to me) to be a loop, leaving it unclear how to represent "complex" geometries.

phyy-nx:
- NXaperture was marked as deprecated but it is not, so fix the documentation
- Explicitly add an unnamed NXoff_geometry field, the existence of which was implied by the GEOMETRY deprecation

Fixes #1231

### DIFF
--- a/base_classes/NXaperture.nxdl.xml
+++ b/base_classes/NXaperture.nxdl.xml
@@ -27,7 +27,10 @@
     name="NXaperture" 
     type="group" extends="NXobject">
 
-    <doc>A beamline aperture. This group is deprecated, use NXslit instead.</doc>
+    <doc>
+      A beamline aperture.
+
+      Note, the group was incorrectly documented as deprecated, but it is not and it is in common use.</doc>
     <!-- TODO compare with "screens" in SHADOW -->
 
     <field name="depends_on" type="NX_CHAR">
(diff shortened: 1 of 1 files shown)


phyy-nx:
Hi all, reviewing NXaperture more today I noticed that the NXtransformations and (deprecated) NXgeometry fields are unnamed, so I thought it reasonable that we add NXoff_geometry as unnamed as well, instead of using the name SHAPE like we did on the Telco.  Thoughts?  @mkoennecke

Also, it's possible since the deprecated field NXgeometry implies that there should be an NXoff_geometry field, that this PR is only a documentation fix, and therefore doesn't require a vote:

> GEOMETRY: (optional) [NXgeometry](https://manual.nexusformat.org/classes/base_classes/NXgeometry.html#nxgeometry)
> DEPRECATED: Use the field depends_on and [NXtransformations](https://manual.nexusformat.org/classes/base_classes/NXtransformations.html#nxtransformations) to position the aperture and [NXoff_geometry](https://manual.nexusformat.org/classes/base_classes/NXoff_geometry.html#nxoff-geometry) to describe its shape
> 

I'll wait for any feedback on this PR, but unless I hear otherwise, I'll start a vote tomorrow.

phyy-nx:
Hello, please vote by providing an emoji on this comment.  Thanks.

phyy-nx:
Vote has passed.  Needs one approving review before it can be merged @nexusformat/developers.  Thanks.

PeterC-DLS (approved):
LFTM

PR #1343, opened 2024-02-06
Better rewrite of NXdata scaling_factor and offset fields

### LINKED ISSUE #1332: Rework NXdata scaling_factor and offset

phyy-nx:
Adds `FIELDNAME_scaling_factor` and `FIELDNAME_offset` as fields to NXdata.  These fields should be used instead of `scaling_factor` and `offset`, as it is ambiguous which fields to apply them to in the case of multiple signals.  `_scaling_factor` and `_offset` are added as reserved suffixes.

Additionally adds a formula for how to apply these fields:

```
corrected values = (FIELDNAME + offset) * scaling_factor
```

For NXmx, adds clarification these fields can be used as pedestal and gain correction fields, and defines these terms.  Also the possible rank options are elaborated. These rank options were implied (in my opinion) in the original wording, but in NXmx I made it more explicit.

Closes #1332.  Same set of changes as in #1333 with an additional commit to resolve feedback from #1333.


phyy-nx:
[691f56a](https://github.com/nexusformat/definitions/pull/1343/commits/691f56a20824265d2f098c268216e89b3723397c) addresses comments from the review of #1333:

- NXmx: fully define pedestal, bias, and gain and how they relate to offset and scaling_factor.  Comments from @prjemian
- Add reserved suffixes _scaling_factor and _offset. Comment from @PeterC-DLS 
- NXdata: define better how the stored values in FIELDNAME are converted to physical values. Comment from @benajamin 
- NXdata: more clarification on how scaling_factor and offset without FIELDNAME are ambiguous. Comment from @paulmillar

Thanks for the feedback!  This is ready for further review :)

base_classes/NXdata.nxdl.xml
```diff
@@ -446,21 +446,47 @@
...
-			An optional offset to apply to the values in data.
+			An optional offset to apply to the values in FIELDNAME (usually the signal).
+
+			See :ref:`FIELDNAME_scaling_factor &lt;/NXdata/FIELDNAME_scaling_factor-field&gt;` for more information.
 		</doc>
 	</field>
 
+	<field name="scaling_factor" type="NX_FLOAT" deprecated="Use FIELDNAME_scaling_factor instead">
+		<doc>
+			scaling_factor is ambiguous in the case of multiple signals, so use
```
paulmillar, line 476: I think this sentence is too busy: it's trying to covey different pieces of information, which makes it somewhat hard to parse.  I tend to favour shorter sentences with more direct meaning; e.g.,

    The scaling_factor and FIELDNAME_scaling_factor fields have similar semantics.
    However, scaling_factor is ambiguous in the case of multiple signals.  Therefore
    scaling_factor is deprecated.  Use FIELDNAME_scaling_factor instead, even when
    only a single signal is present.

A similar phrasing could be used for `offset` / `FIELDNAME_offset`.

phyy-nx: Hi @paulmillar, I wasn't intending to explicitly deprecate sclaing_factor and offset in favor of FIELDNAME_scaling_factor and FIELDNAME_offset.  Is that something we should consider?  Comments appreciated.

paulmillar: Hi @phyy-nx ,
I don't really have a strong opinion on deprecating `scaling_factor` and `offset` -- I'm mildly in favour of it, as I think it could lead to simpler analysis/viewing software; at least, in the long run.

However, I'm a bit confused by your statement:
> I wasn't intending to explicitly deprecate sclaing_factor and offset in favor of FIELDNAME_scaling_factor and FIELDNAME_offset.

I understood the proposed change does exactly this, by updating the field definition to state:
> <field [..] deprecated="Use FIELDNAME_scaling_factor instead">

phyy-nx: You're right, I forgot!  Applied your suggestions.

applications/NXmx.nxdl.xml
```diff
@@ -142,6 +142,46 @@
                         <dim index="4" value="k" required="false"/>
                     </dimensions>
                 </field>
+
+                <field name="data_scaling_factor" type="NX_FLOAT" optional="true">
+                    <doc>
+                        An optional scaling factor to apply to the values in ``data``.
+
+                        The elements in ``data`` are often stored as integers for efficiency reasons and need
```
PeterC-DLS, line 150: Maybe a datatype narrower than floats

applications/NXmx.nxdl.xml
```diff
@@ -142,6 +142,46 @@
...
+                        .. code-block::
+
+                                corrected_data = (data + offset) * scaling_factor
+
+                        This formula will derive the corrected value, when necessary.
+
+                        data_scaling_factor is sometimes known as gain or bias and data_offset is sometimes
+                        known as pedestal or background, depending on the community.
+
+                        Use these fields to specify constants that need to be applied
```
PeterC-DLS, line 163: Does the data field's `units` attribute need to be specified in the `scaling_factor` or in `data`?

applications/NXmx.nxdl.xml
```diff
@@ -142,6 +142,46 @@
...
+                        to the data to correct it to physical values.  For example, if the detector gain
+                        is 10 counts per photon and a constant background of 400 needs to be subtracted
+                        off the pixels, specify data_scaling_factor as 0.1 and data_offset as -400 to
+                        specify the required conversion from raw counts to corrected photons. It
+                        is implied processing software will apply these corrections on-the-fly during processing.
+
+                        The rank of these fields should either be a single value for the full dataset, a
+                        single per-pixel array applied to every image (rank (i, j) or (i, j, k)),
+                        a per-image correction specified with an array whose slowest rank is nP (rank
```
PeterC-DLS, line 172: "or a per-image ...". Same here with rank vs dimensions vs shape.

applications/NXmx.nxdl.xml
phyy-nx: Not sure what you mean here.  data_scaling_factor and data_offset are both NX_FLOAT.  What data type is narrower than float?

applications/NXmx.nxdl.xml
```diff
@@ -142,6 +142,46 @@
...
+                        further correction, generating floats.  The two fields data_scaling_factor and
+                        data_offset allow linear corrections using the following convention:
+
+                        .. code-block::
+
+                                corrected_data = (data + offset) * scaling_factor
+
+                        This formula will derive the corrected value, when necessary.
+
+                        data_scaling_factor is sometimes known as gain or bias and data_offset is sometimes
```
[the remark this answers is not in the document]
phyy-nx: Removed bias to avoid confusion

applications/NXmx.nxdl.xml
phyy-nx: I don't think so.  We don't specify a units attribute for data, so even though we could be converting from ADU to photons here, that wouldn't necessarily be a universal application for scaling_factor.

applications/NXmx.nxdl.xml
```diff
@@ -142,6 +142,46 @@
...
+
+                        Use these fields to specify constants that need to be applied
+                        to the data to correct it to physical values.  For example, if the detector gain
+                        is 10 counts per photon and a constant background of 400 needs to be subtracted
+                        off the pixels, specify data_scaling_factor as 0.1 and data_offset as -400 to
+                        specify the required conversion from raw counts to corrected photons. It
+                        is implied processing software will apply these corrections on-the-fly during processing.
+
+                        The rank of these fields should either be a single value for the full dataset, a
+                        single per-pixel array applied to every image (rank (i, j) or (i, j, k)),
```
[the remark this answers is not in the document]
phyy-nx: Agree that rank was confusing here, so switched to dimensions, as is used elsewhere (see eg incident_wavelength).

applications/NXmx.nxdl.xml
phyy-nx: Fixed, with the same swap to dimensions

phyy-nx:
@paulmillar @PeterC-DLS I've come back to this after a while and responded to your comments.  Can you review and resolve the comments that are ok now?

Also, I'm not seeing why the checks are failing.  Any help on this @PeterC-DLS?

Thanks!!

phyy-nx:
Feedback from Telco: need to deprecate NXdata's scaling_factor and offset instead of replacing them with FIELDNAME_scaling_factor and FIELDNAME_offset

phyy-nx:
Woops I actually had deprecated the original fields.  So @paulmillar and @PeterC-DLS this is ready to review.  Thanks!

applications/NXmx.nxdl.xml
PeterC-DLS: I think I meant after correction the result could just be integer if scale and offset were also integers

base_classes/NXdata.nxdl.xml
```diff
@@ -446,21 +446,49 @@
 	</field>
 
 	<!-- Data vs. plot coordinates -->
-	<field name="scaling_factor" type="NX_FLOAT">
+	<field name="FIELDNAME_scaling_factor" type="NX_FLOAT">
```
PeterC-DLS, line 449: Should this be DATA to align with changes in #1396

applications/NXmx.nxdl.xml
```diff
@@ -142,6 +142,46 @@
...
+                    <doc>
+                        An optional scaling factor to apply to the values in ``data``.
+
+                        The elements in ``data`` are often stored as integers for efficiency reasons and need
+                        further correction, generating floats.  The two fields data_scaling_factor and
+                        data_offset allow linear corrections using the following convention:
+
+                        .. code-block::
+
+                                corrected_data = (data + offset) * scaling_factor
```
paulmillar, line 156: [NON-BLOCKING]

Just as an observation, what (I believe) is being described here is a linear model for mapping between a continuous/observed values and the discrete/digitised signal.

A linear mapping is not the only approach.  Other approaches have been tried (in other fields) when it is desirable to focus more of the fidelity of digitised dynamic range in particular areas of interest; for real-world examples, see the [A-law](https://en.wikipedia.org/wiki/A-law_algorithm) and [μ-law](https://en.wikipedia.org/wiki/%CE%9C-law_algorithm) algorithms.

To be very clear: this comment is NON-BLOCKING.  It is *not* intended to delay or prevent the proposed change, which I think is reasonable, but rather to raise a question whether NeXus might need to support non-linear mappings in the future.

(My apologies if this is all "very obvious".  I'm not a domain expert, so I don't know to what extent is just part of "what everyone knows" :-) )

paulmillar:
Hi @phyy-nx 

> Woops I actually had deprecated the original fields. So @paulmillar and @PeterC-DLS this is ready to review. Thanks!

My earlier comment has been addressed.

I've added a NON-BLOCKING comment, meant as a general observation rather than a comment about this particular pull request.

The pull request looks reasonable to me.

prjemian:
@paulmillar Once this PR is resolved, visibility of your non-blocking comment will drop.  I suggest elevating that visibility by making your comment a new issue (use the GitHub feature).
image

paulmillar:
Hi @prjemian 

> @paulmillar Once this PR is resolved, visibility of your non-blocking comment will drop. I suggest elevating that visibility by making your comment a new issue

Thanks. I've followed your suggestion and created #1400.

base_classes/NXdata.nxdl.xml
phyy-nx: I brought in the language from #1396.  FIELDNAME to me makes the most sense as one could apply scaling factor and offset to other parts of the file.  Or am I reading too broadly here?

applications/NXmx.nxdl.xml
phyy-nx: Fixed, upgraded to NX_NUMBER to allow ints or floats.  This is the same type as the DATA field anyways.

phyy-nx:
I think I've addressed all the comments @paulmillar @PeterC-DLS.  If folks agree, can you resolve your comments and approve it?  Approving here doesn't mean merging as it likely needs a vote.  Just that it's ready for a vote.

manual/source/datarules.rst
```diff
@@ -242,10 +244,13 @@ suffix              reference                                  meaning
 ``_mask``           ..                                         Field containing a signal mask, where 0 means the pixel is not masked. If required, bit masks are defined in :ref:`NXdetector` ``pixel_mask``.
 ``_set``            :ref:`target values <target_value>`        Target value of ``DATASET``
 ``_weights``        ..                                         divide ``DATASET`` by these weights [#]_
+``_scaling_factor`` :ref:`NXdata`                              Multiply ``DATASET`` by this factor [#]_
```
PeterC-DLS, line 247: Need to extend width of first column to fix build

manual/source/datarules.rst
```diff
@@ -231,21 +231,26 @@ the following table lists the suffixes reserved by NeXus.
...
+``_end``             :ref:`NXtransformations`                   end points of the motions that start with ``DATASET``
+``_errors``          :ref:`NXdata`                              uncertainties (a.k.a., errors)
+``_increment_set``   :ref:`NXtransformations`                   intended average range through which the corresponding axis moves during the exposure of a frame
+``_indices``         :ref:`NXdata`                              Integer array that defines the indices of the signal field which need to be used in the ``DATASET`` in order to reference the corresponding axis value
+``_mask``            ..                                         Field containing a signal mask, where 0 means the pixel is not masked. If required, bit masks are defined in :ref:`NXdetector` ``pixel_mask``.
+``_set``             :ref:`target values <target_value>`        Target value of ``DATASET``
+``_weights``         ..                                         divide ``DATASET`` by these weights [#]_
+``_scaling_factor``  :ref:`NXdata`                              Multiply ``DATASET`` by this factor [#]_
```
PeterC-DLS, line 247: This is referring to new footnotes

PeterC-DLS (approved):
LGTM

phyy-nx:
This PR is ready for a NIAC vote. Please vote with an emoji on this comment. Options include thumbs up for yes, thumbs down for no, and anything else to abstain. Thank you. Voting will close in two weeks.

applications/NXmx.nxdl.xml
```diff
@@ -171,12 +171,16 @@
                         single per-pixel array applied to every image (dimensions (i, j) or (i, j, k)),
                         or a per-image correction specified with an array whose slowest rank is nP (dimensions
                         (np, 1), (np, i, j) or (np, i, j, k)).
+
+                        When ommited, the scaling factor is assumed to be 1.
```
PeterC-DLS, line 175: Mispelt "omitted"

phyy-nx: Fixed, thanks

PR #1359, opened 2024-02-23
Issue 1148: add general NXenvironment below NXsample

### LINKED ISSUE #1148: add more NXenvironment groups into NXsample
Hi,
at DESY (with SECoP@HMC) we are trying to improve our description of  sample environment.
However, there are only `magnetic_field_env` and `temperature_env` of ` NXenvironment` in `NXsample` only. 
On the other hand  `NXsensor.measurement` defines much more possible physical quantities.
Therefore, I think it would be good to add more NXenvironment groups into NXsample which correspond to NXsensor.measurement possible values. 
Otherwise there is not place to put NXsensor group  related to a sample  which has a different measurement than `temperature` and `magnetic_field`.

jkotan:
It resolves #1148 by adding a general NXenvironment group below the NXsample group

### DIFF
--- a/base_classes/NXsample.nxdl.xml
+++ b/base_classes/NXsample.nxdl.xml
@@ -378,6 +378,11 @@
 			Note, it is recommended to use NXtransformations instead.
 		</doc>
 	</field>
+        <group type="NXenvironment">
+         <doc>
+           This group collects NXsensor groups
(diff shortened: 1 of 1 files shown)


phyy-nx:
Resolved by #1413

PR #1372, opened 2024-03-09
removed unused namespace prefixes xs and ns

### LINKED ISSUE #1371: NXdetector and NXdetector_channel define redundant namespace prefixes
The definition `xmlns:xs` and `xmlns:ns` attributes in NXdetector and NXdetector_channel are not used and therefore redundant. As I'm not that familiar with XML schema's, here is why I think so:

# Schema's

NXDL schema attributes (https://www.w3schools.com/xml/schema_schema.asp)

```xml
# nxdl.xsd

<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
	targetNamespace="http://definition.nexusformat.org/nxdl/3.1"
	xmlns:nx="http://definition.nexusformat.org/nxdl/3.1"
	...>
    ...
</xs:schema>
```

* `xmlns:xs`: elements and data types with prefix **xs** that are **used in this schema** come from this namespace
* `targetNamespace`: elements **defined by this schema** come from this namespace.
* `xmlns:nx`: elements and data types with prefix **nx** that are **used in this schema** come from this namespace

```xml
# nxdlTypes.xsd
<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
	targetNamespace="http://definition.nexusformat.org/nxdl/3.1"
	xmlns:nxdl="http://definition.nexusformat.org/nxdl/3.1"
	...>
    ...
</xs:schema>
```

* `xmlns:nxdl`: elements and data types with prefix **nxdl** that are **used in this schema** come from this namespace

To harmonize we should probably choose one prefix for the NXDL namespace: `xmlns:nx` or `xmlns:nxdl`.

# Definitions

Most definitions have these attributes

```xml
# base_classes/NXentry.nxdl.xml

<definition ...
	xmlns="http://definition.nexusformat.org/nxdl/3.1"
	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
	xsi:schemaLocation="http://definition.nexusformat.org/nxdl/3.1 ../nxdl.xsd"
	>
    ...
</definition>
```

* `xmlns`: elements and data types without prefix that are **used in this schema** come from this namespace (i.e. the default namespace)
* `xmlns:xsi`: elements and data types with prefix **xs** that are **used in this schema** come from this namespace
* `xsi:schemaLocation`: namespace and the location of the XML schema to use for that namespace

However `NXdetector` and `NXdetector_channel` have these attributes

```xml
# base_classes/NXdetector.nxdl.xml

<definition ...
            xsi:schemaLocation="http://definition.nexusformat.org/nxdl/3.1 ../nxdl.xsd"
            xmlns="http://definition.nexusformat.org/nxdl/3.1"
            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
            xmlns:xs="http://www.w3.org/2001/XMLSchema"
            xmlns:ns="http://definition.nexusformat.org/nxdl/@NXDL_RELEASE@"
            >
```

* `xmlns:xs`: elements and data types with prefix **xs* that are **used in this schema** come from this namespace
* `xmlns:ns`: elements and data types with prefix **ns** that are **used in this schema** come from this namespace

There is no harm in doing so but the prefix **xs** and **ns** are not used so these are redundant.

woutdenolf:
Closes #1371

### DIFF
--- a/base_classes/NXdetector.nxdl.xml
+++ b/base_classes/NXdetector.nxdl.xml
@@ -21,14 +21,15 @@
 #
 # For further information, see http://www.nexusformat.org
 -->
-<definition category="base" extends="NXobject" name="NXdetector"
-            type="group"
-            xsi:schemaLocation="http://definition.nexusformat.org/nxdl/3.1 ../nxdl.xsd"
-            xmlns="http://definition.nexusformat.org/nxdl/3.1"
-            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-            xmlns:xs="http://www.w3.org/2001/XMLSchema"
-            xmlns:ns="http://definition.nexusformat.org/nxdl/@NXDL_RELEASE@"
-            >
+<definition
+  name="NXdetector"
+  type="group"
+  extends="NXobject"
+  category="base"
+  xmlns="http://definition.nexusformat.org/nxdl/3.1"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://definition.nexusformat.org/nxdl/3.1 ../nxdl.xsd"
+  >
 
   <symbols>
     <doc>
--- a/base_classes/NXdetector_channel.nxdl.xml
+++ b/base_classes/NXdetector_channel.nxdl.xml
@@ -21,14 +21,15 @@
 #
 # For further information, see http://www.nexusformat.org
 -->
-<definition category="base" extends="NXobject" name="NXdetector_channel"
-            type="group"
-            xsi:schemaLocation="http://definition.nexusformat.org/nxdl/3.1 ../nxdl.xsd"
-            xmlns="http://definition.nexusformat.org/nxdl/3.1"
-            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-            xmlns:xs="http://www.w3.org/2001/XMLSchema"
-            xmlns:ns="http://definition.nexusformat.org/nxdl/@NXDL_RELEASE@"
-            >
+<definition
+  name="NXdetector_channel"
+  type="group"
+  extends="NXobject" 
+  category="base"
+  xmlns="http://definition.nexusformat.org/nxdl/3.1"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://definition.nexusformat.org/nxdl/3.1 ../nxdl.xsd"
+  >
 
   <symbols>
     <doc>


phyy-nx:
Feel like I should chime in since it touches NXdetector_channel, but dunno enough about namespaces to help :)

woutdenolf:
> Feel like I should chime in since it touches NXdetector_channel, but dunno enough about namespaces to help :)

I'm not a specialist either, but this is what I understand: whenever you use an element from a schema (like NXDL or XSD) in an XML file, you need to specify which XML schema defines that element (i.e. which namespace). This is done by prefixing

```xml
<myprefix:parent_element xmlns:myprefix="http://some/url">

    <myprefix:some_schema_element ...>
       ...
    </myprefix:some_schema_element>

</myprefix:parent_element>
```

The word "myprefix" only matters in this file. You could use another word to refer to the same namepace in other files.

I case you use elements from more than one schema

```xml
<myprefix1:parent_element xmlns:myprefix1="http://some/url" xmlns:myprefix2="http://some/url">

    <myprefix1:some_schema1_element ...>
       ...
    </myprefix1:some_schema1_element>

    <myprefix2:some_schema2_element ...>
       ...
    </myprefix2:some_schema2_element>

</myprefix1:parent_element>
```

You can also specify a default namespace which allows you to omit prefixes

```xml
<parent_element xmlns="http://some/url">

    <some_schema_element ...>
       ...
    </some_schema_element>

</parent_element>
```

In all our `.nxdl.xml` files we obviously use NXDL as the default namespace

```xml
<definition
    xmlns="http://definition.nexusformat.org/nxdl/3.1"
>
   <doc ...>
      ...
   </doc>
   <field ...>
      ...
   </field>
</definition>
```

We could have done this but that would be unnecessary verbose

```xml
<nxdl:definition
    xmlns:nxdl="http://definition.nexusformat.org/nxdl/3.1"
>
   <nxdl:doc ...>
      ...
   </nxdl:doc>
   <nxdl:field ...>
      ...
   </nxdl:field>
</nxdl:definition>
```

Now in `NXdetector.nxdl.xml` and `NXdetector_channel.nxdl.xml` we define 

```xml
<definition
    xmlns="http://definition.nexusformat.org/nxdl/3.1"
    xmlns:xs="http://www.w3.org/2001/XMLSchema"
    xmlns:ns="http://definition.nexusformat.org/nxdl/@NXDL_RELEASE@"
>
</definition>
```

So we use NXDL as the default namespace but that we define two other prefixes: `xs` refers the XSD (which is the schema used the define the NXDL schema itself) and `ns` which refers to a particular NXDL definition. Neither of these prefixes out used in `NXdetector.nxdl.xml` and `NXdetector_channel.nxdl.xml`. Moreover it adds confusion to people like me how are not very familiar with XML schema's. So I propose to remove them.

PR #1373, opened 2024-03-09
doc elements can have dim elements but this was not the intention

woutdenolf:


benajamin:
Are you saying that this is just a typo/bug that this pull fixes?

woutdenolf:
Yes, currently it is this:

```xml
<doc>
    <dim ... />
</doc>
```

This is invalid syntax. 

When reading the doc, it seems clear that the idea was to actually document the NXDL `dim ... />` syntax rather than have an actual XML element inside the doc element. We can achieve this by replacing `<` with ` &lt;` and `>` with ` &gt;`

benajamin (approved):

PR #1375, opened 2024-03-21
Add first NXraman Version

RonHildebrandt:
I want to implement the first NeXus definition for NXraman.

As the latest version of NXellipsometry is derived from NXopt, the NXraman here is as well derived from NXopt.
Therefore, the Version here used "NXbeam_path", which will be replaced in the future by using "NXbeam_device" via "previous_devices". This was motivated by the fact, that there is right now no "NXlens_opt(NXbeam_device)" or "NXdetector(NXbeam_device)".

**What is necessary?**
From a physical point of view, these are the most important things for a Raman scattering experiment:
1. The incident wavelength (a quite trivial parameter and the spectral shape of a Raman spectrum can be extremely sensitive to this)
2. The scattering configuration, i.e. what is the incident and scattered light direction and polarization. 

Beyond this, there are many nice to have parameters, but from my point of view not necessary to be able to use this spectrum.

**Questions:**
A) No symbols were used here, as these were not used in the original NXellipsometry(NXopt) definition. It this fine that way, or at which pointare these symbols used?
B) Is the "scattering_configuration" implemented fine this way? Usually, the "Porto notation" is enough to describe the experiment and beatiful simple. But, it can be the case, that you want to give the specific vectors instead of x,y,z. I just wanted to do this as List of 4 Vectors with dimension 3. Is this okay?
C) Depending on the NXbeam_path or NXbeam_device approach, I may have to completely redo this PR. What is the strategy?

**TL;DR:** Created NXraman.yaml file similar to NXellipsometry.yaml


RonHildebrandt:
Sorry for this mistake - wrong base was selected in creation of the draft

Issue #1381, opened 2024-04-24
Change to the axes attribute meaning


jacobfilik:
Hello,

It has recently been raised at DLS that the documentation for NXdata now explicitly says that the axes attribute does not have to be the same length as the rank of the signal dataset.

I do not understand the reasoning for this change, the check that the axes attribute length matches the rank of the signal dataset has been a validation check we have used since the 2014 changes to the NXdata.

I have seen there is lots of discussion around this change in various issue and PRs, but I do not believe it is a sensible choice, since the length of the axes attribute matching the rank of the signal is a simple check (and if it wasn't necessary why have the "." marker at all).

In the docs, there are three examples where the axes attribute has been commented with ""the order does NOT matter" - the latter two showing the issue raised.

The first example looks reasonable, an x and y value for all values in the grid - likely to be the readback values from the stages in a grid scan (although if this was the case it would probably have been preferable to include the _set  demand axes as well and tag them as axes for each dimension - if it is not a "meshed" grid scan I would wonder why all the datasets were not vectors).

The second example I would argue you should still have [x, . , y] as the axes attribute, so the _indices are consistent with that attribute - then at least software that doesn't understand the _indices would still correctly parse the NXdata.

The third example I don't know why the axes attribute is not ["x", "y", "energy"] or ["x", "y", "wavelength"] or ["x", "y", "energy", "."] - with the writer of the file deciding the priority (or which default axis to show - either wavelength or energy or just index). Again, any software that expects axes attribute length to match the signal rank would parse this, and anything that understands _indices should give the option to let the user change the axis for the final dimension.

With this change the axes attribute just becomes a list of axes names (which I can get by parsing the NXdata for anything ending in _indices). The first thing I have to do is now check axes length == signal rank and if it doesnt I completely ignore the axes attribute because it tells me nothing.

Can you give an explicit use case for this change of behaviour, or explain why the above interpretation is wrong?

jacobfilik:
Also, the statement:

 "The names of all [AXISNAME] fields are listed in the [axes] attribute."

Is I believe a big change, and has made all the files where we write read back and target/set/demand axes like (for a line scan in x):

```
data:NXdata
  @signal = "data"
  @axes = ["x_set"]
  @x_indices = [0]
  @x_set_indices = [0]
  data: float[10]
  x_set: float[10]
  x: float[10]
 
```
Are now not valid NeXus.

jacobfilik:
Also, from the 2014 spec here:

axes:

[...]

PR #1392, opened 2024-06-26
Retighten specification of NXdata's axes attribute for #1381

PeterC-DLS:
Tweak examples and clarify that alternative choices for axis values can be indicated by the presence of its corresponding AXISNAME_indices attribute

### DIFF
--- a/base_classes/NXdata.nxdl.xml
+++ b/base_classes/NXdata.nxdl.xml
@@ -86,7 +86,7 @@
 		.. index:: coordinates
 
 		The :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` fields contain the axis coordinates associated with the data values.
-		The names of all :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` fields are listed in the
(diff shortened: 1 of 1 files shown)


PeterC-DLS:
@jacobfilik, please review too.

woutdenolf (commented):
As I mentioned in https://github.com/nexusformat/definitions/issues/1381 I do agree we approved https://github.com/nexusformat/definitions/pull/1213 without voting while it does change the standard. So we should put it back.

But I don't think this tweak is enough, that's how we ended up with an inconsistent and impossible to read NXdata doc in the first place.

I'll try to make an alternative PR that explicitly states that the  length of `@axes` must be equal to rank of the signals. And then discuss how you need to find all axes of NXdata and change the examples to show what to do with the inconsistent definition as a result of the constraint.

woutdenolf:
> I'll try to make an alternative PR that explicitly states that the length of `@axes` must be equal to rank of `signal`. And then discuss how you need to find all axes of NXdata and change the examples to show what to do with the inconsistent definition as a result of the constraint.

https://github.com/nexusformat/definitions/pull/1396

rayosborn (approved):

PeterC-DLS:
Abandoned in favour of #1396

PR #1395, opened 2024-06-28
NXazint contribution

woutdenolf:
Organize this contribution: https://github.com/nexuscontributions/nxazint/issues/3

**[HTML rendering of this contribution](https://hdf5.gitlab-pages.esrf.fr/nexus/nxazint/classes/contributed_definitions/NXazint.html)**

MAX IV example: https://h5web.panosc.eu/h5grove?file=azint/lambda_integrated_maxiv.h5

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,4 +1,4 @@
-name: CI
+name: Documentation build
 
 on:
   push:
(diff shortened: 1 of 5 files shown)


zdemat (commented):
There is a suggestion from Zdenek and Mads (MAX IV) to replace `long_name` **momentum transfer magnitude** with **scattering vector magnitude**. Both of them are possible but azint processing is mainly used by small angle scattering and powder diffraction communities and they may prefer the latter naming convention.

PR #1396, opened 2024-06-30
Fix: 1381 change to the axes attribute meaning

### LINKED ISSUE #1381: Change to the axes attribute meaning

woutdenolf:
Closes #1381 

The NXdata definition got a makeover in PR #1213 to make it more understandable. In this effort, the `@axes` attribute was made less restrictive to allow for an unambiguous discovery of axis fields. This does not make existing files invalid but it does change thing for existing readers as pointed out in issue #1381 by @jacobfilik .

The sole purpose of this PR is to rectify PR #1213 and NOT introduce anything new. The alternative PR https://github.com/nexusformat/definitions/pull/1392 by @PeterC-DLS fixes the situation by carefully modifying some sentences here and there. I would argue however that the entire "axes" section  in the introduction has been structured with the less restrictive `@axes` in mind. In this PR I refactor the entire "axes" section as well.

FYI the less restrictive `@axes` attribute definition

```
The `@axes` attribute provides the names of all the `AXISNAME` fields in the NXdata group. The corresponding `@AXISNAME_indices` attributes provide the signal dimensions they span and when missing the position(s) of `AXISNAME` in `@axes` are taken as the signal dimensions spanned by `AXISNAME`. The shape of an `AXISNAME` field is required to be equal to the shape of the spanned signal dimensions.
```

This definition is much simpler and more concise than the definition in the "axes" section in the PR. However it removes the restriction that `length(axes) == rank(signal)` so we need to put it back in.

When reviewing this PR, please keep in mind that the purpose is to rectify NXdata, not improve it. Use #1381 for that.

### DIFF
--- a/base_classes/NXdata.nxdl.xml
+++ b/base_classes/NXdata.nxdl.xml
@@ -50,8 +50,8 @@
 	
 	<doc>
 		The :ref:`NXdata` class is designed to encapsulate all the information required for a set of data to be plotted.
-		NXdata groups contain plottable data (sometimes referred to as *signals* or *dependent variables*) and their
-		associated axis coordinates (sometimes referred to as *axes* or *independent variables*).
+		NXdata groups contain plottable data (also referred to as *signals* or *dependent variables*) and their
+		associated axis coordinates (also referred to as *axes* or *independent variables*).
 
 		The actual names of the :ref:`DATA &lt;/NXdata/DATA-field&gt;` and :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` fields
 		can be chosen :ref:`freely &lt;validItemName&gt;`, as indicated by the upper case (this is a common convention in all NeXus classes).
--- a/base_classes/NXdata.nxdl.xml
+++ b/base_classes/NXdata.nxdl.xml
@@ -85,124 +85,170 @@
 		.. index:: axes (attribute)
 		.. index:: coordinates
 
-		The :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` fields contain the axis coordinates associated with the data values.
-		The names of all :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` fields are listed in the
-		:ref:`axes &lt;/NXdata@axes-attribute&gt;` attribute.
+		The :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` fields contain the axis coordinates associated with the signal values.
 
-		`Rank`
+		`Default axes`
 
-		:ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` fields are typically one-dimensional arrays, which annotate one of the dimensions.
+		The :ref:`axes &lt;/NXdata@axes-attribute&gt;` attribute provides the names of the :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;`
+		fields to be used as the default axis for each dimension of the :ref:`data &lt;/NXdata/DATA-field&gt;` fields.
+		As a result the length of the :ref:`axes &lt;/NXdata@axes-attribute&gt;` attribute must be equal to the rank of the :ref:`data &lt;/NXdata/DATA-field&gt;`
+		fields. When a particular dimension has no default axis, the string “.” is used in that position.
 
-		An example of this would be
+		`Spanned dimensions`
+
+		The :ref:`AXISNAME_indices &lt;/NXdata@AXISNAME_indices-attribute&gt;` attributes describe the
+		:ref:`data &lt;/NXdata/DATA-field&gt;` dimensions spanned by the corresponding :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` fields.
+
+		When :ref:`AXISNAME_indices &lt;/NXdata@AXISNAME_indices-attribute&gt;` is missing for a given :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` field,
+		the position(s) of the :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` field name in the :ref:`axes &lt;/NXdata@axes-attribute&gt;` attribute
+		is(are) used. The shape of an :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` field must be equal to the shape of the
+		:ref:`data &lt;/NXdata/DATA-field&gt;` dimensions it spans.
+
+		Note that an :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` field can have more than one dimension and can therefore span
+		more than one :ref:`data &lt;/NXdata/DATA-field&gt;` dimension. Conversely, one :ref:`data &lt;/NXdata/DATA-field&gt;` dimension
+		can be spanned by more than one :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` field. As stated before, the default
+		axis name (if any) can be found in the :ref:`axes &lt;/NXdata@axes-attribute&gt;` attribute.
+
+		`Available axes`
+
+		With the current definition of the :ref:`axes &lt;/NXdata@axes-attribute&gt;` and :ref:`AXISNAME_indices &lt;/NXdata@AXISNAME_indices-attribute&gt;` attributes
+		it is not straightforward to discover all available axes. All strings in the :ref:`axes &lt;/NXdata@axes-attribute&gt;` attribute
+		(excluding the “.” string) are axis field names. In addition the prefix of an attribute ending with the string "_indices" is also
+		an axis field name.
+
+		This indirect way of discovering axes exists so that data which does not have the :ref:`AXISNAME_indices &lt;/NXdata@AXISNAME_indices-attribute&gt;` attributes
+		does not become invalid. Such data requires the length of the :ref:`axes &lt;/NXdata@axes-attribute&gt;` attribute to be equal to the rank of the
+		:ref:`data &lt;/NXdata/DATA-field&gt;` fields.
+
+		**Axes examples:**
+
+		`1. Single-dimensional axes`
+
+		:ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` fields are typically one-dimensional arrays that span a single :ref:`data &lt;/NXdata/DATA-field&gt;` dimension.
+
+		An example of this would be 2D data on a regular coordinate grid
 
 		.. code-block::
 
 			data:NXdata
 			  @signal = "data"
-			  @axes = ["x", "y"]  --> the order matters
+			  @axes = ["x", "y"]
 			  data: float[10,20]
 			  x: float[10]        --> coordinates along the first dimension
 			  y: float[20]        --> coordinates along the second dimension
 
 		In this example each data point ``data[i,j]`` has axis coordinates ``[x[i], y[j]]``.
 
-		However, the fields can also have a rank greater than 1, in which case the rank of each
-		:ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` must be equal to the number of data dimensions it spans.
+		Note that `@x_indices` and `@y_indices` attributes can be omitted in this case. However it is strongly encouraged to provide them.
 
-		An example of this would be
+		`2. Multi-dimensional axes`
+
+		When coordinates do not form a regular grid, multi-dimensional :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` fields
+		can be used. Just remember that the shape of an :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` field must be equal to the shape
+		of the :ref:`data &lt;/NXdata/DATA-field&gt;` dimensions it spans.
+
+		An example of this would be 2D scatter data where each data point has unique coordinates
 
 		.. code-block::
 
 			data:NXdata
 			  @signal = "data"
-			  @axes = ["x", "y"]  --> the order does NOT matter
+			  @axes = ["x", "y"]
 			  @x_indices = [0, 1]
 			  @y_indices = [0, 1]
 			  data: float[10,20]
 			  x: float[10,20]     --> coordinates along both dimensions
 			  y: float[10,20]     --> coordinates along both dimensions
 
-		In this example each data point ``data[i,j]`` has axis coordinates ``[x[i,j], y[i,j]]``.
+		In this example each data point ``data[i,j]`` has axis coordinates ``[x[i,j], y[i,j]]`` and when
+		plotting, `x` is used along the first data dimension by default and `y` along the second data dimension.
+		Since `x` and `y` span both dimensions, a reader could choose to use `y` for the first dimension
+		and `x` for the second as an alternative to the default. A writer could also choose to not specify
+		any default by specifying `@axes = [".", "."]` and leave the decision up the the reader.
 
-		`Dimensions`
+		Note that omitting `@x_indices` would result in `@x_indices = [0]` as derived from the position of `"x"` in `@axes`.
+		This would be invalid since the shape `[10,20]` of `x` is not equal to the shape `[10]` of the spanned data dimensions.
 
-		The data dimensions annotated by an :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` field are defined by the
-		:ref:`AXISNAME_indices &lt;/NXdata@AXISNAME_indices-attribute&gt;` attribute. When this attribute is missing,
-		the position(s) of the :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` string in the
-		:ref:`axes &lt;/NXdata@axes-attribute&gt;` attribute are used.
-
-		When all :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` fields are one-dimensional, and none of the data dimensions
-		have more than one axis, the :ref:`AXISNAME_indices &lt;/NXdata@AXISNAME_indices-attribute&gt;` attributes
-		are often omitted. If one of the data dimensions has no :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` field,
-		the string “.” can be used in the corresponding index of the axes list.
-
-		An example of this would be
+		Omitting indices for multi-dimensional axes can only be done by repeating the :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;`
+		name in all positions of the :ref:`axes &lt;/NXdata@axes-attribute&gt;` attribute which they span.
 
 		.. code-block::
 
 			data:NXdata
 			  @signal = "data"
-			  @axes = ["x", ".", "z"]  --> the order matters
-			  data: float[10,20,30]
-			  x: float[10]             --> coordinates along the first dimension
-			  z: float[30]             --> coordinates along the third dimension
+			  @axes = ["x", "x"]
+			  data: float[10,20]
+			  x: float[10,20]     --> coordinates along both dimensions
 
-		When using :ref:`AXISNAME_indices &lt;/NXdata@AXISNAME_indices-attribute&gt;` this becomes
+		Note that since ``NXdata`` does not describe how the data is to be plotted or even the dimensionality of the plot,
+		a reader would probably treat this particular example as a 1D signal and plot it as such.
 
-		.. code-block::
+		`3. Dimensions spanned by more than one axis`
 
-			data:NXdata
-			  @signal = "data"
-			  @axes = ["x", "z"]       --> the order does NOT matter
-			  data: float[10,20,30]
-			  @x_indices = 0
-			  @z_indices = 2
-			  x: float[10]             --> coordinates along the first dimension
-			  z: float[30]             --> coordinates along the third dimension
+		In the case of multi-dimensional axes, single-dimensional axes are often introduced as default axes
+		to support readers that does not use the :ref:`AXISNAME_indices &lt;/NXdata@AXISNAME_indices-attribute&gt;` attributes
+		or cannot handle multi-dimensional axes.
 
-		When providing :ref:`AXISNAME_indices &lt;/NXdata@AXISNAME_indices-attribute&gt;` attributes it is recommended
-		to do it for all axes.
-
-		`Non-trivial axes`
-
-		What follows are two examples where :ref:`AXISNAME_indices &lt;/NXdata@AXISNAME_indices-attribute&gt;` attributes
-		cannot be omitted.
-
-		The first is an example where data dimensions have alternative axis coordinates. The NXdata group represents
-		a stack of images collected at different energies. The ``wavelength`` is an alternative axis of ``energy``
-		for the last dimension (or vice versa).
+		The 2D scatter example can be enriched as follows
 
 		.. code-block::
 
 			data:NXdata
 			  @signal = "data"
-			  @axes = ["x", "y", "energy", "wavelength"]  --> the order does NOT matter
-			  @x_indices = 0
-			  @y_indices = 1
-			  @energy_indices = 2
-			  @wavelength_indices = 2
-			  data: float[10,20,30]
-			  x: float[10]              --> coordinates along the first dimension
-			  y: float[20]              --> coordinates along the second dimension
-			  energy: float[30]         --> coordinates along the third dimension
-			  wavelength: float[30]     --> coordinates along the third dimension
+			  @axes = ["x_set", "y_set"]
+			  @x_indices = [0, 1]
+			  @y_indices = [0, 1]
+			  @x_set_indices = 0  --> or [0]
+			  @y_set_indices = 1  --> or [1]
+			  data: float[10,20]
+			  x: float[10,20]     --> coordinates along both dimensions
+			  y: float[10,20]     --> coordinates along both dimensions
+			  x_set: float[10]    --> coordinates along the first dimension
+			  y_set: float[20]    --> coordinates along the second dimension
 
-		The second is an example with coordinates that span more than one dimension. The NXdata group represents data
-		from 2D mesh scans performed at multiple energies. Each data point ``data[i,j,k]`` has axis coordinates
-		``[x[i,j,k], y[i,j,k], energy[k]]``.
+		Technically `@x_set_indices` and `@y_set_indices` can be omitted. However it is strongly encouraged to provide them.
+
+		Another common case is the need to specify alternative axes for the same dimension
 
 		.. code-block::
 
 			data:NXdata
 			  @signal = "data"
-			  @axes = ["x", "y", "energy"]  --> the order does NOT matter
-			  @x_indices = [0, 1, 2]
-			  @y_indices = [0, 1, 2]
-			  @energy_indices = 2
-			  data: float[10,20,30]
-			  x: float[10,20,30]            --> coordinates along all dimensions
-			  y: float[10,20,30]            --> coordinates along all dimensions
-			  energy: float[30]             --> coordinates along the third dimension
+			  @axes = ["eta", "q"]
+			  data: float[10,20]
+			  @eta_indices = [0]
+			  @tth_indices = [1]
+			  @q_indices = [1]
+			  eta: float[10]    --> coordinates along the first dimension
+			  tth: float[20]    --> coordinates along the second dimension (alternative)
+			  q: float[20]      --> coordinates along the second dimension (default)
+
+		Note that to recognize `tth` as an axis, `@tth_indices` must be present. Readers that do
+		not use :ref:`AXISNAME_indices &lt;/NXdata@AXISNAME_indices-attribute&gt;` attributes will
+		have no way to understand that `tth` is an axis and simple ignore it.
+
+		`4. Axes without defaults`
+
+		Expanding on the 2D scatter example, a stack of 2D scatter data where the stack dimension
+		does not have an axis can be described as follows
+
+		.. code-block::
+
+			data:NXdata
+			  @signal = "data"
+			  @axes = [".", "x_set", "y_set"]
+			  @x_indices = [1, 2]
+			  @y_indices = [1, 2]
+			  @x_set_indices = 1
+			  @y_set_indices = 2
+			  data: float[5,10,20]
+			  x: float[10,20]     --> coordinates along the last two dimensions
+			  y: float[10,20]     --> coordinates along the last two dimensions
+			  x_set: float[10]    --> coordinates along the second dimension
+			  y_set: float[20]    --> coordinates along the third dimension
+
+		Note that the first dimension does not have a default axis but this doesn't necessarily mean
+		the first dimension is not spanned by any axis, although in this example that is the case.
 
 		**Uncertainties:**
 
@@ -319,8 +365,8 @@
 			The ``AXISNAME_indices`` attribute is a single integer or an array of integers that defines which :ref:`data &lt;/NXdata/DATA-field&gt;`
 			dimension(s) are spanned by the corresponding axis. The first dimension index is ``0`` (zero).
 
-			When the ``AXISNAME_indices`` attribute is missing for an :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` field, its value becomes the index
-			(or indices) of the :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` name in the :ref:`axes &lt;/NXdata@axes-attribute&gt;` attribute.
+			When the ``AXISNAME_indices`` attribute is missing for a given :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` field, its value becomes
+			the index (or indices) of the :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` name in the :ref:`axes &lt;/NXdata@axes-attribute&gt;` attribute.
 
 			.. note::  When ``AXISNAME_indices`` contains multiple integers, it must be saved as an actual array
 				of integers and not a comma separated string.
@@ -331,7 +377,9 @@
 			.. index:: plotting
 			
 			The ``axes`` attribute is a list of strings which are the names of the :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` fields
-			that contain the values of the coordinates along the :ref:`data &lt;/NXdata/DATA-field&gt;` dimensions.
+			to be used as the default axis along all :ref:`data &lt;/NXdata/DATA-field&gt;` dimensions. As a result the length must
+			be equal to the rank of the :ref:`data &lt;/NXdata/DATA-field&gt;` fields. The string "." can be used for
+			dimensions without default axes.
 
 			.. note::  When ``axes`` contains multiple strings, it must be saved as an actual array
 				of strings and not a single comma separated string.
@@ -341,14 +389,12 @@
 	<!-- Data and coordinate fields -->
 	<field name="AXISNAME" type="NX_CHAR_OR_NUMBER" nameType="any">
 		<doc>
-			Coordinate values along one or more :ref:`data &lt;/NXdata/DATA-field&gt;` dimensions. The rank must be equal
-			to the number of dimensions it spans.
-			
-			As the upper case ``AXISNAME`` indicates, the names of the ``AXISNAME`` fields can be chosen :ref:`freely &lt;validItemName&gt;`.
-			The :ref:`axes &lt;/NXdata@axes-attribute&gt;` attribute can be used to find all datasets in the
-			``NXdata`` that contain coordinate values.
+			Coordinate values along one or more :ref:`data &lt;/NXdata/DATA-field&gt;` dimensions. The shape of an ``AXISNAME`` field
+			must be equal to the shape of the :ref:`data &lt;/NXdata/DATA-field&gt;` dimensions it spans.
 
-			Most AXISNAME fields will be sequences of numbers but if an axis is better represented using names, such as channel names,
+			As the upper case ``AXISNAME`` indicates, the names of the ``AXISNAME`` fields can be chosen :ref:`freely &lt;validItemName&gt;`.
+
+			Most ``AXISNAME`` fields will be sequences of numbers but if an axis is better represented using names, such as channel names,
 			an array of NX_CHAR can be provided.
 		</doc>
 		<attribute name="long_name"><doc>Axis label</doc></attribute>


base_classes/NXdata.nxdl.xml
```diff
@@ -85,124 +85,170 @@
...
 
-		An example of this would be
+		`Spanned dimensions`
+
+		The :ref:`AXISNAME_indices &lt;/NXdata@AXISNAME_indices-attribute&gt;` attributes describe the
+		:ref:`data &lt;/NXdata/DATA-field&gt;` dimensions spanned by the corresponding :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` fields.
+
+		When :ref:`AXISNAME_indices &lt;/NXdata@AXISNAME_indices-attribute&gt;` is missing for a given :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` field,
+		the position(s) of the :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` field name in the :ref:`axes &lt;/NXdata@axes-attribute&gt;` attribute
+		is(are) used. The shape of an :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` field must be equal to the shape of the
```
PeterC-DLS, line 104: The values at those dimensions must match: `AXISNAME.shape[i] == SIGNAL.shape[AXISNAME_indices[i]]` for each `i` in `[0, AXISNAME.ndim)`

base_classes/NXdata.nxdl.xml
```diff
@@ -85,124 +85,170 @@
...
+		:ref:`data &lt;/NXdata/DATA-field&gt;` dimensions it spans.
+
+		Note that an :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` field can have more than one dimension and can therefore span
+		more than one :ref:`data &lt;/NXdata/DATA-field&gt;` dimension. Conversely, one :ref:`data &lt;/NXdata/DATA-field&gt;` dimension
+		can be spanned by more than one :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` field. As stated before, the default
+		axis name (if any) can be found in the :ref:`axes &lt;/NXdata@axes-attribute&gt;` attribute.
+
+		`Available axes`
+
+		With the current definition of the :ref:`axes &lt;/NXdata@axes-attribute&gt;` and :ref:`AXISNAME_indices &lt;/NXdata@AXISNAME_indices-attribute&gt;` attributes
```
PeterC-DLS, line 114: "With this definition ..."

base_classes/NXdata.nxdl.xml
```diff
@@ -85,124 +85,170 @@
...
+		it is not straightforward to discover all available axes. All strings in the :ref:`axes &lt;/NXdata@axes-attribute&gt;` attribute
```
PeterC-DLS, line 115: "~it is not straightforward to discover~does not provide a list of all available axes"

base_classes/NXdata.nxdl.xml
```diff
@@ -85,124 +85,170 @@
...
+		plotting, `x` is used along the first data dimension by default and `y` along the second data dimension.
+		Since `x` and `y` span both dimensions, a reader could choose to use `y` for the first dimension
+		and `x` for the second as an alternative to the default. A writer could also choose to not specify
+		any default by specifying `@axes = [".", "."]` and leave the decision up the the reader.
 
-		When all :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` fields are one-dimensional, and none of the data dimensions
-		have more than one axis, the :ref:`AXISNAME_indices &lt;/NXdata@AXISNAME_indices-attribute&gt;` attributes
-		are often omitted. If one of the data dimensions has no :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` field,
-		the string “.” can be used in the corresponding index of the axes list.
+		Note that omitting `@x_indices` would result in `@x_indices = [0]` as derived from the position of `"x"` in `@axes`.
```
PeterC-DLS, line 169: I think it is better to say an `AXISNAME_indices` attribute must be provided for multi-dimensional `AXISNAME` fields.

base_classes/NXdata.nxdl.xml
```diff
@@ -85,124 +85,170 @@
...
+		This would be invalid since the shape `[10,20]` of `x` is not equal to the shape `[10]` of the spanned data dimensions.
 
-		An example of this would be
+		Omitting indices for multi-dimensional axes can only be done by repeating the :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;`
```
PeterC-DLS, line 172: I think `@axes = [".", "."]` is valid here. Generous readers may attempt to derive indices from positions of matching shape values - this depends on `DATA.shape` possessing unique values.

base_classes/NXdata.nxdl.xml
```diff
@@ -85,124 +85,170 @@
...
-			  x: float[10]             --> coordinates along the first dimension
-			  z: float[30]             --> coordinates along the third dimension
+			  @axes = ["x", "x"]
+			  data: float[10,20]
+			  x: float[10,20]     --> coordinates along both dimensions
+
+		Note that since ``NXdata`` does not describe how the data is to be plotted or even the dimensionality of the plot,
+		a reader would probably treat this particular example as a 1D signal and plot it as such.
+
+		`3. Dimensions spanned by more than one axis`
```
PeterC-DLS, line 186: How does this differ from the title in L144: `2. Multi-dimensional axes`

base_classes/NXdata.nxdl.xml
rayosborn: This is worth pointing out, but it needs amending. Since the earliest days of the format, NeXus allowed axis dimensions to have a size of the corresponding data dimension + 1, because neutron facilities typically store time-of-flight bin boundaries, rather than centres.

base_classes/NXdata.nxdl.xml
woutdenolf: But technically it is still optional right?

base_classes/NXdata.nxdl.xml
woutdenolf: one axis spanning multiple dimensions vs. one dimension covered by more that one axis

base_classes/NXdata.nxdl.xml
woutdenolf: Yes that is what I mean with "matching shapes" as in [numpy.shape](https://numpy.org/doc/stable/reference/generated/numpy.shape.html) but I guess this is not commonly understood.

base_classes/NXdata.nxdl.xml
woutdenolf: I wanted to be clear that we are well aware this is a horrible and non-deterministic way of discovering axes. "not straightforward" is already a euphemism imo. But ok I'll play nice.

base_classes/NXdata.nxdl.xml
woutdenolf: If you omit `x_indices` and you have `@axes = [".", "."]` there is not way for a reader to understand that `x` is an axis (see discovery of axes).

base_classes/NXdata.nxdl.xml
woutdenolf: Your correction did not match the sentence. I did

~it is not straightforward to discover all available axes~ a list of all available axes is not provided directly

base_classes/NXdata.nxdl.xml
woutdenolf: I added this to make it more clear:

"The first dimension is covered by three axes: `x`, `y` and `x_set`. The second dimension is also covered by three axes: `x`, `y` and `y_set`."

base_classes/NXdata.nxdl.xml
woutdenolf: ok

base_classes/NXdata.nxdl.xml
woutdenolf: I added your suggestion after this sentence:

The shape of an AXISNAME field must be equal to the shape of the DATA dimensions it spans. This means that `AXISNAME.shape[i] == DATA.shape[AXISNAME_indices[i]]` for each `i` in `[0, AXISNAME.ndim)`.

base_classes/NXdata.nxdl.xml
woutdenolf: > Generous readers may attempt to derive indices from positions of matching shape values

Please tell me this is a joke.

base_classes/NXdata.nxdl.xml
woutdenolf: In addition I will put the second example first because indeed, `2. Multi-dimensional axes` has both Multi-dimensional axes and axes spanned by more than one dimension.

base_classes/NXdata.nxdl.xml
PeterC-DLS: Ray suggests that `== x` should be slacken to `== x or == x + 1`

base_classes/NXdata.nxdl.xml
PeterC-DLS: I don't see why a reader couldn't guess that `tth` was for the last dimension of `data` in this example

PeterC-DLS:
Also add to `@axes` the following from my [#1392](https://github.com/nexusformat/definitions/pull/1392)
```
    <dimensions rank="1">
        <dim index="1" value="dataRank"/>
    </dimensions>
```

base_classes/NXdata.nxdl.xml
```diff
@@ -85,124 +85,170 @@
...
+			  @x_indices = [0, 1]
+			  @y_indices = [0, 1]
+			  data: float[10,20]
+			  x: float[10,20]     --> coordinates along both dimensions
+			  y: float[10,20]     --> coordinates along both dimensions
 
-		When using :ref:`AXISNAME_indices &lt;/NXdata@AXISNAME_indices-attribute&gt;` this becomes
+		In this example each data point ``data[i,j]`` has axis coordinates ``[x[i,j], y[i,j]]`` and when
+		plotting, ``x`` is used along the first data dimension by default and `y` along the second data dimension.
+		Since ``x`` and ``y`` span both dimensions, a reader could choose to use ``y`` for the first dimension
```
PeterC-DLS, line 183: I think it's the user who chooses - the reader may allow a choice but it should obey the default order given by `@axes`

base_classes/NXdata.nxdl.xml
```diff
@@ -85,124 +85,170 @@
...
+		and ``x`` for the second as an alternative to the default. A writer could also choose to not specify
+		any default by defining ``@axes = [".", "."]`` and leave the decision up the the reader.
```
PeterC-DLS, line 185: Or omit `@axes`

base_classes/NXdata.nxdl.xml
```diff
@@ -341,14 +389,12 @@
 	<!-- Data and coordinate fields -->
 	<field name="AXISNAME" type="NX_CHAR_OR_NUMBER" nameType="any">
 		<doc>
-			Coordinate values along one or more :ref:`data &lt;/NXdata/DATA-field&gt;` dimensions. The rank must be equal
-			to the number of dimensions it spans.
-			
+			Coordinate values along one or more :ref:`DATA &lt;/NXdata/DATA-field&gt;` dimensions. The shape of an ``AXISNAME`` field
+			must be equal to the shape of the :ref:`DATA &lt;/NXdata/DATA-field&gt;` dimensions it spans.
```
PeterC-DLS, line 393: Loosen this to include Ray's +1 suggestion

woutdenolf:
NIAC Jul 2024 comments

* AXISNAME_indices are not allowed to contradict the position of "AXISNAME" in @axes @rayosborn 
* "Axes without defaults" -> DATA dimensions with a default axis
* "However it is strongly encouraged to provide them." -> When AXISNAME_indices is the same a the indices of "AXISNAME" in @axes there is no need to provide them. @benajamin

woutdenolf:
@rayosborn As for you comment about stating that "AXISNAME_indices are not allowed to contradict the position of AXISNAME in @axes". So this is the canonical example

```
data:NXdata
  @signal = "data"
  @axes = ["x", "y"]
  @x_indices=0
  @y_indices=1
  data: float[10,20]
  x: float[10]
  y: float[20]
```

Then this should be invalid


```
data:NXdata
  @signal = "data"
  @axes = ["y", "x"]  # position does not correspond indices
  @x_indices=0
  @y_indices=1
  data: float[10,20]
  x: float[10]
  y: float[20]
```

But then is this invalid as well?

```
data:NXdata
  @signal = "data"
  @axes = ["x", "y"]
  @x_indices = [0, 1]
  @y_indices = [0, 1]
  data: float[10,20]
  x: float[10,20]
  y: float[10,20]
```

woutdenolf:
Refactor again to take NIAC comments into account:

* start with simple example at the very top (@phyy-nx)
* histogram edges (+1 value, @rayosborn )
* when AXISNAME_indices is the same a the indices of "AXISNAME" in @axes there is no need to provide them (@benajamin )
*  indices of "AXISNAME" in axes must be a subset of AXISNAME_indices (comment by @rayosborn to prohibit contradiction between AXISNAME_indices and axes)
* add `<dimensions rank="1">` to `@axes` (@PeterC-DLS )

The structure of the NXdata doc section at the top is now

* Use case: canonical example + list supported cases
* Signals:
  * Defined by ...
  * Definition
  * Example 
* Axes:
  * Defined by ...
  * Definition
  * Example 
* Uncertainties
  * Defined by ...
  * Definition
  * Example 

One example per section is enough. The example in the axes section covers different axes use cases supported by NXdata (histogram edges +1, alternative axes, multi-dimensional axes, "." in axes, missing indices). I made this example very concrete by referring to a specific scientific case and included an image.

base_classes/NXdata.nxdl.xml
```diff
@@ -60,176 +60,181 @@
...
+				  data: float[100]
+				  x: float[100]
+
+		More complex cases are supported
+
+		* histogram data: ``x`` has one more value than ``data``.
+		* alternative axes: instead of a single ``x`` axis you can have several axes, one of which being the default.
+		* signals with more than one dimension: ``data`` could be 2D with axes ``x`` and ``y`` along each dimension.
+		* axes with more than one dimension: ``data`` could be 2D with axes ``x`` and ``y`` also being 2D, providing a
+		  unique ``[x, y]`` coordinate for each ``data`` point.
```
PeterC-DLS, line 81: Use parentheses here: `(x,y)`

base_classes/NXdata.nxdl.xml
```diff
@@ -60,176 +60,181 @@
...
 
-		In this example each data point ``data[i,j]`` has axis coordinates ``[x[i,j], y[i,j]]``.
+		3. When :ref:`AXISNAME_indices &lt;/NXdata@AXISNAME_indices-attribute&gt;` is missing for a given
+		   :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` field, the positions of the :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;`
+		   field name in the :ref:`axes &lt;/NXdata@axes-attribute&gt;` attribute are used.
 
-		`Dimensions`
+		4. When :ref:`AXISNAME_indices &lt;/NXdata@AXISNAME_indices-attribute&gt;` is the same as the indices of "AXISNAME" in the
+		   :ref:`axes &lt;/NXdata@axes-attribute&gt;` attribute, there is no need to provide
+		   :ref:`AXISNAME_indices &lt;/NXdata@AXISNAME_indices-attribute&gt;`.
```
PeterC-DLS, line 144: Only for 1D axis fields?

base_classes/NXdata.nxdl.xml
```diff
@@ -60,176 +60,181 @@
...
 
-		When all :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` fields are one-dimensional, and none of the data dimensions
-		have more than one axis, the :ref:`AXISNAME_indices &lt;/NXdata@AXISNAME_indices-attribute&gt;` attributes
-		are often omitted. If one of the data dimensions has no :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` field,
-		the string “.” can be used in the corresponding index of the axes list.
+		6. The shape of an :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` field must correspond to the shape of the
+		   :ref:`DATA &lt;/NXdata/DATA-field&gt;` dimensions it spans. This means that for each dimension ``i`` in ``[0, AXISNAME.ndim)``
+		   spanned by an :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` field the number of data points along this dimension
+		   ``DATA.shape[AXISNAME_indices[i]]`` must be equal to the number of axis values along this dimension ``AXISNAME.shape[i]``
+		   or the number of bin edges ``AXISNAME.shape[i]+1`` in case of histogram data.
```
PeterC-DLS, line 153: Slightly wrong: for histogram data, `#edges - 1 = #data_points`

base_classes/NXdata.nxdl.xml
```diff
@@ -60,176 +60,181 @@
...
 
-		An example of this would be
+		Highlight concequences that follow the definition
```
PeterC-DLS, line 155: Spelling "consequences"

base_classes/NXdata.nxdl.xml
```diff
@@ -85,124 +85,170 @@
...
-		:ref:`axes &lt;/NXdata@axes-attribute&gt;` attribute.
+		The :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` fields contain the axis coordinates associated with the signal values.
 
-		`Rank`
+		`Default axes`
 
-		:ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` fields are typically one-dimensional arrays, which annotate one of the dimensions.
+		The :ref:`axes &lt;/NXdata@axes-attribute&gt;` attribute provides the names of the :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;`
+		fields to be used as the default axis for each dimension of the :ref:`data &lt;/NXdata/DATA-field&gt;` fields.
+		As a result the length of the :ref:`axes &lt;/NXdata@axes-attribute&gt;` attribute must be equal to the rank of the :ref:`data &lt;/NXdata/DATA-field&gt;`
```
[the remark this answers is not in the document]
PeterC-DLS: That would be better

base_classes/NXdata.nxdl.xml
```diff
@@ -60,176 +60,181 @@
...
 			https://www.nexusformat.org/NIAC2018Minutes.html#nxdata-plottype--attribute
 
+		**Usage:**
+
+		.. admonition:: Plot data versus x axis
+
+			.. code-block::
+
+				data:NXdata
+				  @signal = "data"
```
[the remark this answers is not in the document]
woutdenolf: I did not know this. Not specified in the definition. I'll add it.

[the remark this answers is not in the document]
woutdenolf: I guess when `@axes` is missing, we use `[".", ".", ...]` (as many dots as rank)?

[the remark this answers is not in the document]
prjemian: If `@signal` is missing, there is no rule that says `"data"` is required as a field name.  In this case and multiple fields, selection of the _signal_ field for the default plot is not guaranteed.

base_classes/NXdata.nxdl.xml
woutdenolf: If this applies in general then `@axes=["x", "x"]` would imply `@x_indices=[0,1]`. That would be ok when `x` is 2D.

When `x` is 1D it would not work. It would violate two requirements

* The number of values in AXISNAME_indices must be equal to the rank of the corresponding AXISNAME field.
* The indices of "AXISNAME" in the axes attribute must be a subset of AXISNAME_indices.

So if a 1D `x` needs to work for `@axes=["x", "x"]` we need to change those requirements.

base_classes/NXdata.nxdl.xml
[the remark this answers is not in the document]
PeterC-DLS: I thought `NXdata` evolved from a fixed signal field (usually called `data`, see e.g. https://github.com/nexusformat/definitions/blob/04bfe4ccbbfd3cb715d1a1f42990a2d18fa5ed72/base_classes/NXdata.nxdl.xml#L152) with an (optional?) attribute `@signal=1` so to be backwardly compatible, a modern parser could assume the group attribute `signal` if missing is `"data"`.

base_classes/NXdata.nxdl.xml
PeterC-DLS: Hmm, a pathological case could be `@x_indices=[1,0]`

woutdenolf: Ok I see. So omitting indices can only work unambiguously in 1D cases.

woutdenolf: Another pathological case would be `@x_indices=[0,0]`. I'm not sure what this means but technically the definition does not forbid it. So we should say indices cannot contain duplicates?

prjemian: When `x` is 1-D, then `@x_indices=[0,0]` would be compatible with `@axes=["x", "x"]`, right?

prjemian: Why would it not be allowed?

prjemian: In the case of `@x_indices=[0,0]`, this says the same 1-D array applies to both independent axes.

base_classes/NXdata.nxdl.xml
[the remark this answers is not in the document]
rayosborn: I think that @prjemian is correct. I just scanned through the minutes of all the NIAC meetings, and I can't find any reference to a default name for the `signal`. I believe that the current (non-deprecated) method of defining signals and axes requires both `signal` and `axes` attributes to be defined at group level.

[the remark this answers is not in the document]
prjemian: It was found that assigning these links to fields would break when the fields were linked (local or external) into other groups.  Such field attributes could produce a group with multiple field `@signal` attributes.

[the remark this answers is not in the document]
PeterC-DLS: Perhaps we need to make these attributes `required` then.

rayosborn:
I would suggest changing the following text:

> The [AXISNAME_indices](https://nxdata-axes-fix-hdf5-nexus-e16d3be7bef8684a1ad37775f8cbe25e8412.gitlab-pages.esrf.fr/classes/base_classes/NXdata.html#nxdata-axisname-indices-attribute) attributes describe the [DATA](https://nxdata-axes-fix-hdf5-nexus-e16d3be7bef8684a1ad37775f8cbe25e8412.gitlab-pages.esrf.fr/classes/base_classes/NXdata.html#nxdata-data-field) dimensions spanned by the corresponding [AXISNAME](https://nxdata-axes-fix-hdf5-nexus-e16d3be7bef8684a1ad37775f8cbe25e8412.gitlab-pages.esrf.fr/classes/base_classes/NXdata.html#nxdata-axisname-field) fields.

to:

> The optional [AXISNAME_indices](https://nxdata-axes-fix-hdf5-nexus-e16d3be7bef8684a1ad37775f8cbe25e8412.gitlab-pages.esrf.fr/classes/base_classes/NXdata.html#nxdata-axisname-indices-attribute) attributes describe the [DATA](https://nxdata-axes-fix-hdf5-nexus-e16d3be7bef8684a1ad37775f8cbe25e8412.gitlab-pages.esrf.fr/classes/base_classes/NXdata.html#nxdata-data-field) dimensions spanned by the corresponding [AXISNAME](https://nxdata-axes-fix-hdf5-nexus-e16d3be7bef8684a1ad37775f8cbe25e8412.gitlab-pages.esrf.fr/classes/base_classes/NXdata.html#nxdata-axisname-field) fields. These attributes may be used to indicate alternative axis coordinates to the default.

base_classes/NXdata.nxdl.xml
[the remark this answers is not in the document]
prjemian: Could bring that up again for a NIAC vote.  Not until then, since NIAC voted the other way.  

Consider a simple use case (for not having this attribute).  User wants to save a 2-D image, only the 2-D image.  NeXus file consists of `/entry:NXentry/data:NXdata/image` where `image` is the 2-D image.  We've gotten strong pushback on requiring `@signal` in this use case.

rayosborn:
I would suggest changing `histogram bin edges` to `histogram bin boundaries`, which is more common usage in the neutron time-of-flight community.

rayosborn:
Just a general comment. I think @woutdenolf has done a great job of clarifying the various scenarios for specifying axes, but I found the list following `Additional requirements for a complete definition` to be difficult to follow, even as someone who has participated in this discussion for years. I wonder if it should have a separate heading, such as "Multidimensional Axes" so that people know they only need to get their heads round this if they have specific needs for more complex axis specifications. To emphasize this, I think it might be better if the simple example at the top is two-dimensional, with both `x` and `y` axes, to prevent people from thinking that multidimensional data are intrinsically complicated to plot. The only drawback is that it opens a can of worms about what is the horizontal axis, `x` or `y`, when `axes=["x", "y"]`.

base_classes/NXdata.nxdl.xml
prjemian: Eeeeww!  That's pathological, and likely a mistake.

woutdenolf: > Eeeeww! That's pathological, and likely a mistake.

Agreed but technically the definition does not present it as far as I can see, which we should probably fix but not sure how.

Stating that `AXISNAME_indices` cannot have duplicates would solve it. But it would prohibit the case where you have for example a single 1D x axis of lets say 100 points and you have a 2D signal of 100 x 100 points and you want x to be used for both dimensions. As I said before, you could define x1 and x2 and both can be symlinks to x so effectively x is used for both dimensions even though `@axes=["x1", "x2"]`.

woutdenolf:
NIAC Aug 2024 comments: target audience are scientists on the one hand and software developers that need to writer/read HDF5 files. The rigorous definition should not alienate the first group. Split the description at the top accordingly.

woutdenolf:
As an aside for @rayosborn : code in Silx that takes coordinates on makes an image from it https://github.com/silx-kit/silx/blob/11919895f51353c8467b85a3753f142778c2bbaf/src/silx/gui/plot/items/scatter.py#L547-L797

woutdenolf:
Suggestion for progressing:

1. First you want to understand what NXdata can do
2. Then you need to write scripts/libraries to read/write NXdata

For the first step we could add start the docs with several examples (image + nexus structure). I need suggestions before I make the effort to generate data. It needs to be simple but we also want to make sure we don't guide the user in the wrong direction. For example the image example may assume equal coordinate spacing due to the square pixels but the data. But you have a 1D array of coordinates for an axis (N values), not an offset and pixel size (2 values).

nxdata_examples

For the second use case we need a spec that does not leave room for ambiguity. The current spec is too complex and needs a better formulation (especially separating what is asserted from what is inferred). I propose to express the current spec in code and let NIAC members check it before investigating different way of describing the spec.

```python
from itertools import combinations
import numpy
import h5py


def nxdata_coordinates(nxdata: h5py.Group) -> None:
    """Illustrate the axes specs. Not written for efficiency."""

    # Get the values of the dependent variable.

    nxdata_attrs = dict(nxdata.attrs)
    signal = nxdata_attrs.get("signal")
    if signal is None:
        raise RuntimeError("NXdata does not have a signal")
    data = nxdata[signal]
    data_rank = data.ndim
    data_shape = data.shape

    # The length of the axes attribute must be equal to the data rank.
    axes_attr = nxdata_attrs.get("axes", ["."] * data_rank)
    if len(axes_attr) != data_rank:
        raise ValueError(f"{len(axes_attr)} default axes but the rank is ({data_rank})")

    # All axis names

    axisnames_with_explicit_dims = {
        name[:-8]
        for name in nxdata_attrs
        if name.endswith("_indices")
        and name[:-8] in nxdata
        and isinstance(nxdata[name[:-8]], h5py.Dataset)
    }

    axisnames = set(axes_attr) - {"."} | axisnames_with_explicit_dims

    # Map axis names to the dimensions they span.

    axisname_to_dimensions: dict[str, list[int]] = {}

    for axisname in axisnames:
        axis_shape = nxdata[axisname].shape

        default_dimensions = [i for i, name in enumerate(axes_attr) if name == axisname]

        if axisname in axisnames_with_explicit_dims:
            axisname_indices = nxdata_attrs[f"{axisname}_indices"]
            if isinstance(axisname_indices, numpy.ndarray):
                dimensions = axisname_indices.tolist()
            else:
                dimensions = [axisname_indices]

            if len(set(axisname_indices)) < len(axisname_indices):
                raise ValueError(
                    f"'{axisname}_indices' cannot have duplicate dimensions"
                )

            if not set(default_dimensions).issubset(dimensions):
                raise ValueError(
                    f"default dimensions spanned by axis '{axisname}' are {default_dimensions} but the explicit dimensions are {axisname_indices}"
                )
        else:
            dimensions = default_dimensions

        axisname_to_dimensions[axisname] = dimensions

    # Map dimensions to the axis names that span them.

    dimensions_to_axisnames: dict[int, list[str]] = {
        dimension: [] for dimension in range(data_rank)
    }
    for axisname, dimensions in axisname_to_dimensions.items():
        for dimension in dimensions:
            is_default = axisname == axes_attr[dimension]
            if is_default:
                dimensions_to_axisnames[dimension].insert(0, f"{axisname}*")
            else:
                dimensions_to_axisnames[dimension].append(axisname)

    # Axis dimensions must be equal to the data dimensions they span.

    for axisname, dimensions in axisname_to_dimensions.items():
        axis_shape = nxdata[axisname].shape
        spanned_shape = tuple(data_shape[dimension] for dimension in dimensions)
        if axis_shape not in _possible_shapes(spanned_shape):
            raise ValueError(
                f"axis '{axisname}' has shape {axis_shape} but the spanned data dimensions have shape {spanned_shape}"
            )

    # Print NeXus structure.

    print(f"{nxdata.name}:NXdata")
    for name, value in nxdata_attrs.items():
        print(
            f"  @{name}={value.tolist() if isinstance(value, numpy.ndarray) else value}"
        )
    for name, value in nxdata.items():
        print(f"  {name}: {value.dtype}{list(value.shape)}")

    # Print axes <-> dimensions mapping.

    print()
    for axisname, dimensions in axisname_to_dimensions.items():
        print(f"Axis '{axisname}' spans dimensions {dimensions}")

    print()
    for dimension, axes_names in dimensions_to_axisnames.items():
        print(f"Dimension '{dimension}' is spanned by axes {axes_names}")


def _possible_shapes(shape: tuple[int]) -> list[tuple[int]]:
    """Every dimension can have N+1 coordinates where N is the number of data points along any dimension."""
    shapes = [shape]
    base_shape = numpy.array(shape)

    rank = len(base_shape)
    indices = list(range(rank))
    for nhistogram in range(1, rank + 1):
        for idx_boundaries in combinations(indices, nhistogram):
            shape = base_shape.copy()
            shape[list(idx_boundaries)] += 1
            shapes.append(tuple(shape.tolist()))

    return shapes


def generate_nxdata() -> tuple[str, str]:
    hdf5_filename = "test.h5"
    with h5py.File(hdf5_filename, "w") as nxroot:
        nxroot.attrs["NX_class"] = "NXroot"
        nxentry = nxroot.create_group("scan1")
        nxentry.attrs["NX_class"] = "NXentry"
        nxdata = nxroot.create_group("data")
        nxdata.attrs["NX_class"] = "NXdata"

        nxdata.attrs["signal"] = "data"
        nxdata["data"] = numpy.zeros((10, 10, 20, 30, 30))

        nxdata["x"] = numpy.zeros((10, 10))
        nxdata["y"] = numpy.zeros((30,))  # cannot be used for more than 1 dimension
        nxdata["y1"] = h5py.SoftLink("y")
        nxdata["y2"] = h5py.SoftLink("y")

        nxdata.attrs["axes"] = ["x", ".", ".", "y1", "y2"]  # same as [0]
        nxdata.attrs["x_indices"] = [0, 1]  # superset of [0]
        # nxdata.attrs["y1_indices"] = [3]
        nxdata.attrs["y2_indices"] = [4]

        return hdf5_filename, nxdata.name


if __name__ == "__main__":
    hdf5_filename, data_path = generate_nxdata()
    with h5py.File(hdf5_filename, "r") as nxroot:
        nxdata = nxroot[data_path]
        nxdata_coordinates(nxdata)
```

```
/data:NXdata
  @NX_class=NXdata
  @axes=['x', '.', '.', 'y1', 'y2']
  @signal=data
  @x_indices=[0, 1]
  @y2_indices=[4]
  data: float64[10, 10, 20, 30, 30]
  x: float64[10, 10]
  y: float64[30]
  y1: float64[30]
  y2: float64[30]

Axis 'y2' spans dimensions [4]
Axis 'y1' spans dimensions [3]
Axis 'x' spans dimensions [0, 1]

Dimension '0' is spanned by axes ['x*']
Dimension '1' is spanned by axes ['x']
Dimension '2' is spanned by axes []
Dimension '3' is spanned by axes ['y1*']
Dimension '4' is spanned by axes ['y2*']
```

@jacobfilik Since you detected the original check, could you check whether the code describes the specs as you understand them?

base_classes/NXdata.nxdl.xml
woutdenolf: @prjemian Can you modify the code provided in https://github.com/nexusformat/definitions/pull/1396#issuecomment-2380113555 to prove that the current specs allow a 1D signal to be used for multiple dimensions while not breaking support for multi-dimensional axis?

I don't think it is possible.

prjemian: Not sure I know how to proceed on this directly.  For the variety of different formats considered by canSAS, the [discussion](https://wiki.cansas.org/index.php?title=2012_Data_Discussion_Examples) resulted in this [documentation](https://www.cansas.org/canSAS2012/examples.html) where the examples are repeated.  Those are the places I would examine to test multi-dimensional data.

phyy-nx:
👍  from the NIAC as to adding a gallery to the NXdata page of examples

phyy-nx:
This PR can be divided into two sets of fixes.  The first resolves the problem in #1381, and can be summarized as follows:

The `@axes` attribute should give the default axis to be used for every dimension in the given order as it is in the signal ("." is allowed).

The above will be resolved by the NIAC by vote in [NIAC2024](https://www.nexusformat.org/NIAC2024.html).

The second aspect of this PR is to greatly improve the documentation.  No vote needed for that, we're just happy go have it :)

rayosborn:
@woutdenolf, I see that you have posted some useful code examples, which I will study soon to see how they can be incorporated into the nexusformat package. In the main documentation, we have used Sphinx tabs to provide examples using both `nexusformat` and `h5py`. Do you know if that is possible in the context of what you are proposing?

woutdenolf:
**Status + progress**

The NIAC asked to make the `doc` section simpler.

I did this by keeping two types of people in mind:

1. People that create NeXus files:

    - use a gallery to cover the common cases
    - there is an issue reported by @tacaswell which wrongly adds the string "orphan:" the NXdata (https://github.com/sphinx-gallery/sphinx-gallery/issues/854)

2. People that create generic NeXus readers (h5web, silx, nexpy, etc.): 

    - an unambiguous specification is required
    - the complex "axes" section is divided in three sub-sections:
        - The fields and attributes are defined as follows
        - The fields and attributes have the following constraints
        - Highlight consequences of these constraints

image

woutdenolf:
Concerning histograms (axis has one value more that the signal): we do not define whether the signal contains the bin height or the bin area. I'm assuming it is the height. Should we add this to "The fields and attributes have the following constraints ..."?

rayosborn:
@woutdenolf, you have opened up another can of worms here. What is stored could well depend on the conventions of the community writing the data, and we have probably been lax in NeXus in providing a mechanism to define which convention is being used. 

In time-of-flight neutron scattering (at least when I was writing code), the convention was to divide the counts by the bin width when storing the data, so that we were effectively plotting in units of  'neutrons/microsecond', i.e., bin heights. When we wanted to convert to other units, e.g., to energy, we simply computed the energy values at all the bin boundaries and then multiplied by (dt/dE), a numerical Jacobian, to switch to units of 'neutrons/meV'. Because I come from that background, NeXpy just assumes the values are bin heights when plotting. 

However, another community might well choose to store the original counts, i.e., bin integrals, and there is no way of telling with the current NeXus standard. We probably need to define an attribute that says which convention is being used, with one of them being the default. Of course, when the bin widths are all constant, as they nearly always are, it doesn't really matter, but we nevertheless shouldn't assume that.

I will contact some friendly Mantid developers to make sure that 'neutrons/microsecond' is still the convention at neutron sources, and perhaps you could check with your powder diffraction colleagues to see if anyone is storing bin integrals.

ggoneiESS:
@PeterC-DLS has mentioned this PR in an issue I raised.  I think this is all good, pictorial examples will be very useful, but there are some things that need to be explicit (some are raised in the issue and some raised by the current solution):

- how non-contiguous data is recorded (it is not correct to say a bin has value 0 if it is not present in the data)
- whether the bin value should be taken from the centre, leading, or trailing edge
- how overflow is recorded

Raised by above discussion in this PR:
- how AXIS_errors* is defined (I can't see any alternative except ignoring it if AXIS has length len(DATA)+1 which would define NXdata as containing a histogram)
- in case of constant width bins, could axis just be defined by a start and end value, with bins given by len(DATA[INDEX])?
- how different should scatter-like and histogram-like data look in the resulting NXdata group?

I think there's inspiration to be taken from
https://www.mathworks.com/help/matlab/ref/matlab.graphics.chart.primitive.histogram-properties.html which has properties for Matlab histograms (so we can see if an NXdata group would contain enough to generate a plot in Matlab)
https://www.silx.org/doc/silx/latest/modules/io/nxdata.html (some definitions are better here already than in the nexusformat definitions)

There's also https://root.cern.ch/doc/master/classTH1.html (a very complete histogram class for ROOT in C++) and it might be interesting that they have separate plot types for histograms and graphs (point-like data), which might help (or provide motivation for) divergence of NXdata trying to add too much information and essentially start plotting data (which, as far as I am aware, is explicitly not the aim of the class).

Anecdotally, it's been very difficult to get started with NXdata, so I'm encouraged by this discussion :) whilst we are talking about AXISNAME, it would be very useful to add some clarity to
the attributes:
```
@distribution: (optional) [NX_BOOLEAN](https://manual.nexusformat.org/nxdl-types.html#nx-boolean)

``0|false``: single value, ...
0|false: single value, 1|true: multiple values

@first_good: (optional) [NX_INT](https://manual.nexusformat.org/nxdl-types.html#nx-int)

Index of first good value

@last_good: (optional) [NX_INT](https://manual.nexusformat.org/nxdl-types.html#nx-int)

Index of last good value
```

since it seems that `.*st_good` is irrelevant with both the current definitions and the proposed, and that distribution is also overruled by the rest of the definitions surrounding the axis.

ggoneiESS:
I reopened my issue, but I'll hold off submitting the PR until this issue is implemented.

I still wonder whether the extra information about the bin plotting (leading/trailing edge or centre) and the _errors_ info should form part of this, or whether it should all be pushed into my new PR (which, having compared the two, will also change the standard slightly rather than just the documentation).

galleries/nxdata/plot_fscan2d.py
```diff
@@ -0,0 +1,89 @@
...
+  @default = "scan1"
+  scan1:
+    @NX_class = "NXentry"
+    @default = "data"
+    data:
+      @NX_class = "NXdata"
+      @axes = ["x_set", "y_set"]
+      @x_encoder_indices = [0, 1]
+      @y_encoder_indices = 1
+      z: NX_FLOAT64[10,16]
```
sanbrock, line 29: note registered as signal

galleries/nxdata/plot_fscan2d.py
```diff
@@ -0,0 +1,89 @@
...
+
+  @NX_class = "NXroot"
+  @default = "scan1"
+  scan1:
+    @NX_class = "NXentry"
+    @default = "data"
+    data:
+      @NX_class = "NXdata"
+      @axes = ["x_set", "y_set"]
+      @x_encoder_indices = [0, 1]
```
sanbrock, line 27: not registered among possible axes

galleries/nxdata/plot_image.py
```diff
@@ -0,0 +1,51 @@
...
+  scan1:
+    @NX_class = "NXentry"
+    @default = "data"
+    data:
+      @NX_class = "NXdata"
+      @axes = ["x", "y"]
+      @signal = "z"
+      x: NX_FLOAT64[16]
+      y: NX_FLOAT64[30]
+      z: NX_FLOAT64[16,30]
```
PeterC-DLS, line 24:
```suggestion
      z: NX_FLOAT64[30,16]
```

PeterC-DLS, line 24:
```suggestion
      z: NX_FLOAT64[30,16]
```

galleries/nxdata/plot_image.py
```diff
@@ -0,0 +1,51 @@
...
+.. code::
+
+  @NX_class = "NXroot"
+  @default = "scan1"
+  scan1:
+    @NX_class = "NXentry"
+    @default = "data"
+    data:
+      @NX_class = "NXdata"
+      @axes = ["x", "y"]
```
PeterC-DLS, line 20:
```suggestion
      @axes = ["y", "x"]
```

galleries/nxdata/plot_image.py
```diff
@@ -0,0 +1,51 @@
...
+      x: NX_FLOAT64[16]
+      y: NX_FLOAT64[30]
+      z: NX_FLOAT64[16,30]
+
+Explanation:
+
+1. ``@axes`` has two values which corresponds to the signal rank of two.
+
+2. ``z`` is the default signal to be plotted versus ``x`` and ``y``.
+"""
```
PeterC-DLS, line 31: 3. Row-major storage ordering means image data is indexed as row/column pairs

galleries/nxdata/plot_fscan2d.py
woutdenolf: You mean `x_encoder`? 

That's exactly what I thought as well. Which lead to my changes in https://github.com/nexusformat/definitions/pull/1213 to be wrong as mentioned in https://github.com/nexusformat/definitions/issues/1381, the whole reason for the PR to begin with.

`@axes` gives the default axis to be used for each signal dimension. Hence this part in the new description

> Highlight consequences of these constraints:
> ...
> A list of all available axes is not provided directly. All strings in the axes attribute
> (excluding the “.” string) are axis field names. In addition the prefix of an attribute
> ending with the string “_indices” is also an axis field name.

galleries/nxdata/plot_image.py
woutdenolf: I don't understand what this means.

woutdenolf: Ok "row-major" means "C-order" or "the last dimension is the fastest dimension".

I'm trying to understand now what "indexed as row/column pairs" means.

Do you mean that `z: NX_FLOAT64[30,16]` refers to the fact that the z matrix has 30 rows and 16 columns? Or is the C-order aspect somehow implied by the definition?

If I think of `z` as a numpy array with `z.shape == (30 16)`, it does not matter whether it is C-order or F-order afaik.

woutdenolf: Instead of you suggestion (which is good but too complex for me at least) I added

3. ``z`` has 30 rows and 16 columns.

4. ``y`` spans the first dimension of ``z`` and ``x`` the second.

This comment made me also swap the dimensions of hist2d, hist2dmix and fscan2d to be consistent with image: y is the first dimension (i.e. the slow one assuming C-order) and x the second dimension (i.e. the fast on assuming C-order).

galleries/nxdata/plot_fscan2d.py
```diff
@@ -0,0 +1,95 @@
...
+      y_encoder: NX_FLOAT64[16]
+      x_set: NX_FLOAT64[6]
+      y_set: NX_FLOAT64[16]
+
+Explanation:
+
+1. ``@axes`` has two values which corresponds to the signal rank of two.
+
+2. ``x_set`` and ``y_set`` are the default axes which can be used by readers
+   that cannot handle multi-dimensional coordinated.
```
PeterC-DLS, line 41:
```suggestion
   that cannot handle multi-dimensional coordinates.
```

galleries/nxdata/plot_fscan2d.py
phyy-nx: Resolved in Telco.  @sanbrock pointed out the section "Highlight consequences of these constraints" which address this comment.

woutdenolf:
Summary of what this MR provides:

1. "@axes" does not provide a list of axes, it provides the default axis for each signal dimension. This is fixed.
2. The axes can have more than one value than the corresponding signal dimensions.
3. Restructure the description with an example gallery.

base_classes/NXdata.nxdl.xml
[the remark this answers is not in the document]
phyy-nx: @woutdenolf @prjemian Capture this in a new issue?  Then resolve this comment?

base_classes/NXdata.nxdl.xml
phyy-nx: From Telco: this could be resolved as a new flow chart that fully explores the possibilities of the general case of NXdata.  Consider the code snippet [below](https://github.com/nexusformat/definitions/pull/1396#issuecomment-2380113555) that is a general reader.  What would a diagram like the ones [here](https://manual.nexusformat.org/datarules.html#version-3) look like?  Maybe a new webpage for the NeXus developer?  @woutdenolf feel free to either: a) add something like this to this PR or b) move this to a new issue.

woutdenolf:
A technical detail: we see *orphan:* in the rendering. @tacaswell Do you have a workaround since https://github.com/sphinx-gallery/sphinx-gallery/issues/854 is not addressed?

base_classes/NXdata.nxdl.xml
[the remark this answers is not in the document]
phyy-nx: This PR is now under a vote.  We decided today that we'd like to get it merged in parallel with any further discussion on this thread.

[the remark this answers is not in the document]
woutdenolf: The current disposition seems to be that `@signal` does not have a default value. As @PeterC-DLS pointed out and confirmed by https://manual.nexusformat.org/datarules.html, there is ''older NeXus methods to find the default"

image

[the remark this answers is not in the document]
woutdenolf: The "older NeXus methods" are version 2 and version 1 while the current version is referred to as version 3. So reader needs to start the flowchart of version 3, which links to the chart of version 2 and version 1.

https://manual.nexusformat.org/_images/flowchart-v2-NXdata-signal.png

image

https://manual.nexusformat.org/_images/flowchart-v1-NXdata-signal.png

image

The NXdata docs only present version 3 and have a link that leads to the flow charts in the "@signal" attribute description:

https://hdf5.gitlab-pages.esrf.fr/nexus/nxdata_axes_fix/classes/base_classes/NXdata.html#nxdata-signal-attribute

I think it makes sense to only present version 3 as NXdata is already mindbogglingly complex. So the link in "@signal" is fine. However I don't now know about it so perhaps in the "Signals" section of the "description" we should have a link to https://manual.nexusformat.org/datarules.html#find-plottable-data.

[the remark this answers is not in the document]
woutdenolf: The current MR is not changing the way the older versions 2 and 1 are mentioned in the current NeXus docs:

- Data field "@signal" attribute https://hdf5.gitlab-pages.esrf.fr/nexus/nxdata_axes_fix/classes/base_classes/NXdata.html#nxdata-data-signal-attribute
- Axis field "@axis" attribute https://nxdata-axes-fix-hdf5-nexus-e16d3be7bef8684a1ad37775f8cbe25e8412.gitlab-pages.esrf.fr/classes/base_classes/NXdata.html#nxdata-axisname-axis-attribute
- And the link in the "@signal" attribute of the NXdata group itself https://hdf5.gitlab-pages.esrf.fr/nexus/nxdata_axes_fix/classes/base_classes/NXdata.html#nxdata-signal-attribute

So if we want improvement we can move the discussion to https://github.com/nexusformat/definitions/issues/1544. It does not interfere with the current MR imo.

base_classes/NXdata.nxdl.xml
woutdenolf: The original comment @PeterC-DLS is still unanswered though: 

> When AXISNAME_indices is the same as the indices of “AXISNAME” in the axes attribute, there is no need to provide AXISNAME_indices.

Does this apply only to 1D axes?

woutdenolf: I don't think the extra condition of this statement to only apply to 1D axes is needed.

If a writer is creating "@x_indices=[0,1]" with "@axes=["x", "x", "z"]" he can omit "x_indices" since it is equal to the indices if "x" in "@axes".

@PeterC-DLS did mention however that the indices if "x" in "@axes" can be either "[0, 1]" or "[1, 0]". So maybe we should add the 1D restriction.

woutdenolf:
Voters: consider the two open discussions

https://github.com/nexusformat/definitions/pull/1396#discussion_r1972917331

https://github.com/nexusformat/definitions/pull/1396#discussion_r1972933136

rayosborn (approved):
I don't think any issues should delay the vote, but at some point, the following should be changed.
"signals with more than one dimension: data could be 2D with axes x and y along each dimension."
In its current position in the documentation, this might confuse newcomers about whether 3- or higher-dimensional data is allowed. Later on, examples of 3D data are shown so it's clear the reference to 2D is just an example, but it's not obvious from the way it's phrased. Minor changes like this can be made later, I believe, without a vote, so I am approving this PR now.

phyy-nx:
The vote has passed!  Before this can be merged, the comments above need to be resolved.  Participants include @woutdenolf, @rayosborn, @prjemian, and @PeterC-DLS.  Please make sure any unresolved concerns are moved to new issues, since once this is merged they are likely to be forgotton!!  Otherwise @woutdenolf feel free to merge and indeed, thanks for all the work on this!

base_classes/NXdata.nxdl.xml
woutdenolf: We can resolve in another issue: https://github.com/nexusformat/definitions/issues/1549

woutdenolf:
I resolved the two remaining comments which can be addressed in https://github.com/nexusformat/definitions/issues/1544 and https://github.com/nexusformat/definitions/issues/1549

woutdenolf:
For reference: this got fixed somehow in the current builds of the docs.

The "orphan" tag used to be added but sphinx-gallery but this is no longer the case.

<details>
  <summary>New build (fixed)</summary>

  image
</details>

<details>
  <summary>Old build (with orphan tag)</summary>

  image
</details>

PR #1402, opened 2024-08-09
"term" should be all caps (indicates it is a flexible name)

### LINKED ISSUE #1401: What does it mean to have an `maxOccurs=unbounded` `nameType=specified`?
e.g.
https://github.com/nexusformat/definitions/blob/69f92a320f15effe29e11237fcf2cad6674f55c3/base_classes/NXparameters.nxdl.xml#L30-L34

This reads as - the field must be named `term`, but there can be several of them?

Similarly, there are fields with `maxOccurs=1` and `nameType=any`, e.g.
https://github.com/nexusformat/definitions/blob/69f92a320f15effe29e11237fcf2cad6674f55c3/base_classes/NXtransformations.nxdl.xml#L187

although I think I sort of see what is being gone for here - that the name is technically specified, because it's intended to be based off the name of another parameter. Except it seems more technically correct to say that there can be multiple instances of this field (otherwise, you seem to be saying "you can have many AXIS but only one of them can have an AXISNAME_end").

prjemian:
- close #1401

@ndevenish - Thanks for reporting this!

### DIFF
--- a/base_classes/NXparameters.nxdl.xml
+++ b/base_classes/NXparameters.nxdl.xml
@@ -27,7 +27,7 @@
     name="NXparameters" 
 	type="group" extends="NXobject">
     <doc>Container for parameters, usually used in processing or analysis.</doc>
-    <field name="term" minOccurs="0" maxOccurs="unbounded" type="NX_CHAR">
+    <field name="TERM" minOccurs="0" maxOccurs="unbounded" type="NX_CHAR">
         <!-- maxOccurs="unbounded" is intended but is not allowed by current syntax -->
         <doc>A parameter (also known as a term) that is used in or results from processing.</doc>
         <attribute name="units"/>
--- a/base_classes/NXparameters.nxdl.xml
+++ b/base_classes/NXparameters.nxdl.xml
@@ -27,7 +27,7 @@
     name="NXparameters" 
 	type="group" extends="NXobject">
     <doc>Container for parameters, usually used in processing or analysis.</doc>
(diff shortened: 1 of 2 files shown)


woutdenolf (approved):

woutdenolf (approved):

base_classes/NXparameters.nxdl.xml
```diff
@@ -27,7 +27,7 @@
     name="NXparameters" 
 	type="group" extends="NXobject">
     <doc>Container for parameters, usually used in processing or analysis.</doc>
-    <field name="term" minOccurs="0" maxOccurs="unbounded" type="NX_CHAR">
+    <field name="TERM" minOccurs="0" maxOccurs="unbounded" nameType="any">
```
woutdenolf, line 30: `type="NX_CHAR"` is removed which changes the definition?

woutdenolf: `nameType="any"` makes the capitalized `TERM` explicit. I did not know this. Otherwise from an NXDL pov the name "TERM" would be a fixed name.

Perhaps we should do something in the HTML rendering to make this explicit because I thought we just relied on all-caps to know whether it is a fixed name or not.

prjemian: The default data type for all NeXus fields is `NX_CHAR`, so nothing is changed.  It was not necessary to specify it here.

prjemian:
@woutdenolf Thanks for the review!

PR #1407, opened 2024-09-23
FAIRmat 2024: small additions to NXsource

lukaspie:
This PR makes some changes to `NXsource`, mostly to expand the range of where it can be used and to what detail the source can be described:
- class docstring: indicate that `NXsource` is for any kind of radiation source, not just for neutron or x-ray storage ring/facilities
- add new items to enum `types`: LED, Laser, etc. with the possibility to specify `type_other` through an attribute on `type`
  - this is a direct follow-up to the docstring change indicated above
- add new item to enum for `probe`: `photon`
  - this is for experiments where e.g. the X-ray energy is scanned across a wide range, such that it is neither just `x-ray` or `ultraviolet`. Think survey scans in XPS as an example.
- add new unnamed groups: `NXaperture`, `NXdeflector`, `NXlens_em` (see discussion below), `NXfabrication`
  - these are either used to describe the inner parts of the source or the way it was fabricated
  - note that NXfabrication is a new base class, but it was already approved with #1414 (see vote in https://github.com/nexusformat/definitions/issues/1451#issuecomment-2382523273)
- add new fields: wavelength, pulse_energy, peak_power, anode_material, filament_current, emission_current, gas_pressure,
previous_source
  - these are additional parameters that can be measured on different source types during operation (see discussion below)
  - previous source: this is to be used if multiple sources are placed right after another to describe for a secondary source             the instance(s) of NXsource from which a beam originated to reach this source.


base_classes/NXsource.nxdl.xml
```diff
@@ -56,12 +61,28 @@
...
 			<item value="UV Plasma Source" />
 			<item value="Metal Jet X-ray" />
-		</enumeration>
-	</field>
+            <item value="Laser"/>
+            <item value="Dye-Laser"/>
+            <item value="Broadband Tunable Light Source"/>
+            <item value="Halogen lamp"/>
+            <item value="LED"/>
+            <item value="Mercury Cadmium Telluride"/>
```
phyy-nx, line 69: Add Lamp

base_classes/NXsource.nxdl.xml
```diff
@@ -56,12 +61,28 @@
 			<item value="Ion Source" />
 			<item value="UV Plasma Source" />
 			<item value="Metal Jet X-ray" />
-		</enumeration>
-	</field>
+            <item value="Laser"/>
+            <item value="Dye-Laser"/>
+            <item value="Broadband Tunable Light Source"/>
+            <item value="Halogen lamp"/>
```
phyy-nx, line 67: Capitalize Lamp

phyy-nx, line 67:
```suggestion
            <item value="Halogen Lamp"/>
```

base_classes/NXsource.nxdl.xml
```diff
@@ -56,12 +61,28 @@
 			<item value="Ion Source" />
 			<item value="UV Plasma Source" />
 			<item value="Metal Jet X-ray" />
-		</enumeration>
-	</field>
+            <item value="Laser"/>
+            <item value="Dye-Laser"/>
```
phyy-nx, line 65:
```suggestion
            <item value="Dye Laser"/>
```

base_classes/NXsource.nxdl.xml
```diff
@@ -56,12 +61,28 @@
...
+            <item value="Broadband Tunable Light Source"/>
+            <item value="Halogen lamp"/>
+            <item value="LED"/>
+            <item value="Mercury Cadmium Telluride"/>
+            <item value="Deuterium Lamp"/>
+            <item value="Xenon Lamp"/>
+            <item value="Globar"/>
+            <item value="other"/>
```
phyy-nx, line 73:
```suggestion
            <item value="Other"/>
```

base_classes/NXsource.nxdl.xml
```diff
@@ -163,55 +184,118 @@
...
+	<field name="emission_current" type="NX_FLOAT" units="NX_CURRENT">
+		<doc>
+				Emission current of the generated beam.
+		</doc>
+	</field>
+	<field name="gas_pressure" type="NX_FLOAT" units="NX_PRESSURE">
+		<doc>
+				Gas pressure inside ionization source.
+		</doc>
+	</field>
```
phyy-nx, line 222: Move these into sub-classes (wavelength through gas_pressure)

phyy-nx:
Proposal: accept changes with additions requested.  Cannot merge until NXlens_em is reviewed.

base_classes/NXsource.nxdl.xml
```diff
@@ -56,12 +61,28 @@
...
+            <item value="Halogen lamp"/>
+            <item value="LED"/>
+            <item value="Mercury Cadmium Telluride"/>
+            <item value="Deuterium Lamp"/>
+            <item value="Xenon Lamp"/>
+            <item value="Globar"/>
+            <item value="other"/>
+        </enumeration>
+    </field>
+    <field name="type_other">
```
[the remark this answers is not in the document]
lukaspie: done

base_classes/NXsource.nxdl.xml
lukaspie: done

phyy-nx:
@sanbrock @lukaspie the above comment says 'Cannot merge until NXlens_em is reviewed.'.  What PR was that referring to?

Also there is an unresolved comment about moving some of this to subclasses.

Also, please add a description to the top of this PR.  Thanks!

base_classes/NXsource.nxdl.xml
lukaspie: IMO, there is no really good to move these into sub-classes. Some of these fields are for X-ray tubes, some for more ionization sources, some for pulsed source. `wavelength` on the other hand is certainly a concept for many different kinds of sources.

I could create 3 or 4 classes here (each inheriting) from `NXsource`, like `NXxray_tube` and `NXpulsed_source`, each of which takes one or two of these new concepts. But what about an pulsed X-ray source, that would be again the merger of the two.

To me, it makes more sense to keep them in the `NXsource` base class.

lukaspie:
> @sanbrock @lukaspie the above comment says 'Cannot merge until NXlens_em is reviewed.'. What PR was that referring to?

In `NXsource`, we are now making use of `NXlens_em`, which so far has been living in `contributed_definitions`. In order for this to be standardized, during the NIAC meeting the suggestion was that we move it to `base_classes`, because an officially accepted base class should not depend on a contributed one. However, this can also be done in a separate PR if you agree.

> Also there is an unresolved comment about moving some of this to subclasses.

Here, I am a bit stumped as to what the idea was. I discussed it a bit above in the comment itself.

> Also, please add a description to the top of this PR. Thanks!

Done.

base_classes/NXsource.nxdl.xml
phyy-nx: I'm ok with your reasoning here but technically it's not what we voted on.  Let's review how to proceed on Monday.

phyy-nx:
Comments from Telco:
- Remove type_other and Other from the type enumeration
- Create a new PR that adds to enumeration an attribute `open=False`, which would allow the enumeration to allow values not in the list.  Note that accepting that change would require adapting sphinx to render it nicely, and affect validators.  They should not produce errors but could produce a warning.  That PR would need a NIAC vote.
- Don't go with subclasses for wavelength, gas_pressure, etc.  Not technically what we voted on, but the Telco agrees that we'd have to do mixed inheritance which isn't technically possible given our new inheritance model.
- Remove NXlens_em.  When NXlens_em is approved to move out of contributed definitions, in the same PR that implements that, NXlens_em can be re-added to NXsource

Once the above is done, a simple review and merge can be done.  No vote needed.

base_classes/NXsource.nxdl.xml
phyy-nx: Resolved in the telco to keep as is

phyy-nx:
Hi all, as discussed in the Telco, now that the above points have been addressed (thanks @lukaspie), we can now move this PR to an online vote.  NIAC committee members please vote on this PR using emojis.  👍 for yes, 👎  for no, anything else (for example 👀) to abstain.  We needs 14 votes to hit quorum so please review and vote!

phyy-nx:
Whoops, this one didn't need a vote.  Comment deleted

phyy-nx (approved):

lukaspie:
Thanks @phyy-nx for resolving and merging!

PR #1410, opened 2024-09-23
Fairmat 2024: add target attribute to NXdata

lukaspie:
This PR introduces a new concept for `NXdata`. Essentially, the idea is here that a new attribute is added to `NXdata` to indicate the **origin** of the data. The whole idea comes from discussion with instrument manufacturers who would like to combine multiple detector readings into the top-level NXdata group (e.g., for spin-resolved photoemission measurements) and would like to keep track for both them and the user where this data originated.

In the case of a link, the value of this attribute would be the target of the link. That is, if an `NXdata` group in `entry/data` is linked from an `NXdata` group on the detector, the value of this attribute would be `entry/dectector/detector_data`. Similarly, both for the `AXISNAME` and `DATA` fields, such an attribute is added to indicate directly where this axis/data field is originating from.

This new attribute serves multiple purposes:
1) it provides human-readable information about links that would otherwise need to be extracted from whatever API one was using to write/read the file.
2) As @rayosborn correctly pointed out below, if you use the nexusformat API, using the `nxlink`, attribute it is easy to resolve to the linked object (however, it is less easy using h5py or any other HDF5 library). In any case, also with the nexusformat API, this resolving however works only if e.g. the `NXdata` and the referenced data group are in the same file (as far as I understand, please correct me if I'm wrong). With the target attribute, there is at least the option to keep track from where this data originally came from, even if the link is not directly resolvable anymore.
3) Another use case would be that you do not actually want to make a link, but rather a _copy_ of the data. An example would be that you create a field somewhere that has multiple other attributes, but you would really only like to bring the field values and not the attributes to your `NXdata/AXISNAME`. With a link, that would be impossible because in HDF5, the two concepts would then refer to the same object. Thus, one would rather like to make a copy the field values, but still indicate with this new attribute where the data was originally from.
4) In addition to linking, this new attribute could also be used for a string notation describing e.g. "the axis xy comes from here and there". This was sugggested by the manufacturers at some point, but I would (at least for now) rather limit it to the cases that I outlined above (where at least a target also exists somewhere else).

Please note that this attribute was originally called `@reference`, with the idea of _referring back_ to other data. It was changed to `@target` upon request by the NIAC in the NIAC meeting 2024, but in my opinion, these concepts are slightly different and a different name would fit that attribute better.


base_classes/NXdata.nxdl.xml
```diff
@@ -309,6 +309,20 @@
...
+             Points to the path of a field defining the data to which the `DATA` group refers.
+
+             This concept allows to link the data to a respective field in the NeXus hierarchy, thereby
+             defining the physical quantity it represents.
+
+             Example:
+                 If the data corresponds to a readout of a detector, ``@reference`` links
+                 to that detectors data:
+
+                   @reference: '/entry/instrument/detector/data' for a 2D detector
```
benajamin, line 323: can remove "2D"

benajamin:
This is fine for a path in the same file, but what if the NXdata and the referenced data field are not in the same file?

base_classes/NXdata.nxdl.xml
lukaspie: Agreed, I changed it

base_classes/NXdata.nxdl.xml
```diff
@@ -309,6 +309,20 @@
 
 		</doc>
 	</attribute>
+    <attribute name="reference">
+        <doc>
+             Points to the path of a field defining the data to which the `DATA` group refers.
+
+             This concept allows to link the data to a respective field in the NeXus hierarchy, thereby
```
benajamin, line 316: suggest rewording to avoid using the word "link":
"This attribute indicates the origin of the data, thereby providing an association to further metadata that define the physical quantity that the data represents"

prjemian:
Compare with the existing `target` attribute. This is used with NeXus links.

phyy-nx:
The NIAC asks if it's possible to use the target attribute instead of adding a new attribute `reference`.  This could affect the target documentation because in the past it's only been used for links, but here it's desired to use the target attribute without links.  This could also affect nexusformat API.

prjemian:
I didn't see why it could not be used. But the content and structure must
remain unchanged.

lukaspie:
Needs #1491

rayosborn:
I think this PR is premature. I believe that at the recent NIAC, we agreed to look into whether the same thing could be achieved with the `target` attribute. That work has not been done. At the very least, this should wait until the next Telco.

rayosborn:
As a quick follow-up to my last comment, I see that the proposal has now been changed to use 'target' instead of 'reference'. I still think this needs some further discussion, because this is a major structural change to the format. 

Personally, I believe that the PR should be resubmitted, with a detailed introduction explaining the rationale for all these changes and what the consequences are. I don't believe it's good practice to submit a PR with absolutely no explanation because it makes it impossible for future readers to know the context of the request and it makes it very difficult for current reviewers to understand what is proposed and why. I know we discussed it in the NIAC but I didn't really understand why it was needed then and I understand it less now.

For example, the current docstrings say the target "attribute indicates the origin of the data, thereby providing an association to further metadata that define the physical quantity that the data represents." 

What does "origin" mean here if the data is not actually linked? Can the field with the `target` attribute have no corresponding field (at any time) in the target group? Can it be present but with different contents? I seem to recall (perhaps incorrectly) that part of the issue was the need to create the NXdata group before the target exists in the NXdetector group. This can be achieved now without making any changes. In HDF5 and the nexusformat API, you can create a soft link before the target exists.

In the following, the soft link is created before the target object. The `nxtarget` attribute shows its path, but the `nxlink` attribute, which resolves to the linked object, shows an invalid path since it doesn't yet exist.  Once it's created, the `nxlink` attribute returns the linked object.
```
>>> root['entry/data/data'] = NXlink(target='entry/instrument/detector/data', soft=True)
>>> root['entry/data/data'].nxtarget
'/entry/instrument/detector/data'
>>> root['entry/data/data'].nxlink
NeXusError: Invalid path
>>> root['entry/instrument/detector/data'] = np.array(...)
>>> root['entry/data/data'].nxlink
NXfield('data')
>>> root['entry/data/data'].nxlink.nxpath
'/entry/instrument/detector/data'
```
Does that accomplish what you need? It's true that, if you use h5py instead of the nexusformat API, you might want to manually add the `target` attribute to the linked object after it has been created, but you could use other h5py functions to identify the target path.

The point of this discussion is that you can probably achieve what you need without a major structural change to the format, but we need to know the precise use case you have in mind to know how. I am willing to amend the nexusformat API to facilitate your use case.

lukaspie:
> As a quick follow-up to my last comment, I see that the proposal has now been changed to use 'target' instead of 'reference'. I still think this needs some further discussion, because this is a major structural change to the format.
> 
> Personally, I believe that the PR should be resubmitted, with a detailed introduction explaining the rationale for all these changes and what the consequences are. I don't believe it's good practice to submit a PR with absolutely no explanation because it makes it impossible for future readers to know the context of the request and it makes it very difficult for current reviewers to understand what is proposed and why. I know we discussed it in the NIAC but I didn't really understand why it was needed then and I understand it less now.
> 
> For example, the current docstrings say the target "attribute indicates the origin of the data, thereby providing an association to further metadata that define the physical quantity that the data represents."
> 
> What does "origin" mean here if the data is not actually linked? Can the field with the `target` attribute have no corresponding field (at any time) in the target group? Can it be present but with different contents? I seem to recall (perhaps incorrectly) that part of the issue was the need to create the NXdata group before the target exists in the NXdetector group. This can be achieved now without making any changes. In HDF5 and the nexusformat API, you can create a soft link before the target exists.
> 
> In the following, the soft link is created before the target object. The `nxtarget` attribute shows its path, but the `nxlink` attribute, which resolves to the linked object, shows an invalid path since it doesn't yet exist. Once it's created, the `nxlink` attribute returns the linked object.
> 
> ```
> >>> root['entry/data/data'] = NXlink(target='entry/instrument/detector/data', soft=True)
> >>> root['entry/data/data'].nxtarget
> '/entry/instrument/detector/data'
> >>> root['entry/data/data'].nxlink
> NeXusError: Invalid path
> >>> root['entry/instrument/detector/data'] = np.array(...)
> >>> root['entry/data/data'].nxlink
> NXfield('data')
> >>> root['entry/data/data'].nxlink.nxpath
> '/entry/instrument/detector/data'
> ```
> 
> Does that accomplish what you need? It's true that, if you use h5py instead of the nexusformat API, you might want to manually add the `target` attribute to the linked object after it has been created, but you could use other h5py functions to identify the target path.
> 
> The point of this discussion is that you can probably achieve what you need without a major structural change to the format, but we need to know the precise use case you have in mind to know how. I am willing to amend the nexusformat API to facilitate your use case.

@rayosborn thanks for the pointers in the nexusformat API. This idea is not necessarily about the order when links are created, but also for cases where not necessarily a link in HDf5 is created, but one would still like to say where the data originated from. I outlined multiple usage above, hope that helps to clarify the purpose of this new attribute. I/we would be happy to further discuss this.

base_classes/NXdata.nxdl.xml
```diff
@@ -309,6 +309,20 @@
...
+			Points to the path of a field defining the data to which the `DATA` group refers.
+
+			This attribute indicates the origin of the data, thereby providing an association to
+			further metadata that define the physical quantity that the data represents.
+
+			Example:
+			    If the data corresponds to a readout of a detector, ``@target`` points
+				to that detectors data:
+
+				  @target: '/entry/instrument/detector/data' for a detector
```
phyy-nx, line 323: Typo: detectors --> detector's

Also, use a different example that doesn't use detector twice.

base_classes/NXdata.nxdl.xml
```diff
@@ -309,6 +309,20 @@
 
 		</doc>
 	</attribute>
+	<attribute name="target">
+        <doc>
+			Points to the path of a field defining the data to which the `DATA` group refers.
```
phyy-nx, line 314: field --> group

phyy-nx (changes requested):
When these changes are done, ok to move to online vote, as discussed by NIAC meeting Dec 18 2024

phyy-nx:
NIAC telco:
- It appears this PR is redundant with existing functionality, namely that any group or field can be a link, and the [specification](https://manual.nexusformat.org/design.html#design-links) notes that if so, add a target attribute.
- However, it is unclear to the user that this functionality exists.  So the question is, how to make it either more clear or more explicit
- One proposal is to add `@target` to NXobject, just to call out this feature, but there are two problems: 1) it's only applicable for links, and 2), `@target` can also be used for fields

Charge the NIAC: how to resolve these problems?

base_classes/NXdata.nxdl.xml
phyy-nx: 👍

base_classes/NXdata.nxdl.xml
```diff
@@ -309,6 +309,29 @@
 
 		</doc>
 	</attribute>
+	<attribute name="target">
+        <doc>
+			Points to the path of a group defining the data to which this NXdata group refers.
```
prjemian, line 314:
```suggestion
			Names the HDF5 address of the group defining the data to which this NXdata group refers.
```

phyy-nx:
From Telco: some of the confusion is because the idea "a link is a field" is not written down anywhere.  Could adding such a sentence or description to https://manual.nexusformat.org/design.html#design-links help clarify this issue?  Could the xsd formally describe this relationship?  Right now link type and field type are separate entities.  But link type could be a subtype of field type.

prjemian:
A link is a field **or** a group.

prjemian:
From https://github.com/BCDA-APS/bluesky_training/issues/223:

image

phyy-nx:
Telco: change nxdl.xsd to add target to field and group, not just link.  Then a small note could be added to NXdata reminding the user of this feature.  This is a change that allows provenance everywhere, not just NXdata, while also minimizing the word count in NXdata.

rayosborn:
> 2. What is less clear to me is for the `@target` atttribute on the NXdata `group`. We wanted to use this for referencing a whole NXdata `group` somewhere else, e.g. on the detector. Is this also possible with the existing `@target` atttribute?
 
The `target` attribute obviously refers to the parent of whatever object is linked, whether it's a group or field. It cannot reference the group containing that parent object directly, but that is easily determined from its path. In the `nexusformat` package, you would do the following:
```
>>> print(chopper['entry/data'].tree)
data:NXdata
  @axes = ['polar_angle', 'time_of_flight']
  @signal = 'data'
  data = int32(148x750)
    @long_name = 'Neutron Counts'
    @units = 'counts'
  polar_angle -> /entry/instrument/detector/polar_angle
  time_of_flight -> /entry/instrument/detector/time_of_flight
>>> print(chopper['entry/data/time_of_flight'].nxlink.nxgroup.tree)
detector:NXdetector
  distance = float32(148)
    @units = 'm'
  polar_angle = float32(148)
    @target = '/entry/instrument/detector/polar_angle'
    @units = 'degrees'
  time_of_flight = float32(751)
    @target = '/entry/instrument/detector/time_of_flight'
    @units = 'microseconds'
```
The `nxlink` attribute just resolves to the parent object while the `nxgroup` attribute gives the group it is in. In one line, it is possible to access the whole `NXdetector` group, allowing , *e.g.*, the detector distances to be used to convert the time-of-flight to wavelength. When it’s this easy, adding a reference directly to the `NXdetector` group becomes redundant.

> * **Data Copying**: As stated above, we also want to use this attribute (whatever it's called) if you do not actually want to make a link, but rather a **copy** of the data. An example would be that you create a field somewhere that has multiple other attributes (like a field in `NXinstrument` with many attributes), but you would really only like to bring the field values and not the attributes to your `NXdata/AXISNAME`. With a link, that would be impossible because in HDF5, the two concepts would then refer to the same object. Thus, one would rather like to make a copy the field values, but still indicate with this new attribute where the data was originally from.

You can always copy values from one field to another, but you have to decide whether you want your copy updated when the parent is updated. This is not easily possible if they refer to different objects. Otherwise, I’m not sure why you need to know where the copy came from. It might even be dangerous if the parent and copy diverge in their contents.

> * **Origin Tracking**: We argue, that in some cases the community using NeXus would like to know where the data was originated from. Therefore, we need an attribute to be able to register for any group or dataset this that this object was **derived** from somewhere else (another good way to think about this is as describing the **data origin**). Please note the difference, that we do not assume that the data object here would be the same as the referenced one. For example, the one here may contain only the relevant section what a monitor was measuring during the experiment, or the one here is converted to a different unit compared to the referenced one. This is a big difference compared to a simple (HDF5) link (or even a soft link), for which the `@target` attribute is designed. Hence, additionally to the data (which is either a new dataset, a hard/soft/external link, or even a virtual dataset which one it is just an hdf5 implementation details when NeXus is used on top of hdf5) we would like to allow attaching an attribute telling where it is coming from.

Again, I’m not really sure why this is necessary. Perhaps there is some value to a scientific historian, who wants to understand how the file was constructed, but I don’t see how this has any value to someone writing analysis code.

> So fundamentally, the ideas we were bringing in here are about **following the provenance of the data to their origin**. If all of these use cases are covered by the `@target` atttribute (both for fields and groups in `NXdata`), I would be happy to drop this PR and simply use recommended `@target` atttributes for all fields in `NXmpes/NXdata` and for the whole `NXmpes/NXdata` group. However, if the `@target` is explicitly **not** designed for any of these cases, I still think a new attribute like `@reference` or `@origin` should be added, as it was originally in this PR.

If you want the copy to be mutable, then I’m not sure that its provenance has any meaning. Once you have changed the values in a copy, then knowing where it came from originally is not of much use and might cause confusion.

If you have a specific example of a required operation that is not possible with the current design, please post it, along with code (or at least pseudo-code) showing how the change enables it. Examples are worth a thousand words. At the moment, these all seem hypothetical.

lukaspie:
I am closing this PR now as it will clearly not gain a positive vote from NIAC. Will aim to find a resolution as outlined by @phyy-nx in https://github.com/nexusformat/definitions/pull/1410#issuecomment-2802167450.

phyy-nx:
@rayosborn no worries, we chatted a bunch about this and the alternate proposal should be ok once it's ready.

lukaspie:
> Telco: change nxdl.xsd to add target to field and group, not just link. Then a small note could be added to NXdata reminding the user of this feature. This is a change that allows provenance everywhere, not just NXdata, while also minimizing the word count in NXdata.

@phyy-nx Note that `@target` for links is different to what we proposed in this PR initially because it is an XML attribute, _not_ a NeXus attribute of `attributeType` (see e.g. here: https://github.com/nexusformat/definitions/blob/main/applications/NXreftof.nxdl.xml#L122 that it is used within the XML tag of `field`). If I simply copy the definition for target to `fieldType` and `groupType`, that means I can write in an application definition `<fieldname="my_field" target="some_field_elsewhere"></field>`. But that does _not_ help me to define a recommended NeXus attribute like this:
```
<field name="energy" type="NX_NUMBER" recommended="true" units="NX_ENERGY">
    <attribute name="target" type="NX_CHAR" recommended="true">
        <doc>
            Some documentation what @target means here.
        </doc>
    </attribute>
</field>
```
against which I can validate.

So I don't see how changing the nxdl.xsd would resolve the latter case. Is there any way to define in the XSD that fields can always have an attribute (i.e, of `attributeType`, not an XML attribute) that has `name="target"`? Does that exist for any other similar concept (like `@variant` for example)?

If not, I don't see how can I simply remove these attributes from NXmpes as [you suggested](https://github.com/nexusformat/definitions/pull/1424#discussion_r2043117398). If I write an application that reads NeXus files and come across a field `energy` with an attribute `@target`, how would I know if that is documented anywhere if I don't define it in either the .nxdl.xml files or in the nxdl.xsd? Do I just need to hardcode this into my data consumer? There will be no documentation that I can read out and attach to this attribute. Currently, I can just read out the documentation from NXmpes.nxdl.xml and attach it to this concept. See below the screenshot where we do exactly this for our research data management system NOMAD (albeit with some renaming).

image

How can I define what this `@target` attribute is if I can't define it in any of the NXDL files or the nxdl.xsd?

lukaspie:
Just to finally explain our use case in #1424 again:

For a given `AXISNAME` like `angular1`, we want to record which angle (that we typically define in the transformations, but could also do elsewhere) we are referring to with `angular1`.

We have 6 angles describing each electrons (3 angles for the rotations of the sample and 3 for the rotations of the detector) and then we write to angles `angular0` and `angular1` in `NXdata`. Essentially, with the `@reference` attribute, we wanted to say that `angula0` corresponds to e.g. the tilt of the sample.  This can, but doesn't explicitly have to, be a link using `@target`, but basically just stating that this  `angula0` references the `sample_tilt_angle`.

For me, it would be fine to keep `@reference` in the application definition (not in the base class) as suggested in https://github.com/nexusformat/definitions/pull/1424#discussion_r2044247767

rayosborn:
If I understand your example, you have reasons for requiring the axes in the NXdata group to have specific names, 'angular0', _etc_., but want to find out what angles they refer to by referencing a field in another group with a more specific name, _e.g_., 'sample_tilt_angle'. 

This is possible with the current standard. Although NeXus links have to refer to the same object, they don't have to have the same name. This is true for both hard and soft links.

The following is valid:
```
>>> print(root.tree)
root:NXroot
  entry:NXentry
    data:NXdata
      parameter -> /entry/sample/temperature
    sample:NXsample
      temperature = 40.0
```
Adding a reference that is not actually a link runs the risk of the values of the two fields being different. When calculating `kx`, do you use the values stored in `angular0` or the values stored in `sample_tilt_angle`? They are not guaranteed to be the same. If they have to be the same, then I would recommend using a link.

lukaspie:
> If I understand your example, you have reasons for requiring the axes in the NXdata group to have specific names, 'angular0', _etc_., but want to find out what angles they refer to by referencing a field in another group with a more specific name, _e.g_., 'sample_tilt_angle'.
> 
> This is possible with the current standard. Although NeXus links have to refer to the same object, they don't have to have the same name. This is true for both hard and soft links.
> 
> The following is valid:
> 
> ```
> >>> print(root.tree)
> root:NXroot
>   entry:NXentry
>     data:NXdata
>       parameter -> /entry/sample/temperature
>     sample:NXsample
>       temperature = 40.0
> ```

This is obviously possible, but not what we want to do.
> 
> Adding a reference that is not actually a link runs the risk of the values of the two fields being different. When calculating `kx`, do you use the values stored in `angular0` or the values stored in `sample_tilt_angle`? They are not guaranteed to be the same. If they have to be the same, then I would recommend using a link.

In order to calculate `kx`, you need the values stored in `angular0`, the information about which angle these values represent and the values stored in `sample_tilt_angle` (which can be different from the values stored in  `angular0`). In case they are the same, we can use a link (possibly with `@target`, but this is currently also not documented for fields in the NXDL files), if they are not, we currently don't have a way of referencing `sample_tilt_angle`. I am cc-ing @rettigl here since this is basically his use case.

rettigl:
Here is one use case, where we read this attribute to find out which coordinate to map a certain NXdata axes to:
https://github.com/FAIRmat-NFDI/arpes/blob/dd85c41407a676e84c94dedee45cc4a42cc268b3/src/arpes/endstations/plugin/nexus.py#L126
As mentioned, for our purpose it would be sufficient to just add these to the named axes in NXmpes, rather than NXdata.

rettigl:
Additionally, if we would rely on links, could we ensure that the respective axes are indeed links (but without specifying the link target ahead of time) in the application definition?

rayosborn:
Links can be added to the application definitions. Here is an example from NXfluo:
```
<group type="NXdata" name="data">
    <link name="energy"
        target="/entry/instrument/fluorescence/energy">
    </link>
    <link name="data" target="/entry/instrument/fluorescence/data">
    </link>
</group>
```

lukaspie:
> @rettigl, Links can be added to the application definitions. Here is an example from NXfluo:
> 
> ```
> <group type="NXdata" name="data">
>     <link name="energy"
>         target="/entry/instrument/fluorescence/energy">
>     </link>
>     <link name="data" target="/entry/instrument/fluorescence/data">
>     </link>
> </group>
> ```

Again, we don't want to define the target beforehand (as stated by @rettigl above) because we don't know it (it depends on the experiment). The use of `target` is [required ](https://github.com/nexusformat/definitions/blob/main/nxdl.xsd#L1082) for `linkType`, so we cannot use it here.

Issue #1043, opened 2022-04-04
Clarify experiment_identifier/collection_identifier/entry_identifier


woutdenolf:
I'm looking for a destination for the following metadata associated to a data collection:

- synchrotron: "ESRF"
- beamline: "ID31"
- proposal id, defined in the scope of the ESRF: "hg-124"
- data collection id, defined in the scope of the proposal: "S_220323_00006_0001"
- sample ID, defined in the user scope: "S-220323-00006"
- sample UUID, defined in the user scope: "4590be84-3493-4bd2-91fe-4cf39cfcf71f"
- name of the technique(s): ["x-ray powder diffraction"]
- command: "fscan 0 10 100 0.1"

I'm currently thinking about this but I'm not sure I'm using the fields correctly:

```
# ./S_220323_00006_0001.h5

10.1: NXentry  # this is 1 scan in the file S_220323_00006_0001.h5

    experiment_identifier = "hg-124"
    entry_identifier =  ["x-ray powder diffraction"]
    collection_identifier = "S_220323_00006_0001"

    title = "fscan 0 10 100 0.1"

    instrument: NXinstrument
        name = "ESRF-ID31"
            @short_name = "ID31"
    
    sample: NXsample
        name = "S-220323-00006"
        uuid = "4590be84-3493-4bd2-91fe-4cf39cfcf71f"
```

I'm especially confused about `experiment_identifier`, `entry_identifier` and `collection_identifier`. Could someone clarify those? What I have is a proposal name "hg-124", a data collection name "S_220323_00006_0001" and and the techniques used ["x-ray powder diffraction"].

woutdenolf:
So this basically requires a better description on the `experiment_identifier`, `entry_identifier` and `collection_identifier` fields in NXentry. @prjemian @FreddieAkeroyd From the history it seems both of you have worked on this.

prjemian:
`collection_identifier` is a group of files (or group of database records, for the [Bluesky framework](https://blueskyproject.io)) of which this data is a part.

Sometimes that's a parent folder, could also be a single SPEC data file or set of SPEC files, or a collection in mongodb (as in bluesky).

prjemian:
`entry_identifier` is a placholder for the identification _of this data_ provided by the facility.

prjemian:
`experiment_identifier` and `entry_identifier` seem identical but (probably) provide a distinction to some facilities.

prjemian:
Here's my suggestion

```
[...]

PR #1412, opened 2024-09-23
Fairmat 2024: new fields for experiment description in NXentry and NXsubentry

### LINKED ISSUE #1450: NXentry

### LINKED ISSUE #1460: NXsubentry

lukaspie:
Here, several new fields are added to `NXentry`, to give a more detailed of where (`experiment_{location,facility,institution}` ) and when (`experiment_{start_date,end_date}`) the experiment was performed.


woutdenolf:
Do we know what these are identifiers of? There is an issue about it: https://github.com/nexusformat/definitions/issues/1043

phyy-nx:
Instead of this, add PID to NXobject as an attribute.

woutdenolf:
Will a group ever need more than one PID?

lukaspie:
Note that previous versions of this PR contained changes to the `*_identifier` fields in `NXentry` to make use of the then-proposed new base class `NXidentifier`. Currently, all discussion of identifiers has been removed.

@phyy-nx @sanbrock since there has anyway been no vote, I would suggest we wait with this PR until the discussion around #1486 has been settled and then make the necessary changes here on top afterwards.

phyy-nx:
@lukaspie since #1486 has been merged is this one ready for discussion again?

lukaspie:
> @lukaspie since #1486 has been merged is this one ready for discussion again?

We were thinking about replacing the existing identifiers with fields starting with `identifierNAME`. For example, adding an `identifier_entry` and deprecating `entry_identifier`. With the idea that _all_ identifiers in NeXus should follow this general idea of starting with `identifier...`. I haven't implemented this yet though. Would you think that this is a good way forward? If so, I can implement it and this PR would be ready to be discussed. All other changes are ready anyway.

phyy-nx:
Seems reasonable to me.

lukaspie:
> Seems reasonable to me.

I made the necessary changes and deprecated the `..._identifier` fields. Probably warrants some discussion whether all of these should be kept in the new `identifier_...` format, but this PR should be ready to be discussed now.

phyy-nx:
From Telco 08/11/25: NXentry@experiment_identifier seems baked deep into the DNA of NeXus, so deprecating it in favor of NXobject@identifierNAME could cause some problems.  Would welcome feedback from the NIAC here.

phyy-nx:
Alternatively, we could change NXobject@identiferNAME to NXobject@PREFIXidentifierSUFFIX, which would allow the use of experiment_identifer as is.  But that would lock away the string 'identifer' entirely, such that it could never be used by a user.

phyy-nx:
From Telco 08/25/25:

Two proposals:

1.  Change `NXobject/identifierNAME` to `NXobject/PREFIXidentifierSUFFIX`.
2. Add notes to `NXentry/XXX_identifier` that these should be considered identifiers as described in `NXobject/identifierNAME`

Regardless, we shouldn't deprecate `NXentry/XXX_identifier`

Pros/cons for 1.: simpler but locks away the word 'identifer' forever
Pros/cons for 2.: no functional change but complicates validation

Can I get a straw poll here on the preferred method?  Vote :one: or :two:.  Poll closes in a few days!

phyy-nx:
Moar votes on the straw poll pls?  :)

rayosborn:
Does the new `NAMEidentifier` partial name permit the use of an underscore? Or would `NAME_identifier` be better? I think the NeXus convention has been to split portmanteau names with underscores, so I was always a little uncomfortable with `identifierNAME` but went along because I think I was late to the discussion.

mkuehbach:
RE: @phyy-nx @sanbrock 
Wrt to these last parts of the discussions on this PR thread, discussed at the 2025/09/16 telco:

* https://github.com/nexusformat/definitions/blob/23b2ffd86d50b1eee89e2e0ffcda6b2fa6dcd830/nxdl.xsd#L187 states that "_" is not replaceable hence NAME_identifier thus forcing folks to go with a leading "_" always even if the "NAME" gets dropped, I cannot vote for it but I am against it, does any concept's start with "_" currently? So, if that change is required, "NAMEidentifier" should be it. Wrt to when to add underscores the documentation is clear  [https://manual.nexusformat.org/applying-nexus.html](separate multi worded names with underscores).
* Assuming that the straw poll above turns out such that it supports to convert the above request into a proper PR changing NX{em,apm,object} and alike: Is then one PR sufficient to change all FAIRmat appdefs and does this need a vote?

mkuehbach:
https://github.com/nexusformat/definitions/blob/41d8157e740610355071c2d0c6a59900a1fb475f/nxdl.xsd#L186
"(empty string allowed)" means "NAME_identifier" allows for "_identifier".

phyy-nx:
Ok, this PR is now ready for a vote.  Please vote using emojis, 👍 for yes, 👎 for now, and anything else e.g. 👀 for abstain.  Voting will close in 2 weeks.  Thanks!

benajamin:
I have a problem with the "experiment_location" field. The doc string says: "City and country where the experiment took place", which means that 2 pieces of information are intended for a single field. I would assume that it is intended to be a single comma separated list of "city, country" but we should not be in the business of making assumptions!

A separate issue is why these fields are being placed in NXentry, rather than NXroot? It does not make sense to me for such things to vary between different NXentry's in the one file.

mkuehbach:
On the first point I suggest to make NXresearch_project(NXnote) there one could have single fields city, country, geolocation, like for NXuser. Then in NXentry one would hook a group research_project but agree but yay maybe better to move this idea to an issue to the NXDL2026 milestone and instead now add city, country this is already good, geolocation I find better but there are different use cases with different location precision demands.

On the issue with assuming that each NXentry in a file was generated at the same location. Admitted this is a frequent case where multiple methods are combined at the same beamline or lab but what if I use a NeXus file to bundle such information for an interdisciplinary team where the sample was physically shipped and indeed the measurements or computer simulations/analyses where done in different places. Pinning that information hard into the root does no longer cleanly support such use case.

benajamin:
I don't think we need to go so far as to completely change the field into an NXresearch_project and would rather suggest adding to the doc string that the value of the field should be a single string comprised of the city, a comma and then the country.

I do agree that someone might want to append an NXentry to an existing file after moving an experiment to a new location (perhaps for a complementary technique), thus making a location described in the root inaccurate. However, this would be considered by many to be poor practice since data files should not be altered after a measurement is concluded. In the given example, better practice would be to have separate NeXus files for the measurements at different locations, and if a combined file is desired, then a new meta-file should be created that links to the relevant data files. I don't see any good reason why such a meta-file should include location information.

lukaspie:
@rettigl I think you initiated these `experiment_*` fields at some point? Would this still be something that you would want to standardize?

Tommaso-Pincelli:
Hi! This is Tommaso, for who doesn't know me, I worked on building the basic structure of NXmpes with Laurenz. I added the experiment_* fields for two reasons:

1. to specifically standardize the description of facility, beamlines and groups for data accessibility. It is important that the user knows explicitly in which facility, part of the world and time period the data was acquired.
2. it can be helpful also for the functionality of modular loaders and parsers at a more granular level than an appdef.

Of course this can be at a lower level into NXinstrument and NXsource, but for the moment these base classes do not impose standards on the geolocation and beamline identifiers, so they would need to be explicitly defined there too. 

At the time I just thought that having them there would make logical sense and provide high-level information if a NXroot contains multiple NXentry fields coming from different experiments. 

So I would say: yes, it makes sense to standardize, for example using the beautiful ideas put forward by @paulmillar, and I would still think that it makes sense to have them (at least linked) into NXentry, but I am open to discussion on this.

paulmillar:
Hi @Tommaso-Pincelli ,

First, thanks for all the effort you've already invested into NeXus and this pull request: it's very much appreciated.

Second, I support your desire to improve NeXus, particularly in improving information about the facility.

As an aside, I'm working with a small team looking into minting DOIs for beamlines.  Building on the work from FAIRmat, we plan to embed these DOIs in NeXus files.  We also plan to introduce information about the beamline (not available through DataCite metadata schema) using JSON-LD.  Therefore, it would be interesting to learn what information about a beamline (or its facility) you would find useful.

Specifically for NeXus, it would be helpful if you could check the current definitions of NXsource and NXinstrument classes, to see where the current definitions don't support the information you need, or where the information is too vaguely specified to be useful.

PR #1413, opened 2024-09-23
Fairmat 2024: several new base classes in NXsample and NXsample_component

lukaspie:
~~This PR is aimed at a more detailed description of the components of a sample. It introduces the following new base classes:
`NXsample_component_set`, `NXsubstance`, `NXunit_cell`, `NXsingle_crystal`, `NXrotation_set`~~

~~- `NXsample_component_set`: this is meant as a wrapper that takes in all the sub sample components (i.e. instances of `NXsample_component`) and their relationships (mass fractions, volume fractions, etc.)~~
~~- `NXsubstance`: this is meant to describe a chemical substance. Examples can be single chemical elements, chemical compunds, or alloys (e.g. CuO, C,). It is envisioned that if such a substance is crystalline, it could have an instance of~~
~~- `NXunit_cell`: Description of a unit cell, i.e., the crystal structure of a single thermodynamic phase. This was can be used within `NXsample_(component)` and/or in~~
~~- `NXsingle_crystal`: this is for describing a single crystalline material/phase within another material. It makes use of the new base_class  `NXrotation_set`, which provides a rigorous way of describing rotations, orientations, and disorientations.~~

~~Eventually, it is envisioned that each sample can contain one or more instances `NXsample_component_set`, each of which has one or more instances of `NXsample_component`. This component is either not mixed (and thus can contain an instance of `NXsubstance`) or it by itself contains one or more instances `NXsample_component_set` with multiple instances of `NXsample_component`. The image provides a useful visualization of this nesting.~~

This PR was originally aimed at a more detailed description of the components of a sample (see crossed out text above). However, we realized in the meantime that this is not yet stable and more (internal) discussion is needed. We will bring this up at a later point (likely for the release after the next one) again.

Thus, this PR is much shorter than initially. It brings in `NXhistory` and `NXactivity` (as already approved in #1419) to `NXsample`, to describe the history of physical activities that have happened on the sample (preparation, calcination, _in situ_ treatment, ...). It also add `NXsubstance` (meant to describe a "pure" chemical substance) to the contributed_definitions.


base_classes/NXsample.nxdl.xml
```diff
@@ -379,13 +382,63 @@
...
+             NXsample_component.
+        </doc>
+    </group>
+    <!--exists: [max, 1]-->
+    <group type="NXfabrication">
+        <doc>
+             Details about the sample vendor (company or research group)
+        </doc>
+    </group>
+    <group name="identifier" type="NXidentifier">
```
sanbrock, line 423: comes from FABRICATION

base_classes/NXsample_component.nxdl.xml
```diff
@@ -139,6 +144,48 @@
...
+             using NXsample_component.
+        </doc>
+    </group>
+    <!--exists: [max, 1]-->
+    <group type="NXfabrication">
+        <doc>
+             Details about the sample component vendor (company or research group)
+        </doc>
+    </group>
+    <group name="identifier" type="NXidentifier">
```
sanbrock, line 172: comes with FABRICATION

base_classes/NXsubstance.nxdl.xml
```diff
@@ -0,0 +1,119 @@
...
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+#
+# For further information, see http://www.nexusformat.org
+-->
+<definition xmlns="http://definition.nexusformat.org/nxdl/3.1" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" category="base" type="group" name="NXsubstance" extends="NXobject" xsi:schemaLocation="http://definition.nexusformat.org/nxdl/3.1 ../nxdl.xsd">
+    <doc>
```
sanbrock, line 25: can now also inherit IDENTIFIER

base_classes/NXrotation_set.nxdl.xml
```diff
@@ -0,0 +1,256 @@
...
+             In practice one often faces situations where indeed these assumed
+             symmetries are anyway not fully observed, and thus an accepting of
+             eventual inaccuracies just for the sake of reporting a simplified
+             symmetrized description should be avoided.
+        </doc>
+        <dimensions rank="1">
+            <dim index="1" value="n_phases"/>
+        </dimensions>
+    </field>
+    <field name="rotation_quaternion" type="NX_NUMBER" units="NX_DIMENSIONLESS">
```
sanbrock, line 119: NX_QUATERNION is now available

lukaspie: changed

base_classes/NXsample_component.nxdl.xml
lukaspie: removed

base_classes/NXsubstance.nxdl.xml
lukaspie: will be implemented when #1486 has been discussed

lukaspie:
Dev notes to keep track:

- NXactivity: covered by https://github.com/nexusformat/definitions/pull/1419
- NXhistory: covered by https://github.com/nexusformat/definitions/pull/1419

ToDo:
 - NXsubstance: use `identifier`

base_classes/NXsubstance.nxdl.xml
lukaspie: implemented

base_classes/NXhistory.nxdl.xml
```diff
@@ -0,0 +1,57 @@
...
+         
+         Ideally, a full report of the previous operations (or links to a chain of operations).
+         Alternatively, notes allow for additional descriptors in any format.
+    </doc>
+    <group type="NXactivity">
+        <doc>
+             Any activity that was performed on the physical entity prior or during the experiment.
+        </doc>
+    </group>
+    <attribute name="identifierNAME">
```
mkuehbach, line 36: Why not inherited from NXobject ?

phyy-nx:
From Telco:
- Remove NXfabrication as it's in NXcomponent
- Note that NXsubstance does not need a vote as it's a contributed definition
- Note that NXhistory and NXactivity are part of #1419

lukaspie:
> From Telco:
> 
> * Remove NXfabrication as it's in NXcomponent

Done!

phyy-nx:

(note that NXsubstance does not need a vote as it's a contributed definition and note that NXhistory and NXactivity are part of https://github.com/nexusformat/definitions/pull/1419 which is due to be merged)

base_classes/NXsample.nxdl.xml
```diff
@@ -379,13 +379,32 @@
...
 	</group>
-        <group type="NXoff_geometry" minOccurs="0">
-         <doc>
-           This group describes the shape of the sample
-          </doc>
-        </group>
+	<group type="NXoff_geometry" minOccurs="0">
+		<doc>
+			This group describes the shape of the sample
+    	</doc>
```
PeterC-DLS, line 387: Make indentation consistent

lukaspie: done!

lukaspie:
Had to run black on `dev_tools/tests/test_nxdl_utils.py` to get the CI to pass.

base_classes/NXactivity.nxdl.xml
```diff
@@ -0,0 +1,62 @@
...
+            Start time of this activity. It is recommended to include local time zone information.
+        </doc>
+    </field>
+    <field name="end_time" type="NX_DATE_TIME">
+        <doc>
+             End time of this activity. It is recommended to include local time zone information.
+        </doc>
+        <attribute name="estimated" type="NX_BOOLEAN">
+            <doc>
+                In some cases, the end time of an activity can only be estimated. In this case,
```
paulmillar, line 41: Three points...

1. What if the start_time is an estimate?  Although unusual, I think it would be better to support estimates in a symmetric fashion, and support either (or both) start- and end times being estimates.
2. All measurements are subject to uncertainty.  Isn't this estimated just the same: a measurement with a significant uncertainty?  If so, then the question is rather: what is the uncertainty in this time measurement.  A boolean value is a rather crude way of capturing this information.
3. ISO 8601 already supports (to a limited degree) specifying uncertainty by leaving off  digits.  Would this be sufficient to cover the use-case?

base_classes/NXactivity.nxdl.xml
```diff
@@ -0,0 +1,62 @@
...
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+#
+# For further information, see http://www.nexusformat.org
+-->
+<definition xmlns="http://definition.nexusformat.org/nxdl/3.1" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" category="base" type="group" name="NXactivity" extends="NXobject" xsi:schemaLocation="http://definition.nexusformat.org/nxdl/3.1 ../nxdl.xsd">
+    <doc>
+         A planned or unplanned action that has a temporal extension and for some time depends on some entity.
+         
+         This class is a super class for all other activities.
```
paulmillar, line 28: It strikes me that the intent is to capture provenance information; i.e., some DAG of actions leading up to the current state.

If so, you might like to review some existing work in recording provenance; e.g., [W3C's PROV Ontology](https://www.w3.org/TR/prov-o/)

base_classes/NXhistory.nxdl.xml
```diff
@@ -0,0 +1,57 @@
...
+        <doc>
+             An ID or reference to the location or a unique (globally
```
paulmillar, line 38: This attribute seems to capture two concepts: the location where something happened and some identifier for the history event.  These should be recorded separately.

base_classes/NXhistory.nxdl.xml
```diff
@@ -0,0 +1,57 @@
...
+             persistent) identifier of e.g. another file which gives
+             as many as possible details of the history event.
```
paulmillar, line 40: I'm a little confused: does NXhistory represent a sequence of activities (a sequence of history events) or a single activity (a single history event).

The documentation mentions "the history event", suggesting NXhistory is describing a single history event, which (I think) isn't the intention.

base_classes/NXhistory.nxdl.xml
```diff
@@ -0,0 +1,57 @@
...
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+#
+# For further information, see http://www.nexusformat.org
+-->
+<definition xmlns="http://definition.nexusformat.org/nxdl/3.1" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" category="base" type="group" name="NXhistory" extends="NXobject" xsi:schemaLocation="http://definition.nexusformat.org/nxdl/3.1 ../nxdl.xsd">
+    <doc>
+         A set of activities that occurred to a physical entity prior/during experiment.
```
paulmillar, line 26: There seems to be some ambiguity in how to group activities.

Here's an illustrative example:

Let's say a sample-component's timeline involves two NXactivities: activity-A, activity-B.

These two activities could be placed:
  * in the same NXhistory, under the same group (of type NXactivity).
  * in the same NXhistory under two groups (each of type NXactivity).
  * in two different NXhistory.

That gives us three different ways of placing two activities.  Given this flexibility, there should be some semantic differences between these three possibilities: different placements should mean something different --- otherwise why support this?

If there _is_ a semantic different between these different layouts then it doesn't seem to be described in NXhistory's description.  If there isn't a semantic difference then why support this ambiguity?

This problem might be resolved by simply improving the documentation: a better description of the semantics.

base_classes/NXsample.nxdl.xml
```diff
@@ -379,13 +379,32 @@
...
 	<group type="NXpositioner">
-	  <doc>Any positioner (motor, PZT, ...) used to locate the sample</doc>
+		<doc>Any positioner (motor, PZT, ...) used to locate the sample</doc>
+	</group>
+	<group type="NXoff_geometry" minOccurs="0">
+		<doc>
+			This group describes the shape of the sample
+		</doc>
+	</group>
+	<field name="physical_form">
```
paulmillar, line 389: Should this be an enumeration?

contributed_definitions/NXsubstance.nxdl.xml
```diff
@@ -0,0 +1,124 @@
...
+        <doc>
+             User-defined chemical name of the substance
+        </doc>
+    </field>
+    <field name="molecular_mass" type="NX_FLOAT" units="NX_MOLECULAR_WEIGHT">
+        <doc>
+             Molecular mass of the substance
+        </doc>
+    </field>
+    <field name="cas_number">
```
paulmillar, line 41: Why not just use identifier_cas ?

contributed_definitions/NXsubstance.nxdl.xml
```diff
@@ -0,0 +1,124 @@
...
+                <item value="URL"/>
+            </enumeration>
+        </attribute>
+    </field>
+    <group name="cas_image" type="NXnote">
+        <doc>
+             CAS REGISTRY image
+        </doc>
+    </group>
+    <field name="cas_synonyms">
```
paulmillar, line 67: This seems to create a cache copy of information stored in CAS, with the risk that this information become stale (=="wrong") over time.

contributed_definitions/NXsubstance.nxdl.xml
```diff
@@ -0,0 +1,124 @@
...
+    <field name="identifier_inchi_str">
+        <doc>
+             String InChi identifier.
+             The InChI identifier expresses chemical structures in terms of atomic connectivity,
+             tautomeric state, isotopes, stereochemistry and electronic charge in order to
+             produce a string of machine-readable characters unique to the respective molecule.
+             For further information, see https://iupac.org/who-we-are/divisions/division-details/inchi/.
+        </doc>
+    </field>
+    <field name="identifier_inchi_key">
```
paulmillar, line 81: Having both the `identifier_inchi_str` field and the `identifier_inchi_key` seems redundant.  If you have one, you should be able to figure out the other.  If so, do we really need to store (or support storing) both?

contributed_definitions/NXsubstance.nxdl.xml
```diff
@@ -0,0 +1,124 @@
...
+             Hashed version of the full InChI (using the SHA-256 algorithm).
+        </doc>
+    </field>
+    <field name="iupac_name">
+        <doc>
+             Name according to the IUPAC system (standard).
+             For further information, see https://iupac.org/.
+        </doc>
+    </field>
+    <field name="smile">
```
paulmillar, line 93: Perhaps `identifier_smile` ?

contributed_definitions/NXsubstance.nxdl.xml
```diff
@@ -0,0 +1,124 @@
...
+        <doc>
+             Identifier in the SMILES (Simplified Molecular Input Line Entry System) system
```
paulmillar, line 95: Why support storing non-canonical representations?

contributed_definitions/NXsubstance.nxdl.xml
```diff
@@ -0,0 +1,124 @@
...
+             For further information, see https://www.daylight.com/smiles/.
+        </doc>
+    </field>
+    <field name="canonical_smile">
```
paulmillar, line 99: Perhaps `identifier_smile` ?

dev_tools/tests/test_nxdl_utils.py
```diff
@@ -1,6 +1,4 @@
-"""This is a code that performs several tests on nexus tool
-
-"""
+"""This is a code that performs several tests on nexus tool"""
```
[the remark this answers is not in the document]
lukaspie: It has to be here so that the CI/CD passes (see my comment above). A different version of `black` is installed in the CI actions than what was used for the main branch. The real problem here is probably that the `black` version is not pinned here.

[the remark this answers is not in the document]
paulmillar: Sorry for missing your earlier comment ... and the change is, therefore, reasonable.

Could you open an issue, reporting that the `black` version should be pinned but currently isn't.  That way, we don't forget to fix this.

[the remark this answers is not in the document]
lukaspie: No problem, see https://github.com/nexusformat/definitions/issues/1537

prjemian:
Could the expected behavior be specified in a configuration file for the tool?  This would avoid introducing an additional maintenance requirement.

base_classes/NXsample.nxdl.xml
lukaspie: Here, I have the same problem as for the chirp in NXbeam, namely that I cannot name all possible physical forms of samples. An open enumeration would probably be okay though, but this depends on #1521 and thus cannot be addressed here (as the vote is happening here already).

base_classes/NXhistory.nxdl.xml
lukaspie: I don't think the first option is valid. `NXactivity` clearly only talks about a single activity. I guess if sub-tasks are something that people want in the future, this concept could be further divided.

Regarding the other two points, indeed the idea is that you can do both. However, by using the named group `history(NXhistory)` in both `NXinstrument (see #1419) and `NXsample`, it is strongly suggested that one should place all activities of a single physical entity into the same `NXhistory`  group.

I would be open to add such a named `history` group to e.g. `NXcomponent`, with a strong statment that this is the intended used case. This would need to be a separate PR though.

base_classes/NXsample.nxdl.xml
paulmillar: OK, fair enough.

Could you (similar to chirp) add a "REVISIT" XML comment, to act as a reminder?

base_classes/NXactivity.nxdl.xml
paulmillar: Also, the document says "This class is a super class for all other activities."

However, I believe this is only true if some other class extends NXactivity, which doesn't happen.  Instead NXhistory includes a group of type NXactivity, rather than extending from it.

Perhaps better expressed as "This class is the basis for recording all activities." or something similar.

base_classes/NXsample.nxdl.xml
lukaspie: Done.

lukaspie:
> Could the expected behavior be specified in a configuration file for the tool? This would avoid introducing an additional maintenance requirement.

Wait, is this about the black formatting? Maybe better to be discussed in #1537? That being said, black (intentionally) almost gives no config options to the user, so probably this is not possible.

base_classes/NXactivity.nxdl.xml
lukaspie: By the way, since you were asking for it, `NXactivity` and `NXhistory` were approved in their current form already with #1419 (vote in https://github.com/nexusformat/definitions/issues/1452#issuecomment-2382668722) and are only here to make the documentation build work. Therefore, I will not make any changes to them until #1419 is merged.

paulmillar: Thanks for the clarification.  Let's close this discussion and continue once the rest is approved.

contributed_definitions/NXsubstance.nxdl.xml
lukaspie: I think people often work with just the SMILES system (not the canonical one) and there is an algorithm to create the canonical version based on that. So again, you could enter the smiles and the canonical version would get calculated automatically.

But I am certainly no expert on this, so would be open to changing this.

contributed_definitions/NXsubstance.nxdl.xml
lukaspie: Changed to `identifier_canonical_smiles`, see reasoning above

contributed_definitions/NXsubstance.nxdl.xml
lukaspie: Agreed, this can be removed

contributed_definitions/NXsubstance.nxdl.xml
```diff
@@ -0,0 +1,124 @@
...
+        </doc>
+    </group>
+    <field name="cas_synonyms">
+        <doc>
+             Synonyms in the CAS system.
+        </doc>
+    </field>
+    <field name="identifier_inchi_str">
+        <doc>
+             String InChi identifier.
```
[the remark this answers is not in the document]
lukaspie: Implemented

contributed_definitions/NXsubstance.nxdl.xml
```diff
@@ -0,0 +1,124 @@
...
+             For further information, see https://iupac.org/who-we-are/divisions/division-details/inchi/.
+        </doc>
+    </field>
+    <field name="identifier_inchi_key">
+        <doc>
+             Condensed, 27 character InChI key.
+             Hashed version of the full InChI (using the SHA-256 algorithm).
+        </doc>
+    </field>
+    <field name="iupac_name">
```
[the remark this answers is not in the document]
lukaspie: changed

contributed_definitions/NXsubstance.nxdl.xml
lukaspie: I think people often work with just the SMILES system (not the canonical one) and there is an algorithm to create the canonical version based on that. So again, you could enter the smiles and the canonical version would get calculated automatically.

But I am certainly no expert on this, so would be open to changing this.

contributed_definitions/NXsubstance.nxdl.xml
lukaspie: Changed to identifier_canonical_smiles, see reasoning above

base_classes/NXhistory.nxdl.xml
lukaspie: Since you can have multiple of these, there could be an identifier for the locataion and for the event. I don't really see the problem with the docstring here.

In any case, `NXhistory` as it is here was already voted on with #1419, so I don't plan to make changes to it in this PR here.

lukaspie:
Rebased after merging #1419, so `NXactivity` and `NXhistory` are no longer part of this PR. Comments above should then become issues if needed.

base_classes/NXhistory.nxdl.xml
phyy-nx: Resolving.  @paulmillar if you have more feedback, please move to another issue

contributed_definitions/NXsubstance.nxdl.xml
phyy-nx: Resolving.  @paulmillar if you have more feedback, please move to another issue

contributed_definitions/NXsubstance.nxdl.xml
phyy-nx: Resolving.  @paulmillar if you have more feedback, please move to another issue

contributed_definitions/NXsubstance.nxdl.xml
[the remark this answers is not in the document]
phyy-nx: Resolving.  @paulmillar if you have more feedback, please move to another issue

contributed_definitions/NXsubstance.nxdl.xml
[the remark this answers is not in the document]
phyy-nx: Resolving.  @paulmillar if you have more feedback, please move to another issue

contributed_definitions/NXsubstance.nxdl.xml
phyy-nx: Resolving.  @paulmillar if you have more feedback, please move to another issue

contributed_definitions/NXsubstance.nxdl.xml
phyy-nx: Resolving.  @paulmillar if you have more feedback, please move to another issue

contributed_definitions/NXsubstance.nxdl.xml
phyy-nx: Resolving.  @paulmillar if you have more feedback, please move to another issue

phyy-nx (approved):
The vote has passed. Thanks all. @lukaspie please merge at your convenience.

@paulmillar please be sure any issues you raised that you'd like to pursue are moved to a new issue. Note I resolved all of your comments but I'd like you to rescue your feedback from them! Thanks.

contributed_definitions/NXsubstance.nxdl.xml
lukaspie: Just FYI, I implemented the comments from @paulmillar on NXsubstance already (as it is contributed anyway and not part of this vote).

PR #1415, opened 2024-09-23
Fairmat 2024: use NXcoordinate_system together with NXtransformations

lukaspie:
This PR is concerned with coordinate systems and the transformations that are intricately linked to them. It was started for multiple reasons:
- There is the frequent need to explicitly describe which coordinate system is used in an experiment. 
- NeXus by default uses a [coordinate system](https://manual.nexusformat.org/design.html#the-nexus-coordinate-system) that is the same as that used by [McStas](http://mcstas.org/). However, this can be limiting:
  - there are now several techniques coming in where there is no explicit beam defined: temperature-dependant IV measurements, scanning probe microscopy (STM, AFM, Atom Probe, or multi-beam EM), etc.
  - in some communities, there already exists an agreement on which coordinate system shall be used, which is different than what NeXus uses. As an example, there is an existing [ISO standard for data transfer in surface chemical analysis](https://www.iso.org/standard/24269.html) that uses a coordinate system that is left-handed and defined based on the sample stage. It would be good if in an application definition (e.g. `NXxps`) one could explicitly state which coordinate system is to be used. There has already been some discussion on this, see https://github.com/nexusformat/definitions/issues/1464#issuecomment-2504529382
    - Specifically on this, for the XPS community, there are certain angles (e.g. analyser-to-beam) that are needed to perform rigorous analysis. It would be helpful if one could write these in the existing coordinate system (as in the ISO standard), without needing to transform back from the NeXus coordinate system. 
- NeXus (with its concept of `NXtransformations`) does already allow to define a specific coordinate system (see e.g. discussion in https://github.com/nexusformat/definitions/pull/1278). But, this has several limitations: 
  - these transformations always depend on a tool that can read them out and perform the transformations itself. We would also like to allow that one can give an coordinate system explicitly, both in a machine _and_ human readable way
  - left-handed CSs are currently not supported
  - fractional coordinates are not supported

Therefore, we introduce in this PR two new base classes: `NXcoordinate_system` and `NXrotation_conventions`.

Explanation of using `NXcoordinate_system` and `NXtransformation`:
- An application defintion contains one or more of `NXcoordinate_system` right below `NXentry`. See how to use different number of instances `NXcoordinate_system` from its lengthy docstring.
- in `NXcoordinate_system`, there is a `depends_on`, and `(NXtransformation)` group
- in `NXtransformation` `AXISNAME` `depends_on`, one can place either `"."`, or the path to an `NX_coordinate_system`
- if a `depends_on` attribute or field links to a `NXcoordinate_system`, it should pick the respective `depends_on` field in that class, and apply the specified `TRANSFORMATIONS`

Example of how this is implemented in a (proposed) application definition: [NXXps](https://fairmat-nfdi.github.io/nexus_definitions/classes/contributed_definitions/NXxps#nxxps-entry-geometries-xps-coordinate-system-group). Note that this one still uses the logical grouping `NXcoordinate_system_set`, but we removed this because it is not really needed. We would simply have `xps_coordinate_system(NXcoordinate_system)` directly under `NXxps/NXentry`.


lukaspie:
Copying here discussion that was buried in #1464 for clarity:

LP: we shall align to and extend the existing ISO standard for XPS. The concept of coordinate systems allow connecting multiple conventions.
PM: Which use cases needs a different coordinate system?
LP: ISO standard requires a different coordinate system
SB: Another example is when NeXus coordinate system being linked to incident beam does not exists because no beam exists in a specific experiment type, like temperature dependent IV measurement.
PM: shall we make this a standard, so all sw shall implement the ability of changing coordinate systems?
HB: NeXus shall allow multiple coordinate systems, as MX is also using it and other fields also require that. It is preferred to have a clearly described way how to do that.
PC: NeXus uses McStas as default, but other coordinate systems can also be used. Important is that the use of NXtransformations shall be consistent throughout the datasets.
SB: This is exactly what the proposal is about: provide a simple solution either on how to use NXtransformations when a base change is required to connect two coordinate systems, or on describing a coordinate system which is used in the definition and data file.
PM: Cannot we use a label for this purpose?
SB: Indeed that is what hapenning: Currently, the NXtransformations chain is followed along @depends_on until it ends with '.'. The proposal is that instead of ending with a '.', it could end with a label pointing to a Coordinate System which either describes a coordinate system on its own, or connetcs it to another coordinate system using NXtransformations.

mkuehbach:
"_set" is a restricted keyword thus NXcoordinate_system_set class needs a name change.

NXcoordinate_system_set is meant to provide a specialized NXobject with several commonly used NXcoordinate_system conventions with their own specialized NXcoordinate_system (CS) such that this collection can directly be used in appdefs instead of having each app demanding that if they wish to define one or multiple of the CS defined in NXcoordinate_system need to type the definitions again.

Possibilities:
1. Rename NXcoordinate_system_collection (lousy but straightforward)
2. Move top-level docstring of CS_set class to CS (+ would not loose context but - would loose predefined CS (processing_reference_frame, sample_reference_frame, etc.)
3. Similar to 2. delegate the definition of several CS to each appdef with maxOccurs not constraints, in practice people will then likely write CS1, CS2, CS3 etc.

mkuehbach:
The previous thought brings up one idea, if you can chain transformations as a result of which objects are rotated from one CS to another what if one allows to even chain CS instances, like a chain
McStas > chain of NXtransformations > CS1 > chain of NXtransformations > CS2 ... end point

base_classes/NXcoordinate_system_set.nxdl.xml
```diff
@@ -0,0 +1,240 @@
...
+<!--
+express conventions explicitly and understandable
+use depends_on field - not attribute - to point to conventions used-->
+<definition xmlns="http://definition.nexusformat.org/nxdl/3.1" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" category="base" type="group" name="NXcoordinate_system_set" extends="NXobject" xsi:schemaLocation="http://definition.nexusformat.org/nxdl/3.1 ../nxdl.xsd">
+    <doc>
+         Base class to hold different coordinate systems and representation conversions.
+         
+         How many nodes of type :ref:`NXcoordinate_system_set` should be used in an application definition?
+         
+         * 0; if there is no instance of :ref:`NXcoordinate_system_set` and therein or elsewhere across
```
mkuehbach, line 33: docstring needs edits here as sounds like from the perspective of somebody writing the proposal

mkuehbach:
Some comments from @PeterC-DLS made in #1421 will be taken over here

base_classes/NXcoordinate_system.nxdl.xml
```diff
@@ -0,0 +1,212 @@
...
+    <field name="alias" type="NX_CHAR">
+        <doc>
+             An alternative name given to this coordinate system.
+        </doc>
+    </field>
+    <field name="type" type="NX_CHAR">
+        <doc>
+             Coordinate system type.
+        </doc>
+        <enumeration>
```
lukaspie, line 113: "An open enumeration?"

Comment originally posted by @PeterC-DLS in https://github.com/nexusformat/definitions/pull/1421#discussion_r1914968590

base_classes/NXcoordinate_system.nxdl.xml
```diff
@@ -0,0 +1,212 @@
...
+             Human-readable field telling in which direction the x-axis points if that
+             instance of :ref:`NXcoordinate_system` has no reference to any parent and as such
+             is the mighty world reference frame.
+             
+             Exemplar values could be direction of gravity.
+        </doc>
+    </field>
+    <field name="x" type="NX_NUMBER" units="NX_LENGTH">
+        <doc>
+             Basis unit vector along the first axis which spans the coordinate system.
```
lukaspie, line 144: "You should explain that this coordinate system is defined relative to the laboratory frame."

Comment originally posted by @PeterC-DLS in https://github.com/nexusformat/definitions/pull/1421#discussion_r1914979924

base_classes/NXcoordinate_system.nxdl.xml
```diff
@@ -0,0 +1,212 @@
...
+             This axis is frequently referred to as the x-axis in real space and
+             the i-axis in reciprocal space.
+        </doc>
+        <dimensions rank="1">
+            <dim index="1" value="3"/>
```
lukaspie, line 149: "Why restrict yourself to 3 dimensions?"

Comment originally posted by @PeterC-DLS in https://github.com/nexusformat/definitions/pull/1421#discussion_r1914990929

base_classes/NXcoordinate_system_set.nxdl.xml
lukaspie: done

base_classes/NXcoordinate_system.nxdl.xml
lukaspie: Can only be done after addressing #1521

base_classes/NXcoordinate_system.nxdl.xml
```diff
@@ -0,0 +1,212 @@
...
+    </doc>
+    <field name="origin" type="NX_CHAR">
+        <doc>
+            Human-readable field telling where the origin of this CS is.
+            Exemplar values could be *left corner of the lab bench*, *door handle*
+            *pinhole through which the electron beam exists the pole piece*.
+            *barycenter of the triangle*, *center of mass of the stone*.
+        </doc>
+    </field>
+    <field name="alias" type="NX_CHAR">
```
sanbrock, line 104: suggestion: take this off from this proposal.

lukaspie:
I wanted to add to the discussion on handedness of coordinate system and the implied rotations.

If you think about a rotation around the z-axis from the perspective of viewing from the origin towards the positive end of the z-axis, you have:

- Right-Handed Coordinate System: positive rotation is counterclockwise.
```math
$$
R_z^{\text{right}}(\alpha) = \begin{pmatrix}
\cos\alpha & -\sin\alpha & 0 \\
\sin\alpha & \cos\alpha & 0 \\
0 & 0 & 1
\end{pmatrix}
$$
```

- Left-Handed Coordinate System: Positive rotation is clockwise.

```math
R_z^{\text{left}}(\alpha) = \begin{pmatrix}
\cos\alpha & \sin\alpha & 0 \\
-\sin\alpha & \cos\alpha & 0 \\
0 & 0 & 1
\end{pmatrix}
```

For any arbitrary rotation around axis $\mathbf{u} = (u_x, u_y, u_z)$ by an angle $\theta$, the rotation matrix is:
- For Right-Handed Coordinate System:
```math
R_{\text{right}} = \begin{pmatrix}
\cos\theta + u_x^2 (1 - \cos\theta) & u_x u_y (1 - \cos\theta) - u_z \sin\theta & u_x u_z (1 - \cos\theta) + u_y \sin\theta \\
u_y u_x (1 - \cos\theta) + u_z \sin\theta & \cos\theta + u_y^2 (1 - \cos\theta) & u_y u_z (1 - \cos\theta) - u_x \sin\theta \\
u_z u_x (1 - \cos\theta) - u_y \sin\theta & u_z u_y (1 - \cos\theta) + u_x \sin\theta & \cos\theta + u_z^2 (1 - \cos\theta)
\end{pmatrix}
```

- For Left-Handed Coordinate System:
```math
R_{\text{left}} = \begin{pmatrix}
\cos\theta + u_x^2 (1 - \cos\theta) & u_x u_z (1 - \cos\theta) + u_y \sin\theta & u_x u_y (1 - \cos\theta) - u_z \sin\theta \\
u_z u_x (1 - \cos\theta) - u_y \sin\theta & \cos\theta + u_z^2 (1 - \cos\theta) & u_z u_y (1 - \cos\theta) + u_x \sin\theta \\
u_y u_x (1 - \cos\theta) + u_z \sin\theta & u_y u_z (1 - \cos\theta) - u_x \sin\theta & \cos\theta + u_y^2 (1 - \cos\theta)
\end{pmatrix}
```
Wouldn't that have to be the rotations that we represent by the vectors in `NXtransformations`?

Currently, the latter transformation is not possible with the definition of `NXtransformations`, which explicitly talks about right-handed transformations only.  These matrices should probably be part of the docstring (either in `NXcoordinate_system` or `NXtransformations`) if we want to allow both left- and right-handed CSs and rotations. There (or in `NXrotation_conventions` or a similar class), we should probably also state the orientation that we choose (namely, we are looking from the origin towards positive axis values. This is also in line with what is the current definition of the [NeXus coordinate system](https://manual.nexusformat.org/design.html#the-nexus-coordinate-system))

See also [this](https://www.evl.uic.edu/ralph/508S98/coordinates.html) and [this](https://butterflyofdream.wordpress.com/2016/07/05/converting-rotation-matrices-of-left-handed-coordinate-system/) link.

base_classes/NXrotation_conventions.nxdl.xml
```diff
@@ -0,0 +1,116 @@
...
+        </enumeration>
+    </field>
+    <field name="rotation_convention" type="NX_CHAR">
+        <doc>
+            How are rotations interpreted into an orientation
+            according to convention 3 of
+            DOI: 10.1088/0965-0393/23/8/083501.
+        </doc>
+        <enumeration>
+            <item value="undefined"/>
```
sanbrock, line 61: Telco: take the option of being 'undefined' out, but rather make 'active'  the default. (as NXtransformations are also active transformations)

base_classes/NXrotation_conventions.nxdl.xml
```diff
@@ -0,0 +1,116 @@
...
+        </enumeration>
+    </field>
+    <field name="sign_convention" type="NX_CHAR">
+        <doc>
+             Which sign convention is followed when converting orientations
+             between different parameterizations/representations according
+             to convention 6 of DOI: 10.1088/0965-0393/23/8/083501.
+        </doc>
+        <enumeration>
+            <item value="undefined"/>
```
sanbrock, line 111: Telco: not to use undefined here either.

sanbrock:
Telco: suggestions made shall be implemented, and after a positive feedback from Wout, we could run a vote.

base_classes/NXrotation_conventions.nxdl.xml
```diff
@@ -0,0 +1,116 @@
...
+            <item value="rotation_angle_on_interval_zero_to_pi"/>
+        </enumeration>
+    </field>
+    <field name="sign_convention" type="NX_CHAR">
+        <doc>
+             Which sign convention is followed when converting orientations
+             between different parameterizations/representations according
+             to convention 6 of DOI: 10.1088/0965-0393/23/8/083501.
```
PeterC-DLS, line 108: There does not seem to be an explicit sentence of what convention 6 is in the cited paper. Maybe equation 14 is what needs to be cited.

base_classes/NXcoordinate_system.nxdl.xml
```diff
@@ -0,0 +1,212 @@
...
+            <item value="undefined"/>
+            <item value="cartesian"/>
+        </enumeration>
+    </field>
+    <field name="handedness" type="NX_CHAR">
+        <doc>
+             Handedness of the coordinate system if it is a Cartesian.
+        </doc>
+        <enumeration>
+            <item value="not applicable"/>
```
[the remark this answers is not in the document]
lukaspie: Done

base_classes/NXcoordinate_system.nxdl.xml
lukaspie: Done

base_classes/NXcoordinate_system.nxdl.xml
```diff
@@ -0,0 +1,207 @@
...
+          placed as high up in the node hierarchy (ideally right below an
+          instance of NXentry) of the application definition tree as possible.
+          This coordinate system shall be named such that it is clear how this
+          coordinate system is typically referred to in a community. For the
+          NeXus McStas coordinate system, it is advised to call it ``mcstas``
+          for the sake of improved clarity.
+
+          If this is the case, it is assumed that this ``NXcoordinate_system``
+          overwrites the NeXus default McStas coordinate system, i.e. users
+          can thereby conveniently and explicitly specify the speicific
```
PeterC-DLS, line 56:
```suggestion
          can thereby conveniently and explicitly specify the
```

base_classes/NXcoordinate_system.nxdl.xml
```diff
@@ -0,0 +1,207 @@
...
+-->
+<definition xmlns="http://definition.nexusformat.org/nxdl/3.1" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" category="base" type="group" name="NXcoordinate_system" extends="NXobject" xsi:schemaLocation="http://definition.nexusformat.org/nxdl/3.1 ../nxdl.xsd">
+    <doc>
+        Base class to detail a coordinate system (CS).
+
+        Whenever possible, all instances of :ref:`NXcoordinate_system`
+        should be used at the top-level (i.e, directly below ``NXentry``)
+        within an application definition or within a NeXus file.
+
+        How many nodes of type :ref:`NXcoordinate_system` should be used in
```
PeterC-DLS, line 32:
```suggestion
        How many groups of type :ref:`NXcoordinate_system` should be used in
```

base_classes/NXcoordinate_system.nxdl.xml
```diff
@@ -0,0 +1,207 @@
...
+          coordinate system is assumed.
+
+          McStas is used if no instance of NXcoordinate_system is specified.
+          However, for the sake of clarity, even in this case it is better
+          to be explicit like for every other coordinate system definition
+          to support users with interpreting the content and logic behind
+          every instance of the tree.
+
+        * 1; if only one :ref:`NXcoordinate_system` is defined, it should be
+          placed as high up in the node hierarchy (ideally right below an
```
PeterC-DLS, line 47:
```suggestion
          placed as high up in the tree hierarchy (ideally right below an
```

base_classes/NXcoordinate_system.nxdl.xml
```diff
@@ -0,0 +1,207 @@
...
+          coordinate system is typically referred to in a community. For the
+          NeXus McStas coordinate system, it is advised to call it ``mcstas``
+          for the sake of improved clarity.
+
+          If this is the case, it is assumed that this ``NXcoordinate_system``
+          overwrites the NeXus default McStas coordinate system, i.e. users
+          can thereby conveniently and explicitly specify the speicific
+          coordinate system that they wish to use.
+
+          This case has the advantage that it is explicit and faces no
```
PeterC-DLS, line 59:
```suggestion
          This case has the advantage that it is explicit and offers no
```

base_classes/NXcoordinate_system.nxdl.xml
```diff
@@ -0,0 +1,207 @@
...
+          ambiguities. However, in reality typically multiple coordinate
+          systems have to be mastered especially for complex multi-signal
+          modality experiments.
+
+        * 2 and more; as soon as more than one :ref:`NXcoordinate_system`
+          is specified somewhere in the tree, different interpretations are
+          possible as to which of these coordinate system sets and instances
```
PeterC-DLS, line 66:
```suggestion
          possible as to which of these coordinate systems and instances
```

base_classes/NXcoordinate_system.nxdl.xml
```diff
@@ -0,0 +1,207 @@
...
+          apply or take preference.
+          While these ambiguities should be avoided at all costs, the
+          opportunity for multiple sets and their instances enables to
```
PeterC-DLS, line 69: Reword with the removal of "sets"

base_classes/NXcoordinate_system.nxdl.xml
```diff
@@ -0,0 +1,207 @@
...
+          have branch-specific coordinate system conventions which are
```
PeterC-DLS, line 70: branch? Do you mean of a tree hierarchy?

base_classes/NXcoordinate_system.nxdl.xml
```diff
@@ -0,0 +1,207 @@
...
+          especially useful for deep classes where multiple scientific
```
PeterC-DLS, line 71:
```suggestion
          especially useful for deep groups where multiple scientific
```

base_classes/NXcoordinate_system.nxdl.xml
```diff
@@ -0,0 +1,207 @@
...
+          methods are combined or cases where having a definition of global
+          translation and conversion tables how to convert between
```
PeterC-DLS, line 73:
```suggestion
          conversion tables how to convert between
```
Remove translation as it has more than one meaning

base_classes/NXcoordinate_system.nxdl.xml
```diff
@@ -0,0 +1,207 @@
...
+          representations in different coordinate systems is not desired
+          or available for now.
+
+          If this case is realized, the best practice is that in every
+          case where a coordinate system should be referred to the respective
+          class has a ``depends_on`` field, which resolves the possible
```
PeterC-DLS, line 79:
```suggestion
          group has a ``depends_on`` field, which resolves the possible
```

base_classes/NXcoordinate_system.nxdl.xml
```diff
@@ -0,0 +1,207 @@
...
+          ``NXcoordinate_system``.
+
+        In any case, users are encouraged to write explicit and clean
+        ``depends_on`` fields in all groups that encode information for which
+        the interpretation of coordinate systems is relevant. If these
+        ``depends_on`` hints are not provided, it is automatically assumed
+        that all branches (to arbitrary depth) use either the only
+        ``NXcoordinate_system`` instance in the tree or the application
+        definition is considered underconstrained. In the latter case, which
+        should be avoided at all costs, McStas is assumed again.
```
PeterC-DLS, line 94: This wording is a bit too strong as it condemns all existing files!

base_classes/NXcoordinate_system.nxdl.xml
```diff
@@ -0,0 +1,207 @@
...
+             instance of :ref:`NXcoordinate_system` has no reference to any parent and as such
+             is the mighty world reference frame.
+             
+             Exemplar values could be direction of gravity.
+        </doc>
+    </field>
+    <field name="x" type="NX_NUMBER" units="NX_LENGTH">
+        <doc>
+             Basis unit vector along the first axis which spans the coordinate system.
+             This axis is frequently referred to as the x-axis in real space and
```
PeterC-DLS, line 140: What is the "real" space?

base_classes/NXrotation_conventions.nxdl.xml
lukaspie: Indeed, there is not convention 6. Changed to equation 14 as suggested.

base_classes/NXrotation_conventions.nxdl.xml
```diff
@@ -0,0 +1,116 @@
...
+            <item value="undefined"/>
+            <item value="passive"/>
+            <item value="active"/>
+        </enumeration>
+    </field>
+    <field name="euler_angle_convention" type="NX_CHAR">
+        <doc>
+            How are Euler angles interpreted given that
+            there are several choices (e.g. zxz, xyz)
+            according to convention 4 of DOI: 10.1088/0965-0393/23/8/083501.
```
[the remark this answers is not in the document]
lukaspie: implemented

base_classes/NXrotation_conventions.nxdl.xml
lukaspie: done

base_classes/NXrotation_conventions.nxdl.xml
lukaspie: `undefined` wasremoved from all enumerations in this base class

base_classes/NXcoordinate_system.nxdl.xml
lukaspie: Thanks for the suggestions, I accepted all of them.

base_classes/NXcoordinate_system.nxdl.xml
lukaspie: You are right. Changed to "In the latter case, which should ideally be avoided, McStas is assumed again. Note that this is the case for all files that were created before ``NXcoordinate_system`` was added."

base_classes/NXcoordinate_system.nxdl.xml
lukaspie: Yes, exactly. I also made the language a bit less harsh here, as discussed below.

Changed to 
"While these ambiguities should be avoided if possible, the opportunity for instances enables to have coordinate system conventions that are specific for some part of the NeXus tree. This is especially useful for deep groups where multiple scientific methods are combined or cases where having a definition of global conversion tables how to convert between representations in different coordinate systems is not desired or available for now."

base_classes/NXcoordinate_system.nxdl.xml
lukaspie: Done

base_classes/NXcoordinate_system.nxdl.xml
lukaspie: How about
```
<enumeration open="True">
    <item value="cartesian"/>
    <item value="polar"/>
    <item value="cylindrical"/>
    <item value="spherical"/>
    <item value="homogeneous "/>
</enumeration>

File: base_classes/NXcoordinate_system.nxdl.xml (hunk 21 above)
Comment (mkuehbach) [reply base_classes/NXcoordinate_system.nxdl.xml:142]: Google: "what is the opposite to reciprocal space"
```
The space described by q is called the Fourier space (the q-space) or the reciprocal space. The term “reciprocal” is used because q has the unit of the reciprocal length, i.e., l/λ. On the other hand, the space defined by r is called the real space.
```
Does writing Euclidean space clarify the situation?

File: base_classes/NXcoordinate_system.nxdl.xml (hunk 4 above)
Comment (mkuehbach) [reply base_classes/NXcoordinate_system.nxdl.xml:146]: This is a good point. If I would not specify a dimensionsType, this means that the value is not necessarily a scalar right?
Alternatively, we could define a symbol "d" like we use for the NXcg_* base classes and then define
dim index="1" value="d"
With a docstring comment d>=1

File: base_classes/NXcoordinate_system.nxdl.xml (hunk 3 above)
Comment (mkuehbach) [reply base_classes/NXcoordinate_system.nxdl.xml:141]: Mhh, puzzled by this one: McStas defines a coordinate system, where does this CS life in?
Likely somewhere but for NeXus irrelevant because the assumption is,
McStas is the laboratory frame, i.e. it is THE root coordinate system it defines i.e. space.
NXtransformations then act as "active" transformations which relocate, rotate objects in that fixed McStas frame.
Now people may wish to use NXcoordinate_system to define N instances of CSs different than McStas, where do these life in, my point is why every of these N CSs necessarily be defined in the laboratory frame (McStas), why one of these N not define the laboratory frame itself.
This is why I am puzzled and feel the comment that ```this Cs is defined relative to the lab frame``` not necessarily applies or right now confuses me.

**Comment by woutdenolf** [comment]
### 1.  Find the coordinate system that applies to an NXobject

*NXtransformations* have an explicit *depends_on* chain but `NXcoordinate_system` does not.

To find the `NXcoordinate_system` that applies to the current *NXobject*: look for the first
parent *NXobject* that has a `NXcoordinate_system` child. This `NXcoordinate_system` applies
in the current group. If none is found *McStas* is assumed.

This means that `NXcoordinate_system` must apply to all sibling groups of its parent group?

This also means you cannot have more than on `NXcoordinate_system` child group in the same parent group?

### 2.  Relating coordinate systems

*NXtransformations* provides the relationship between two `NXcoordinate_system` groups.

This comes with a caveat that would be good to mention:

*NXtransformations* are rigid transformations so you cannot always relate two coordinate systems.

Note that McStas is *orthonormal* and *right-handed*. This means that any system that is
not orthonormal and right-handed cannot be related to *McStas*.

### 3. Coordinate system for the basis vectors themselves

The x, y and z values are the coordinates of the basis vectors. These coordinates are given with respect to
the coordinate system after the last transformation in `NXtransformations`. If *NXtransformations*
is not provided, the coordinates cannot be related to in general *McStas* because that would not
work for non-orthonormal and left-handed systems. So it must be the *canonical Euclidean basis* (see below). Is all that correct? We'll need to clarify in the docs.

Do we want to prevent degenerate bases? If we do then we can add

> x, y, and y must be *linearly independent* and *spanning* to form a valid basis

### 4. Type

Currently we only have *cartesian*. Do we foresee others like *spherical*?

For example spherical coordinate systems coordinates are expressed as *(radius, azimuth, angle)* which are
defined with respect to the basis.

How do I know for a given field in some *NXobject* that the *type* applies?

### 5. Handedness

The *handedness* can be derived from the basis vectors. When providing both they could contradict eachother.

```python
def handedness(x, y, z):
    basis_matrix = numpy.column_stack((x, y, z))
    det = numpy.linalg.det(basis_matrix)
    return "right-handed" if det > 0 else "left-handed"
```

### 6. Canonical basis vs. McStas

*Euclidean space* $\mathbb {R}^3$ has an infinite number of orthonormal right-handed bases. They are
all related by a *proper rotation*. The *standard basis* (also called *natural basis* or *canonical basis*)
of Euclidean space $\mathbb {R}^3$ is this one

```
e1​=(1,0,0),e2​=(0,1,0),e3​=(0,0,1)
```

The NeXus coordinate system (a.k.a. *McStas* coordinate system) is the standard basis with Z the direction
of the primary beam and -Y is the direction of gravity. So defining *McStas* with `NXcoordinate_system`
would look like this

```
mcstas@NXcoordinate_system
  x = [1, 0, 0]
  y = [0, 1, 0]
  z = [0, 0, 1]
  @y_direction = "opposite to gravity"
  @z_direction = "direction of the primary beam"
```

Probably a good idea to add this so it is clear how `NXcoordinate_system` is related to what is the default in NeXus.

### 7. NXrotation_conventions

Maybe list all the conventions this supports

> This ensures consistency when switching between rotation representations such as quaternions, Euler angles, rotation matrices and rotation vectors.

How is `NXrotation_conventions` related to `NXcoordinate_system` and `NXtransformations`?

For example if we have an `NXtransformations` axis with `transformation_type = "rotation"`, does `rotation_handedness`
apply? `rotation_convention` cannot apply since `NXtransformations` must be active.

How do I know what convention applies in the current *NXobject*? Perhaps similar to `NXcoordinate_system`?

To find the `NXrotation_conventions` that applies to the current *NXobject*: look for the first parent *NXobject* that has a `NXrotation_conventions` child. This `NXrotation_conventions` applies in the current group. If none is found???

woutdenolf:
Another additional question: since `NXcoordinate_system` applies recursively, and suppose somewhere down the tree you have an `NXattenuator` with an `NXtransformations` group and the last field has an attribute `depends_on = "."`. Does this mean that the transformation chain starts from the `NXcoordinate_system` and not McStas?

mkuehbach:
https://www.iucr.org/__data/iucr/cifdic_html/2/cif_img.dic/Caxis.html

mkuehbach:
@PeterC-DLS what happened it one does not have a root coordinate system, will it be McStas, what is with methods where there is no beam on whose propagation direction McStas depends

mkuehbach:
How and under which conditions can application definitions combine definitions for rotation_conventions with NXcoordinate_system, https://iopscience.iop.org/article/10.1088/0965-0393/23/8/083501/ampdf, assuming orthonormal, canonical basis, right hand by default

woutdenolf:
NIAC mentioned this as the original inspiration for NXtransformations: https://www.iucr.org/__data/iucr/cifdic_html/2/cif_img.dic/Caxis.html

woutdenolf:
I tried writing and reading HDF5 files with `NXtransformations`, `NXcoordinate_system` and `NXcoordinate_system_set`

https://gitlab.esrf.fr/hdf5/nexus/nxspacetime

An example for an ESRF beamline (X along the beam, Z opposite to gravity)

https://gitlab.esrf.fr/hdf5/nexus/nxspacetime/-/blob/main/nxtransformations.h5

ED-XRD example

```
@NXroot
  1.1@NXentry
    esrf@NXcoordinate_system
      depends_on = transformations/beam 
      handedness = right_handed 
      transformations@NXtransformations
        beam in [1. 0. 0.] -> gravity
        beam_along_x = -90.0 degrees in [0. 1. 0.] -> /1.1/mcstas
        gravity in [ 0.  0. -1.] -> gravity_along_minz
        gravity_along_minz = -90.0 degrees in [1. 0. 0.] -> beam_along_x
      x = [1 0 0] 
      y = [0 1 0] 
      z = [0 0 1] 
    instrument@NXinstrument
      aperture@NXaperture
        depends_on = transformations/beam_along_minz 
        transformations@NXtransformations
          beam_along_minz = -90.0 degrees in [0. 1. 0.] -> distance_from_sample
          distance_from_sample = 2.0 mm in [1. 0. 0.] -> twotheta
          twotheta = 50.0 degrees in [0. 0. 1.] -> /1.1/sample
      beamstop@NXbeam_stop
        depends_on = transformations/distance_from_sample 
        transformations@NXtransformations
          distance_from_sample = 4.0 mm in [1. 0. 0.] -> /1.1/sample
      detector@NXdetector
        depends_on = transformations/distance_from_aperture 
        transformations@NXtransformations
          distance_from_aperture = 1.0 mm in [ 0.  0. -1.] -> /1.1/instrument/aperture
      mono@NXmonochromator
        depends_on = transformations/distance_to_sample 
        transformations@NXtransformations
          distance_to_sample = 2.0 mm in [-1.  0.  0.] -> /1.1/sample
    mcstas@NXcoordinate_system_set
      depends_on = transformations/beam 
      transformations@NXtransformations
        beam in [0. 0. 1.] -> gravity
        gravity in [ 0. -1.  0.] -> .
    sample@NXsample
      depends_on = /1.1/esrf 
```

# This is what I learned

## The `NXcoordinate_system` group is equivalent to a single `NXtransformations` field

It became very useful to look at `NXcoordinate_system` as a transformation, just like a `NXtransformations` field is a transformation.

- `NXtransformations` field: describes translation (vector + distance), rotation (vector + angle) or directions (vector)
- `NXcoordinate_system` group: describes linear transformation (change-of-basis matrix with basis vectors as columns)

Note: a linear transformation is not an affine transformation (translations are not supported).

## `depends_on` can refer to a `NXtransformations` field or `NXcoordinate_system` group

Currently this is only the case for the `depends_on` attribute of a `NXtransformations` field. I strongly suggest to also apply this to the `depends_on` dataset of any `NXobject` that has it (`NXaperture`, `NXdetector`, ... `NXcoordinate_system` and `NXcoordinate_system_set`). If not the code to read this become really awkward.

## `NXcoordinate_system` vs. `NXcoordinate_system_set` vs. any `NXobject` with a `depends_on` field

An `NXobject` group has a local coordinate system (e.g. `NXaperture` has origin in the center and the surface normal pointing to the source along Z).

An `NXcoordinate_system_set` has a local coordinate system which is specified with respect to a physical object.

An `NXcoordinate_system` has a local coordinate system which is specified in terms of basis vectors.

# Suggestions

1. `depends_on` can refer to a `NXtransformations` field or `NXcoordinate_system` group: should apply when it is an attribute and a field.
2. `depends_on` can be "." which is McStas but we need something else to say it does NOT depend on McStas. Maybe the simple absence of the field/attribute is enough.
3. The basis vectors of `NXcoordinate_system` group are defined with respect to the reference system it depends on. When "." this is McStas. When nothing it is the canonical Euclidean frame.

woutdenolf:
depends_on attribute and field:

When `depends_on` is missing:
   - If there is one single NXcoordinate_system group up in the tree it depends on that one (NIAC: do we need this?)
   - Else it depends on nothing
When depends_on=".": mcstas
Else: path to NXtransformations field or NXcoordinate_system

phyy-nx:
From Telco Mar 24 2025

- Remove handedness for now (note @yayahjb's comment about difficulty in fully representing handedness, and two types of handedness (coordinate system, but also rotation (clockwise/counterclockwise) and multidimensionality issues, see also [this comment.](https://github.com/nexusformat/definitions/pull/1415#issuecomment-2657229558)).  Note instead that when adding a rotation NXtransformation that depends on a coordinate system, that it's right handed in that coordinate system.
- Need a few sentences at the top of NXcoordinate_system and NXtransformations linking one to the other, describing what each is good at and some advice on how to use one or the other.
- depends_on fields can now also point to an NXcoordinate_system (note this in NXtransformations)
- Add note to NXtransformations that it is an active transformation, meaning that it transforms an object in the space of its parent.

General call for a re-write of NXtransformations, similar to what @woutdenolf recently did for NXdata.  Does not need to be part of this PR itself.

base_classes/NXcoordinate_system.nxdl.xml
```diff
@@ -0,0 +1,208 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<?xml-stylesheet type="text/xsl" href="nxdlformat.xsl"?>
+<!--
+# NeXus - Neutron and X-ray Common Data Format
+#
+# Copyright (C) 2014-2024 NeXus International Advisory Committee (NIAC)
```
mkuehbach, line 6: 2022 or 2023

base_classes/NXcoordinate_system.nxdl.xml
```diff
@@ -0,0 +1,208 @@
...
+          ambiguities which specific coordinate systems is referred to.
+
+          In the case of two or more instances of ``NXcoordinate_system`` it
+          is advised to specify the relationship between the two coordinate
+          systems by using the :ref:`NXtransformations` group within
+          ``NXcoordinate_system``.
+
+        In any case, users are encouraged to write explicit and clean
+        ``depends_on`` fields in all groups that encode information for which
+        the interpretation of coordinate systems is relevant. If these
```
mkuehbach, line 95: suggestion to add "and transformations" between "coordinate systems" and "is relevant" ?

base_classes/NXcoordinate_system.nxdl.xml
```diff
@@ -0,0 +1,208 @@
...
+        ``depends_on`` hints are not provided, it is automatically assumed
+        that all branches (to arbitrary depth) use either the only
+        ``NXcoordinate_system`` instance in the tree or the application
+        definition is considered underconstrained. In the latter case,
+        which ideally should be avoided, McStas is assumed again. Note that
```
mkuehbach, line 100: I think we can drop the last sentence, so far McStas was always the assumed coordinate system in which NXtransformations where used to describe the rotation of objects active interpretation

base_classes/NXtransformations.nxdl.xml
```diff
@@ -170,8 +179,9 @@
 		</attribute>
 		<attribute name="depends_on" type="NX_CHAR">
 			<doc>
-				Points to the path to a field defining the axis on which this
-				depends or the string ".".
+				Points to the path of a field defining the axis on which this instance of
+				NXtransformations depends or the string ".". It can also point to an instance of
```
mkuehbach, line 183: We should end a sentence that tells what does "." mean exactly.

base_classes/NXtransformations.nxdl.xml
```diff
@@ -108,6 +108,15 @@
...
 
 		as needed.
+
+		NOTE
+
+		NXtransformations follows the **active** transformation convention. This means that the transformation describes
+		how an object is moved or rotated within the coordinate system. In other words, the transformation
+		actively changes the position or orientation of the object itself, rather than altering the coordinate system in
+		which the object is described. This is in contrast to a **passive** transformation, which changes the frame of reference
+		or coordinate system, while the object remains fixed. In case it is needed to describe multiple coordinate system,
```
mkuehbach, line 118: multiple coordinate system**s**

base_classes/NXcoordinate_system.nxdl.xml
lukaspie: changed

base_classes/NXtransformations.nxdl.xml
lukaspie: changed

base_classes/NXtransformations.nxdl.xml
lukaspie: added

base_classes/NXcoordinate_system.nxdl.xml
lukaspie: Done

base_classes/NXcoordinate_system.nxdl.xml
lukaspie: I think in the last Telco it was stated that NeXus is mostly about three dimensions, so maybe we can keep that restriction? Not sure about polar coordinates though, for example.

woutdenolf:
@lukaspie @mkuehbach Thanks for the changes!

There is still one thing left that is unclear to me: `NXcoordinate_system` can have an effect on a group down the tree without an explicit `depends_on` field in that group. Is that correct? Can you define this behavior explicitly in the doc at the top of `NXcoordinate_system`? This should also reduce this doc.

As I understand (I might be wrong), any group that has an optional `depends_on` field or any field that has an optional `depends_on` attribute has a fallback when `depends_on` is not provided. It is not MCstas because in that case `depends_on` would have been provided with the value `"."`. Instead, the fallback behavior involves traversing up the hierarchy until you find the first parent that contains one and only one `NXcoordinate_system` group. If such a parent exists, its coordinate system applies. If none is found, then the coordinate system is not defined with respect to anything else.

What about groups that do *not* have an optional `depends_on` field? In what way would `NXcoordinate_system` apply then?

phyy-nx:
Note, this PR will go up for vote staring Thursday, April 17!

lukaspie:
> @lukaspie @mkuehbach Thanks for the changes!
> 
> There is still one thing left that is unclear to me: `NXcoordinate_system` can have an effect on a group down the tree without an explicit `depends_on` field in that group. Is that correct? Can you define this behavior explicitly in the doc at the top of `NXcoordinate_system`? This should also reduce this doc.
> 
> As I understand (I might be wrong), any group that has an optional `depends_on` field or any field that has an optional `depends_on` attribute has a fallback when `depends_on` is not provided. It is not MCstas because in that case `depends_on` would have been provided with the value `"."`. Instead, the fallback behavior involves traversing up the hierarchy until you find the first parent that contains one and only one `NXcoordinate_system` group. If such a parent exists, its coordinate system applies. If none is found, then the current coordinate system is not defined with respect to anything else.
> 

This was indeed the intended behavior, but exactly only if there is only one CS defined in the upper hierarchy (up until NXentry). In this case, if the optional `depends_on` field/attribute is not filled, it is the same as saying `depends_on="my-one-coordinate-system"`. I added a paragraph now that describes this in a bit more detail. Can you please check that you agree with 

> What about groups that do _not_ have an optional `depends_on` field? In what way would `NXcoordinate_system` apply then?

At least every physical component _always_ has an optional `depends_on` field because they all inherit from `NXcomponent`. For the base classes not extending `NXcomponent`, it doesn't really make sense for the CS to apply there, does it? They are anyway not located in space.

woutdenolf:
> * 0; if there is no instance of ``NXcoordinate_system`` across
>         the application definition, the default
>         `NeXus coordinate system &lt;https://manual.nexusformat.org/design.html#the-nexus-coordinate-system&gt;`_
>         (which is the same as the one used by `McStas &lt;https://mcstas.org/&gt;`_) is assumed.

> If none is found, then the current coordinate system is not defined with respect to anything else.

> If it is the string ".", it is explicitly pointing towards the default 
>				`NeXus coordinate system &lt;https://manual.nexusformat.org/design.html#the-nexus-coordinate-system&gt;`_.

The first two seem contradictory. I'm not sure what the current convention is. We depend on mcstas only when we have an explicit "." or even when depends_on is not provided?

Sorry for being difficult but we are almost there imo!

lukaspie:
> > * 0; if there is no instance of `NXcoordinate_system` across
> >   the application definition, the default
> >   `NeXus coordinate system &lt;https://manual.nexusformat.org/design.html#the-nexus-coordinate-system&gt;`_
> >   (which is the same as the one used by `McStas &lt;https://mcstas.org/&gt;`_) is assumed.
> 
> > If none is found, then the current coordinate system is not defined with respect to anything else.
> 
> > If it is the string ".", it is explicitly pointing towards the default
> > `NeXus coordinate system &lt;https://manual.nexusformat.org/design.html#the-nexus-coordinate-system&gt;`_.
> 
> The first two seem contradictory. I'm not sure what the current convention is. We depend on mcstas only when we have an explicit "." or even when depends_on is not provided?
> 
> Sorry for being difficult but we are almost there imo!

No worries, it is complicated :D

I think the first sentence was introduced initially because we didn't want to invalidate all the existing NeXus file before ´NXcoordinate_system` was used.

My suggestion is the following:
- If no NXcoordinate_system is given, you can either do `depends_on=".".` (pointing to Mcstas) or omit `depends_on` (then no assumption about the CS is taken). I think this was already the case, that if you wanted to use mcstas, you needed to end the chain of ".", right?
- If you have a CS somewhere in your tree, either you point directly it to using `depends_on` or you walk up the tree to find it (this is what we have called fallback). `depends_on=".".` still points to Mcstas, which you don't have to define explicitly (but you can do it if you want to, see the example in the docstring).
- If you have more than one CS at the same level, this fallback mechanism does not apply and you _must_ use proper depends_on everywhere.

Would that work? I can change the sentences above accordingly.

woutdenolf:
> Would that work?

Yep that sounds good to me. Thanks!

> I think this was already the case, that if you wanted to use mcstas, you needed to end the chain of ".", right?

No idea. I hope so because not every local coordinate system is relatable to mcstas.

https://manual.nexusformat.org/classes/base_classes/NXtransformations.html

> This class will usually contain all axes of a sample stage or goniometer or a detector. The NeXus default McSTAS coordinate frame is assumed, but additional useful coordinate axes may be defined by using axes for which no transformation type has been specified.

https://manual.nexusformat.org/design.html#nexus-geometry

> NeXus supports description of the shape, position and orientation of objects in [The NeXus Coordinate System](https://manual.nexusformat.org/design.html#design-coordinatesystem). Position and orientation can be defined as [Coordinate Transformations](https://manual.nexusformat.org/design.html#coordinatetransformations) using the [NXtransformations](https://manual.nexusformat.org/classes/base_classes/NXtransformations.html#nxtransformations) class.

It somewhat suggests that McSTAS is "the default coordinate frame" but never says what this means and it never says if you omit `depends_on` it automatically applies or not. So lets assume "not".

woutdenolf:
https://github.com/nexusformat/definitions/pull/1415/commits/dbefd09e874adcf83c38bc02c620ae70fa54bbcf that solves the last issue for me. Ready for a vote imo!

phyy-nx:
Hi all, as discussed in the Telco, we can now move this PR to an online vote. NIAC committee members please vote on this PR using emojis on this comment. 👍 for yes, 👎 for no, anything else (for example 👀) to abstain. We need 14 votes to hit quorum so please review and vote!

@mkuehbach @lukaspie, there are a couple of unresolved comments, can you review and resolve them?  I don't think they affect the vote.  Thanks.

base_classes/NXcoordinate_system.nxdl.xml
PeterC-DLS: Euclidean or physical works for me.

base_classes/NXcomponent.nxdl.xml
```diff
@@ -77,8 +77,12 @@
             to apply to the component starting from 0, 0, 0. The order of these operations
```
PeterC-DLS, line 77:
```suggestion
            to the component starting from 0, 0, 0. The order of these operations
```

base_classes/NXcoordinate_system.nxdl.xml
```diff
@@ -0,0 +1,238 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<?xml-stylesheet type="text/xsl" href="nxdlformat.xsl"?>
+<!--
+# NeXus - Neutron and X-ray Common Data Format
+#
+# Copyright (C) 2023-2024 NeXus International Advisory Committee (NIAC)
```
PeterC-DLS, line 6:
```suggestion
# Copyright (C) 2023-2025 NeXus International Advisory Committee (NIAC)
```

phyy-nx:
Vote has passed!  Once the reviews have finished, comments have been resolved, an approval posted, this can be merged.

base_classes/NXcoordinate_system.nxdl.xml
lukaspie: changed to Euclidean

lukaspie:
@PeterC-DLS could you please resolve your one remaining comment?

@phyy-nx @PeterC-DLS @woutdenolf if you have nothing else you would still like to change, I could appreciate an approval afterwards.

base_classes/NXcoordinate_system.nxdl.xml
```diff
@@ -0,0 +1,238 @@
...
+
+        ``NXcoordinate_system`` can be part of the transformations chain
+        using :ref:`NXtransformations`, where it acts as a linear
+        change-of-basis transformation (using a 3x3 matrix with the basis
+        vectors ``x``, `y``, and ``z`` as columns).
+
+        Any group that has an optional ``depends_on`` field or any field
+        that has an optional ``depends_on`` attribute has a fallback when
+        ``depends_on`` is not provided. The fallback behavior involves
+        traversing up the hierarchy until the first parent that contains one
```
PeterC-DLS, line 43:
```suggestion
        traversing up the hierarchy until the first ancestor that contains one
```

base_classes/NXcoordinate_system.nxdl.xml
```diff
@@ -0,0 +1,238 @@
...
+        and only one ``NXcoordinate_system`` group is found. If such a parent
```
PeterC-DLS, line 44:
```suggestion
        and only one ``NXcoordinate_system`` group is found. If such an ancestor
```

base_classes/NXcoordinate_system.nxdl.xml
```diff
@@ -0,0 +1,238 @@
...
+        exists, its coordinate system applies. If none is found, then the
+        current coordinate system is not defined with respect to anything
+        else. As an example, if there is only one NXcoordinate_system called
```
PeterC-DLS, line 47: Make explicit what happens when there's more than one CS group

base_classes/NXcoordinate_system.nxdl.xml
```diff
@@ -0,0 +1,238 @@
...
+        "my_coordinate_system" defined directly under ``NXentry``, each
+        optional ``depends_on`` field/attribute that is not defined
+        automatically defaults to ``depends_on=my_coordinate_system``.
+
+        How many groups of type ``NXcoordinate_system`` should be used in
+        an application definition?
+
+        * 0; if there is no instance of ``NXcoordinate_system`` across the
+          application definition or the data tree, you can use ``depends_on="."``
```
PeterC-DLS, line 56: "data tree"? Maybe, whole or entire tree.

base_classes/NXcoordinate_system.nxdl.xml
```diff
@@ -0,0 +1,238 @@
...
+          system) can be expressed as follows:
+
+          .. code-block::
+                    
+            mcstas@NXcoordinate_system
+                x = [1, 0, 0]
+                y = [0, 1, 0]
+                z = [0, 0, 1]
+                @y_direction = "opposite to gravity"
+                @z_direction = "direction of the primary beam"
```
PeterC-DLS, line 76: Note, this does not need to be the case if an `NXbeam` defines its direction (at the sample).

base_classes/NXcoordinate_system.nxdl.xml
```diff
@@ -0,0 +1,238 @@
...
+        the interpretation of coordinate systems and transformations is relevant.
+        If these ``depends_on`` instances are not provided or no instance of
+        ``NX_coordinate_system`` exists in the upper part of the hierarchy,
+        the application definition is considered underconstrained. Note that this
+        is the case for all files that were created before ``NXcoordinate_system``
+        was added.
+    </doc>
+    <field name="origin" type="NX_CHAR">
+        <doc>
+            Human-readable field telling where the origin of this CS is.
```
PeterC-DLS, line 136:
```suggestion
            Human-readable field describing where the origin of this CS is.
```

base_classes/NXcoordinate_system.nxdl.xml
```diff
@@ -0,0 +1,238 @@
...
+    <field name="x_alias" type="NX_CHAR">
+        <doc>
+            Opportunity to define an alias for the name of the x-axis.
+        </doc>
+    </field>
+    <field name="x_direction" type="NX_CHAR">
+        <doc>
+            Human-readable field telling in which direction the x-axis points if that
+            instance of :ref:`NXcoordinate_system` has no reference to any parent and as such
+            is the mighty world reference frame.
```
PeterC-DLS, line 160:
```suggestion
            is the world reference frame.
```
or global

base_classes/NXcoordinate_system.nxdl.xml
lukaspie: Added a clarifier that also then the current coordinate system is not defined with respect to anything else

base_classes/NXcoordinate_system.nxdl.xml
lukaspie: changed

base_classes/NXcoordinate_system.nxdl.xml
lukaspie: I still don't understand why/how the beam direction can be defined and is supposed to work, but I added a sentence here that this is assumin the beam doesn't define its direction at the sample.

base_classes/NXtransformations.nxdl.xml
```diff
@@ -148,6 +156,11 @@
 				For ``translation`` axes the direction should be chosen for
 				increasing displacement. For general axes, an appropriate direction
 				should be chosen.
+
+				Note that if the ``NXtransformation`` depends on a coordinate system
+				(i.e., its ``depends_on`` attribute points to an instance :ref:`NXcoordinate_system`),
```
PeterC-DLS, line 161: I think this will cause confusion - it is better to be consistent with geometric interpretations and say the positive rotation is along to which hand (screw) rule dictates for the coordinate system. Then the matrix transformation does not depend on handness.

lukaspie: I agree with you, but this is not what was agreed on in the last discussions. I think this would probably need another discussion/vote to clarify.

base_classes/NXcoordinate_system.nxdl.xml
PeterC-DLS: In my reciprocal space remapper, I use the beam direction vector to work out the momentum transfer vector of a pixel location. If the beam direction, detector and sample transformations chains are recorded in a file then we can break any connection to McStas.

base_classes/NXtransformations.nxdl.xml
PeterC-DLS: The existing documentation for `NXtransformations` should be changed as we did not consider nor was it possible to use left handed CSes. I think it has to be part of this PR as this makes left handed CSes available. So make the change here and bring it up for a re-vote.

base_classes/NXcoordinate_system.nxdl.xml
lukaspie: This is all fine for me, if you don't give an `depends_on` to a CS or define a coordinate system elsewhere up the tree, than you are completely disconnected from McStas or any other CS.

base_classes/NXcoordinate_system.nxdl.xml
```diff
@@ -0,0 +1,241 @@
...
+        Any group that has an optional ``depends_on`` field or any field
+        that has an optional ``depends_on`` attribute has a fallback when
+        ``depends_on`` is not provided. The fallback behavior involves
+        traversing up the hierarchy until the first ancestor that contains one
+        and only one ``NXcoordinate_system`` group is found. If such an ancestor
+        exists, its coordinate system applies. If none is found or more than
+        one instance of ``NXcoordinate_system`` is found at the same level,
+        then the current coordinate system is not defined with respect to
+        anything else. As an example, if there is only one ``NXcoordinate_system``
+        called"my_coordinate_system" defined directly under ``NXentry``, each
```
PeterC-DLS, line 49:
```suggestion
        called "my_coordinate_system" defined directly under ``NXentry``, each
```

base_classes/NXtransformations.nxdl.xml
PeterC-DLS: I think it's in our long-term interest to get this right in the first instance. The implications of applying the right hand rule in both CS handnesses needed more thought on its consequences for reading software.

I'm happy to approve this PR as long as the amendment is raised and discussed promptly. It should be fixed before the next release.

PeterC-DLS (approved):
LGTM

PR #1416, opened 2024-09-23
Fairmat 2024: use NXidentifier in NXuser

lukaspie:


base_classes/NXidentifier.nxdl.xml
```diff
@@ -0,0 +1,49 @@
...
+    </field>
+    <field name="identifier" type="NX_CHAR">
+        <doc>
+             The unique code, IRI or hash to resolve this reference.
+             Typically, this is stated by the service which is considered a complete
+             identifier, e.g., for a DOI it's something of the form `10.1107/S1600576714027575`
+             or `https://doi.org/10.1107/S1600576714027575`, which are both resolvable.
+        </doc>
+    </field>
+    <field name="is_persistent" type="NX_BOOLEAN">
```
paulmillar, line 43: I'm not sure what is the use-case for this field.

For many types of PIDs, the persistency is tacit: an ORCID is a persistent identifier, so why specify this?  Put another way, what does it mean to specify an NXidentifier with `service` "ORCID" and `is_persistent` of "false"?

I suppose the intention here is to allow ad-hoc values in "service" field and allow software to discover these NXidentifier entries are actually PIDs.  Perhaps, to allow software to discover that DOIs are persistent identifiers without that software knowing what are DOIs.

However, in most cases, persistency is a feature of the underlying identifier "service" (or "type").  Including this field seems like it is inviting inconsistent entries.

paulmillar (commented):
Just to be clear, I strongly support for NeXus including NXidentifier :-)

lukaspie:
Dear @paulmillar, thanks for the super valuable feedback. During [NIAC2024](https://www.nexusformat.org/NIAC2024.html), it was [decided](https://github.com/nexusformat/definitions/issues/1451#issuecomment-2382476355) (in a different discussion) that identifier is such a valuable concept that it should be possible to attach it to _any_ base class. Therefore, it will be added to NXobject here: https://github.com/nexusformat/definitions/pull/1486. I will close this PR, but consider your changes in the other PR (which is currently still an early draft).

PR #1423, opened 2024-09-24
Fairmat 2024: proposal on electron microscopy (EM)

lukaspie:
**Essence of changes**
* Refactored version for NXem including feedback from the EM community
* Refactoring of NXimage, NXspectrum base classes and examples for their usage
* Examples for documenting method-specific results with some contextualization for processing EBSD as an example for diffraction, EDS for spectroscopy, as well as images
* More complete representation of microscope components
* Connection to all concepts within [EMglossary](https://emglossary.helmholtz-metadaten.de/) community-agreed upon terminology set where it fits
* Backed by example data converted to NeXus files using NXem for at least one example for each microscope manufacturer and thousands of cases for EBSD as a specific technique
* Removal of reserved suffix _set within the name of base classes

**Dependencies of this PR on other PRs**
This PR depends heavily on others that make changes to existing base classes or introduce base classes that are to be used outside of APM as well. Thus, it will be much smaller once the discussion on these PRs is done and this particular branch here has been rebased:
#1410 (not used here anyway)
#1412 (but concepts not used in EM)
~~#1415~~
~~#1407~~ -> merged
~~#1408~~ -> merged
~~#1413~~ -> merged
~~#1414~~ -> merged
~~#1419~~ -> merged
~~#1420~~ -> merged
~~#1486~~ -> merged

**EM-specific base classes**
There are several new base classes that are currently only used in NXem, they are ordered in a sequence of dependencies to assure that reviewing adds building blocks gradually:

Conceptual modeling of aberrations that are relevant to assure the microscope is calibrated well enough to achive precision and accuracy and sharp images:
NXaberration
NXcorrector_cs

**Base classes for talking about thermodynamic phases and crystal structure as used in NXem_ebsd**
NXphase
NXunit_cell

**Base classes for concepts and/or components of an EM microscope**
NXebeam_column
NXevent_data_em
NXibeam_column
NXmanipulator (essentially the version from #1424)
NXoptical_system_em
NXscanbox_em

**Specialized NXprocess instances for modularized reporting of parameter, configs, and methods for common imaging modalities and types of NXdata collected, postprocessing**. 
NXem_ebsd (that is the most detailed example given the large number of example data in the public domain)
NXem_eds
NXem_eels
NXem_img

**Specializations of NXdata to not let NXem grow even larger** but allow to reduce the number of possible combinations how images and spectra will be named to work towards a more homogeneous presentation of n-dimensional hyperslab data:
NXimage, NXspectrum

**The appdef itself:**
[...]

PR #1420, opened 2024-09-23
Fairmat 2024: specialized base classes on top of NXprocess

lukaspie:
Adds several new base classes that all extend `NXprocess`:
- `NXcalibration`: for describing calibration of an instrument and/or a data set (e.g., an energy axis) -> moved from contributed to base_classes
- `NXdistortion`: post-processing distortion corrections
- `NXregistration`: [image registration](https://en.wikipedia.org/wiki/Image_registration)

PR #1419, opened 2024-09-23
FAIRmat 2024: additional base classes in NXinstrument

lukaspie:
This PR adds four unnamed base classes to `NXinstrument`.
- `NXactuator`: see description/discussion in #1414
- `NXfabrication`: see description/discussion in #1414  
- `NXhistory`: this is a new base class that is meant to describe the history of any physical object (in this case, the instrument history).  `NXhistory` is meant to contain many instances of the new base class `NXactivity`, which describes any physical activity that was performed on the physical entity prior or during the experiment. Note this is intentionally different than `NXprocess`, which describes processing of the _data_.
- `NXresolution`: This is meant to describe the resolution of a measureable physical quantity. Here, we are describing many different ways of a) how such a resolution can be described, and b) how such a resolution can be obtained. For b), we are also moving the base class `NXcalibration` (a subclass of `NXprocess`) from contributed to base_classes and slightly expand its former definition .

In addition, `NXnote` is expanded to contain more information about a given resource (like the URI of the resource,  the checksum of a file, etc.).

PR #1414, opened 2024-09-23
Fairmat 2024: actuators and sensors in NXenvironment

lukaspie:
This PR is aimed at further elaborating `NXenvironment`:
- A new base class `NXactuator` is added that closely resembles `NXsensor`, but instead of sensing something (like e.g. an `NXsensor` describing a thermocouple), it produces an actuation/output (think a sample heater for example).
- Unnamed `NXsensor` and `NXactuator` are added to `NXenvironment`. In addtion, a `value` field is added to `NXenvironment`, which shall only be used when there is no ctuator/sensor that controls/measures the environment parameters, but the user would still like to give a value for it. An example would be a measured at room temperature where the temperature is only estimated.
- To describe the hardware of such sensors and actuators (or indeed any other instrument parts), a new base class `NXfabrication` was added, containing information about the vendor, model, serial number, etc.
- ~~The base class `NXpid` was moved from `contributed` to `base_classes`, since it is used in `NXactuator` and accepted base classes should (as far as we understand) only make use of base classes that are themselves also accepted (see discussion below).~~ **EDIT**: `NXpid` to be handled in #1522.

PR #1424, opened 2024-09-24
Fairmat 2024: proposal on photoemission spectroscopy (MPES)

lukaspie:


rayosborn:
This PR wants to create more than a dozen new base classes, along with major changes to several more, and yet there is minimal description of why they are needed and the links to other issues or other PRs provide little additional information. IMHO, it is impossible for the NIAC to assess such a huge PR as a single entity. I would suggest that there needs to be a separate PR for each base class, new or modified, containing a detailed explanation of why the PR is needed. This will slow the process down considerably, so at the next Telco I would suggest we discuss how to handle such a backlog, e.g., by assigning a couple of members to review each one and report back with comments so that we can have an expedited discussion and vote. However, I would recommend withdrawing this PR until we have decided how to proceed.

lukaspie:
> This PR wants to create more than a dozen new base classes, along with major changes to several more, and yet there is minimal description of why they are needed and the links to other issues or other PRs provide little additional information. IMHO, it is impossible for the NIAC to assess such a huge PR as a single entity. I would suggest that there needs to be a separate PR for each base class, new or modified, containing a detailed explanation of why the PR is needed. This will slow the process down considerably, so at the next Telco I would suggest we discuss how to handle such a backlog, e.g., by assigning a couple of members to review each one and report back with comments so that we can have an expedited discussion and vote. However, I would recommend withdrawing this PR until we have decided how to proceed.

@rayosborn Thanks for the feedback. Indeed, there was a lack of description in this PR, so I added a lengthy one. I also tried to describe that this PR, while having many changes as of right now, essentially comes down to the three mentioned application definitions (two of which are only specializing the other one) + a set of base classes that is essentially needed for describing these kinds of experiments. Somewhat separate from that, it also introduced classes for describing fits to data (`NXfit` etc.). All of these base classes are heavily used within the application definitions, so I don't really see how they can be separated from that discussion here. 

I agree there's a lot of changes in this PR and it can be hard to understand what is happening/changing when trying to review it. However, note that all of the other changes (as discussed above) are already discussed elsewhere. They were brought in here for the reason that the new base classes and application definitions heavily depend on them and it would be impossible to understand these without having the other changes alongside them. Eventually, before merging, this PR can be much smaller once all of the other PRs have been finalized.

What I don't see is why the changes that are proposed here cannot be discussed now in this PR. I agree that it would make sense for a couple of NIAC members to do so, but I can't see how to make the changes here any smaller than they are right now (aside from removing everything that is discussed elsewhere). We are open to whatever is discussed in the next Telco though, but until then, I will leave this PR open.

phyy-nx:
@lukaspie thanks for the description and your reply.

> What I don't see is why the changes that are proposed here cannot be discussed now in this PR.

It's simply because the other unmerged PRs (https://github.com/nexusformat/definitions/pull/1408, https://github.com/nexusformat/definitions/pull/1409, https://github.com/nexusformat/definitions/pull/1410, https://github.com/nexusformat/definitions/pull/1413, https://github.com/nexusformat/definitions/pull/1414, https://github.com/nexusformat/definitions/pull/1415, https://github.com/nexusformat/definitions/pull/1419, https://github.com/nexusformat/definitions/pull/1420) make this one too hard to read, that's all.  Providing that list was incredibly helpful, thank you.  I anticipate once those are merged and this is rebased, the review will be greatly expedited!

phyy-nx:
Re-posting the TODO list including what I think are the dependencies:

#1408, #1409, #1410, #1413 --> #1419 and #1486, #1414, #1415, #1419  --> #1414 and #1486,  #1420 --> #1419

phyy-nx:
From Telco:
- Rename reference to target
- Remove dependency on #1410 because specifying a target attribute is already an existing feature


There are three levels of review the NIAC could offer:

1. Typographic
2. Technical (does the PR conform to the NeXus spec)
3. Scientific (subject expert required)

FairMAT has already gathered expert review so in that way 3) is technically been done through workshops with experts.  Therefore our (the NIAC's job) really should just be 1. and 2.

lukaspie:
@PeterC-DLS @fsaizpoy I have made all changes that were necessary due to changes in the accepted base classes in this PR. Note that this PR still has changes related to #1410 and #1415, which are yet to be resolved there. The rest should be reviewable now.

base_classes/NXdeflector.nxdl.xml
```diff
@@ -21,13 +21,13 @@
 #
 # For further information, see http://www.nexusformat.org
 -->
-<definition xmlns="http://definition.nexusformat.org/nxdl/3.1" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" category="base" name="NXdeflector" extends="NXcomponent" type="group" xsi:schemaLocation="http://definition.nexusformat.org/nxdl/3.1 ../nxdl.xsd">
+<definition xmlns="http://definition.nexusformat.org/nxdl/3.1" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" category="base" type="group" name="NXdeflector" extends="NXcomponent" xsi:schemaLocation="http://definition.nexusformat.org/nxdl/3.1 ../nxdl.xsd">
     <doc>
-         Deflectors as they are used e.g. in an electron analyser.
+        Deflectors as they are used e.g. in an electron analyzer.
```
PeterC-DLS, line 26: Component of an electron analyzer that deflects the paths of electrons

PeterC-DLS: Could this be using magnetic or electrostatic fields?

base_classes/NXelectronanalyzer.nxdl.xml
```diff
@@ -0,0 +1,279 @@
...
+# For further information, see http://www.nexusformat.org
+-->
+<definition xmlns="http://definition.nexusformat.org/nxdl/3.1" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" category="base" type="group" name="NXelectronanalyzer" extends="NXcomponent" xsi:schemaLocation="http://definition.nexusformat.org/nxdl/3.1 ../nxdl.xsd">
+    <symbols>
+        <doc>
+            The symbols used in the schema to specify e.g. dimensions of arrays
+        </doc>
+        <symbol name="nfa">
+            <doc>
+                Number of fast axes (axes acquired simultaneously, without scanning a
```
PeterC-DLS, line 31: Is this the rank of the sensor's data?

base_classes/NXelectronanalyzer.nxdl.xml
```diff
@@ -0,0 +1,279 @@
...
+            </doc>
+        </symbol>
+        <symbol name="n_transmission_function">
+            <doc>
+                Number of data points in the transmission function.
+            </doc>
+        </symbol>
+    </symbols>
+    <doc>
+        Basic class for describing a electron analyzer.
```
PeterC-DLS, line 47:
```suggestion
        Basic class for describing an electron analyzer.
```

base_classes/NXelectronanalyzer.nxdl.xml
```diff
@@ -0,0 +1,279 @@
...
+            a photoelectron in PES is therefore given by
+            :math:`E_K^{\mathrm{meas.}} = h\nu - E_B + \Delta \phi = h\nu - E_B - e \phi_{\mathrm{spectr.}}`.
+            As a result, the measured kinetic energy :math:`E_K^{\mathrm{meas.}}` of a photoelectron is `independent`
+            of the sample work function. Nonetheless, the work function :math:`\phi_s` needs to be known to
+            accurately determine the binding energy scale.
+        </doc>
+    </field>
+    <field name="voltage_energy_range" type="NX_FLOAT" units="NX_VOLTAGE">
+        <doc>
+            Energy range of the voltage supplies. This influences the noise of the supplies,
```
PeterC-DLS, line 93:
```suggestion
            Energy range of the voltage supply. This influences the noise of the supply
```

base_classes/NXelectronanalyzer.nxdl.xml
```diff
@@ -0,0 +1,279 @@
...
+    <field name="fast_axes" type="NX_CHAR">
+        <doc>
+            List of the axes that are acquired simultaneously by the detector.
+            These refer only to the experimental variables recorded by the electron analyzer.
+            Other variables such as temperature, manipulator angles etc. are labeled as fast or slow in the data.
+            
+            .. csv-table:: Examples
+              :header: "Mode", "fast_axes", "slow_axes"
+            
+              Hemispherical in ARPES mode,  "['energy', 'kx']",""
```
PeterC-DLS, line 177:
```suggestion
              "Hemispherical in ARPES mode",  "['energy', 'kx']",""
```

base_classes/NXelectronanalyzer.nxdl.xml
```diff
@@ -0,0 +1,279 @@
...
+            <enumeration>
+                <item value="length"/>
+            </enumeration>
+        </field>
+        <field name="resolution" type="NX_FLOAT" units="NX_LENGTH"/>
+        <field name="resolution_errors" type="NX_FLOAT" units="NX_LENGTH"/>
+    </group>
+    <field name="fast_axes" type="NX_CHAR">
+        <doc>
+            List of the axes that are acquired simultaneously by the detector.
```
PeterC-DLS, line 170: Is the order significant?

base_classes/NXelectronanalyzer.nxdl.xml
```diff
@@ -0,0 +1,279 @@
...
+            
+            The transmission function (TF) specifies the detection efficiency per solid angle for electrons of
+            different kinetic energy passing through the electron analyzer. It depends on the spectrometer
+            geometry as well as operation settings such as lens mode and pass energy.
+            The transmission function is usually given as relative intensity vs. kinetic energy.
+            
+            The TF is used for calibration of the intensity scale in quantitative XPS. Without proper
+            transmission correction, a comparison of results measured from the same sample using different
+            operating modes for an instrument would show significant variations in atomic
+            concentrations.
```
PeterC-DLS, line 210: Is atomic concentration what this analyzer measure?

base_classes/NXenergydispersion.nxdl.xml
```diff
@@ -0,0 +1,177 @@
...
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+#
+# For further information, see http://www.nexusformat.org
+-->
+<definition xmlns="http://definition.nexusformat.org/nxdl/3.1" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" category="base" type="group" name="NXenergydispersion" extends="NXcomponent" xsi:schemaLocation="http://definition.nexusformat.org/nxdl/3.1 ../nxdl.xsd">
+    <doc>
+        Energy dispersion section of a photoelectron analyzer.
```
PeterC-DLS, line 26:
```suggestion
        Energy dispersion section of an electron analyzer.
```

base_classes/NXenergydispersion.nxdl.xml
```diff
@@ -0,0 +1,177 @@
...
+    </doc>
+    <field name="scheme" type="NX_CHAR">
+        <doc>
+            Energy dispersion scheme employed, for example: tof, hemispherical, cylindrical,
+            mirror, retarding grid, etc.
+        </doc>
+    </field>
+    <field name="pass_energy" type="NX_FLOAT" units="NX_ENERGY">
+        <doc>
+            Mean kinetic energy of the electrons in the energy-dispersive portion of the analyzer.
```
PeterC-DLS, line 36:
```suggestion
            Mean kinetic energy of the electrons in the energy-dispersive section of the analyzer.
```

base_classes/NXenergydispersion.nxdl.xml
```diff
@@ -0,0 +1,177 @@
...
+            This term should be used for hemispherical analyzers.
+            
+            This concept is related to term `12.63`_ of the ISO 18115-1:2023 standard.
+            
+            .. _12.63: https://www.iso.org/obp/ui/en/#iso:std:iso:18115:-1:ed-3:v1:en:term:12.63
+        </doc>
+    </field>
+    <field name="kinetic_energy" type="NX_FLOAT" units="NX_ENERGY">
+        <doc>
+            Kinetic energy set for the dispersive analyzer section. Can be either the set kinetic energy,
```
PeterC-DLS, line 46:
```suggestion
            Kinetic energy set for this dispersive section. Can be either the set kinetic energy,
```

base_classes/NXenergydispersion.nxdl.xml
```diff
@@ -0,0 +1,177 @@
...
+        <doc>
+            Center of the energy window
+        </doc>
+    </field>
+    <field name="energy_interval" type="NX_FLOAT" units="NX_ENERGY">
+        <doc>
+            The interval of transmitted energies. It can be two different things depending
+            on whether the scan is fixed or swept. With a fixed scan it is a 2 vector
+            containing the extrema of the transmitted energy window (smaller number first).
+            With a swept scan of m steps it is a 2xm array of windows one for each
```
PeterC-DLS, line 65:
```suggestion
            With a swept scan of m steps it is a 2xm array of windows, one for each
```

PeterC-DLS: Also make `m` a math symbol

base_classes/NXenergydispersion.nxdl.xml
```diff
@@ -0,0 +1,177 @@
...
+    <field name="energy_scan_mode" type="NX_CHAR">
+        <doc>
+            Way of scanning the energy axis
+        </doc>
+        <enumeration open="true">
+            <item value="fixed_analyzer_transmission">
+                <doc>
+                    constant :math:`\Delta E` mode, where the electron retardation (i.e., the fraction of pass energy to
+                    kinetic energy, :math:`R = (E_K - W/E_p)`, is scanned, but the pass energy :math:`E_p` is kept constant.
+                    Here, :math:`W = e \phi` is the spectrometer work function (with the potential difference :math:`\phi_{\mathrm{sample}}`
```
PeterC-DLS, line 88: Use same phi symbol

base_classes/NXfit.nxdl.xml
```diff
@@ -0,0 +1,199 @@
...
+                    as the ``input_independent`` field. Each individual dimension of ``residual``
+                    must have the same number of points as the corresponding dimension in
+                    the ``input_independent`` field.
+                </doc>
+            </dimensions>
+        </field>
+    </group>
+    <group name="peakPEAK" type="NXpeak" nameType="partial">
+        <doc>
+            One peak of the peak model.
```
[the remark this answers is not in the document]
lukaspie: ToDo: rename to "fit model"

base_classes/NXdeflector.nxdl.xml
lukaspie: implemented

base_classes/NXelectronanalyzer.nxdl.xml
lukaspie: You are right. Replaced by "would show significant variations in signal intensity for the same kinetic energies."

base_classes/NXcollectioncolumn.nxdl.xml
```diff
@@ -21,66 +21,80 @@
...
+            Deflectors in the collection column section
         </doc>
     </group>
     <group type="NXlens_em">
         <doc>
-             Individual lenses in the collection column section
+            Individual lenses in the collection column section
         </doc>
     </group>
+    <group type="NXfabrication"/>
```
PeterC-DLS, line 99: Not needed as it's a component

base_classes/NXelectronanalyzer.nxdl.xml
```diff
@@ -0,0 +1,279 @@
...
+            :math:`E_B` below the Fermi level is given by :math:`E_K = h\nu - E_B - W = h\nu - E_B - e \phi_{\mathrm{sample}}`.
+            Here, :math:`W = e \phi_{\mathrm{sample}}` is the work function of the sample surface (which is directly proportional to the potential difference :math:`\phi_{\mathrm{sample}}`
+            between the electrochemical potential of electrons in the bulk and the electrostatic potential of an electron in the vacuum just outside the surface).
+            
+            In PES measurements, the sample and the spectrometer (with work function :math:`\phi_{\mathrm{spectr.}}`)
+            are electrically connected and therefore their Fermi levels are aligned. Due to the difference in local
+            vacuum level between the sample and spectrometer, there exists an electric potential difference (contact potential)
+            :math:`\Delta\phi = \phi_{\mathrm{sample}} - \phi_{\mathrm{spectr.}}`. The measured kinetic energy of
+            a photoelectron in PES is therefore given by
+            :math:`E_K^{\mathrm{meas.}} = h\nu - E_B + \Delta \phi = h\nu - E_B - e \phi_{\mathrm{spectr.}}`.
```
PeterC-DLS, line 85: The last equality is not true if you substitute Delta phi literally

applications/NXmpes.nxdl.xml
```diff
@@ -0,0 +1,1456 @@
...
+            <attribute name="version"/>
+            <enumeration>
+                <item value="NXmpes"/>
+            </enumeration>
+        </field>
+        <field name="title"/>
+        <field name="start_time" type="NX_DATE_TIME">
+            <doc>
+                Datetime of the start of the measurement.
+                Should be a ISO8601 date/time stamp. It is recommended to add an explicit time zone,
```
PeterC-DLS, line 56:
```suggestion
                Should be an ISO8601 date/time stamp. It is recommended to add an explicit time zone,
```

applications/NXmpes.nxdl.xml
```diff
@@ -0,0 +1,1456 @@
...
+                    * spin-resolved angle-resolved X-ray photoelectron spectroscopy (spin-ARPES)
+                    * momentum microscopy
+                
+                .. _ISO 18115-1:2023: https://www.iso.org/standard/74811.html
+                .. _Clause 11: https://www.iso.org/obp/ui/en/#iso:std:iso:18115:-1:ed-3:v1:en:sec:11
+            </doc>
+        </field>
+        <field name="transitions" optional="true">
+            <doc>
+                List of strings representing the electronic core levels and Auger transitions probed in this
```
PeterC-DLS, line 93:
```suggestion
                Array of strings representing the electronic core levels and Auger transitions probed in this
```

applications/NXmpes.nxdl.xml
```diff
@@ -0,0 +1,1456 @@
...
+                <field name="probe" optional="true"/>
+                <group name="device_information" type="NXfabrication" recommended="true">
+                    <field name="vendor" recommended="true"/>
+                    <field name="model" recommended="true"/>
+                    <field name="identifier" recommended="true"/>
+                </group>
+                <field name="associated_beam" type="NX_CHAR">
+                    <doc>
+                        A reference to a beam emitted by this source.
+                        Should be named with the same appendix, e.g.,
```
PeterC-DLS, line 282:
```suggestion
                        Should be named with the same suffix, e.g.,
```

PeterC-DLS: Here and below

applications/NXmpes.nxdl.xml
```diff
@@ -0,0 +1,1456 @@
...
+                            additional or other apertures use the APERTURE group of NXcollectioncolumn.
+                        </doc>
+                    </group>
+                    <group name="contrast_aperture" type="NXaperture" optional="true">
+                        <doc>
+                            The size and position of the contrast aperture inserted in the column. To add
+                            additional or other apertures use the APERTURE group of NXcollectioncolumn.
+                        </doc>
+                    </group>
+                    <group name="iris" type="NXaperture" optional="true">
```
PeterC-DLS, line 475: Usually called an entrance pupil

applications/NXmpes.nxdl.xml
```diff
@@ -0,0 +1,1456 @@
...
+                            </enumeration>
+                        </attribute>
+                        <field name="raw" type="NX_NUMBER">
+                            <doc>
+                                Raw data before calibration.
+                            </doc>
+                        </field>
+                        <field name="pixel_x" type="NX_NUMBER" optional="true" units="NX_ANY">
+                            <doc>
+                                Detector pixel in x direction.
```
PeterC-DLS, line 588:
```suggestion
                                Detector pixel size in x direction.
```

PeterC-DLS: How does this differ from "x_pixel_size" that exists in the parent base class?

applications/NXmpes.nxdl.xml
```diff
@@ -0,0 +1,1456 @@
...
+                            </doc>
+                        </field>
+                        <field name="k_perpendicular" type="NX_NUMBER" optional="true" units="NX_WAVENUMBER">
+                            <doc>
+                                (Un)calibrated perpendicular component in k-space.
+                                
+                                ``k_perpendicular`` is the component that is normal (perpendicular) to the surface.
+                                It is not conserved during photoemission because the electron experiences a potential
+                                change when it exits the  material into vacuum. To determine ``k_perpendicular``
+                                inside the material, one typically needs to estimate the inner potential :math:`V_0`,
```
PeterC-DLS, line 683: Match with phi used elsewhere?

applications/NXmpes.nxdl.xml
```diff
@@ -0,0 +1,1456 @@
...
+                            </doc>
+                        </field>
+                        <field name="angular0" type="NX_NUMBER" optional="true" units="NX_ANGLE">
+                            <doc>
+                                First (un)calibrated angular coordinate. It is envisioned that the axes in angular space
+                                are named ``angular0`` and ``angular1``.
+                                
+                                The angular axes should be named in order of decreasing speed, i.e., ``angular0``
+                                should be the fastest scan axis and ``angular1`` should be the slow-axis angular
+                                coordinate. However, ``angular0`` may also be second slow axis if the measurement
```
PeterC-DLS, line 697: Do you mean second slowest and second fastest?

applications/NXmpes.nxdl.xml
```diff
@@ -0,0 +1,1456 @@
...
+                                For more information, see the
+                                definition of the :ref:`angular0 &lt;/NXmpes/ENTRY/INSTRUMENT/ELECTRONANALYZER/ELECTRON_DETECTOR/raw_data/angular0-field&gt;`
+                                axis.
+                                
+                                This is typically the slower scan axis compared to ``angular0``.
+                            </doc>
+                        </field>
+                        <field name="spatial0" type="NX_NUMBER" optional="true" units="NX_LENGTH">
+                            <doc>
+                                First (un)calibrated spatial coordinate. It is envisioned that the axes in angular space
```
PeterC-DLS, line 715:
```suggestion
                                First (un)calibrated spatial coordinate. It is envisioned that the axes in regular space
```
perhaps, not angular

applications/NXmpes.nxdl.xml
```diff
@@ -0,0 +1,1456 @@
...
+                                For more information, see the
+                                definition of the :ref:`spatial0 &lt;/NXmpes/ENTRY/INSTRUMENT/ELECTRONANALYZER/ELECTRON_DETECTOR/raw_data/spatial0-field&gt;`
+                                axis.
+                                
+                                This is typically the slower scan axis compared to ``spatial0``.
+                            </doc>
+                        </field>
+                        <field name="delay" type="NX_NUMBER" optional="true" units="NX_TIME">
+                            <doc>
+                                (Un)calibrated delay time.
```
PeterC-DLS, line 738: relative to pump start?

applications/NXmpes.nxdl.xml
```diff
@@ -0,0 +1,1456 @@
...
+                Any fit procedures.
+            </doc>
+        </group>
+        <group type="NXsample">
+            <field name="name"/>
+            <field name="identifier" recommended="true"/>
+            <field name="chemical_formula" recommended="true"/>
+            <field name="atom_types" recommended="true">
+                <doc>
+                    List of comma-separated elements from the periodic table
```
PeterC-DLS, line 1074: Is that better as an array?

applications/NXmpes.nxdl.xml
```diff
@@ -0,0 +1,1456 @@
...
+                    <field name="end_time" type="NX_DATE_TIME" recommended="true"/>
+                    <field name="method" recommended="true">
+                        <doc>
+                            Details about the method of sample preparation before the photoemission
+                            experiment.
+                        </doc>
+                    </field>
+                </group>
+            </group>
+            <group name="temperature_env" type="NXenvironment" recommended="true">
```
PeterC-DLS, line 1109: Should this be an `NXmanipulator`?

applications/NXmpes_arpes.nxdl.xml
```diff
@@ -0,0 +1,407 @@
...
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+#
+# For further information, see http://www.nexusformat.org
+-->
+<definition xmlns="http://definition.nexusformat.org/nxdl/3.1" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" category="application" type="group" name="NXmpes_arpes" extends="NXmpes" xsi:schemaLocation="http://definition.nexusformat.org/nxdl/3.1 ../nxdl.xsd">
+    <doc>
+        This is an general application definition for angle-resolved multidimensional
```
PeterC-DLS, line 26:
```suggestion
        This is an general application definition for angle-resolved (multidimensional)
```

applications/NXmpes_arpes.nxdl.xml
```diff
@@ -0,0 +1,407 @@
...
+        photoelectron spectroscopy.
```
PeterC-DLS, line 27:
```suggestion
        photoelectron spectroscopy (ARPES).
```

manual/source/classes/contributed_definitions/mpes-structure.rst
```diff
@@ -35,8 +35,8 @@ Application Definitions
 Base Classes
```
PeterC-DLS, line 35: Add other application definitions above (`NXxps` and `NXmpes_arpes`)

applications/NXmpes.nxdl.xml
```diff
@@ -0,0 +1,1456 @@
...
+                        Note that this method for recording the bias is not advised, but using
+                        NXsensor and NXactuator is strongly recommended instead.
+                    </doc>
+                </field>
+            </group>
+            <group name="drain_current_env" type="NXenvironment" optional="true">
+                <doc>
+                    Drain current of the sample and sample holder.
+                </doc>
+                <group name="amperemeter" type="NXsensor" recommended="true">
```
PeterC-DLS, line 1215:
```suggestion
                <group name="ammeter" type="NXsensor" recommended="true">
```

PeterC-DLS: and below

applications/NXmpes_arpes.nxdl.xml
```diff
@@ -0,0 +1,407 @@
...
+            <field name="depends_on" type="NX_CHAR">
+                <doc>
+                    Link to transformations defining an APRES base coordinate system,
+                    which is defined such that the positive z-axis points towards the analyzer entry,
+                    and the x-axis lies within the beam/analyzer plane.
+                </doc>
+            </field>
+            <group type="NXtransformations">
+                <doc>
+                    Set of transformations, describing the orientation of the ARPES coordinate system
```
PeterC-DLS, line 46:
```suggestion
                    A set of transformations, describing the orientation of the ARPES coordinate system
```

applications/NXmpes_arpes.nxdl.xml
```diff
@@ -0,0 +1,407 @@
...
+                        </enumeration>
+                    </attribute>
+                    <attribute name="vector" type="NX_NUMBER">
+                        <enumeration>
+                            <item value="[0, 0, 1]"/>
+                        </enumeration>
+                    </attribute>
+                    <attribute name="depends_on">
+                        <enumeration>
+                            <item value="offset_tilt"/>
```
PeterC-DLS, line 301: If you're starting the chain with `sample_azimuthal` (L196) then I would expect `offset_azimuthal`, `sample_tilt`, `offset_tilt`, `sample_polar` and `offset_polar` - i.e. offsets are applied first in an active transformation.

applications/NXmpes_arpes.nxdl.xml
```diff
@@ -0,0 +1,407 @@
...
+                            <item value="sample_azimuth"/>
+                        </enumeration>
+                    </attribute>
+                </field>
+            </group>
+        </group>
+        <group name="data" type="NXdata">
+            <attribute name="signal">
+                <doc>
+                    There is an object named data that contains the signal.
```
PeterC-DLS, line 330:
```suggestion
                    There is a field named data that contains the signal.
```

applications/NXmpes_arpes.nxdl.xml
```diff
@@ -0,0 +1,407 @@
...
+                <enumeration>
+                    <item value="['angular0', 'angular1', 'energy']"/>
+                </enumeration>
+            </attribute>
+            <attribute name="energy_indices" type="NX_INT"/>
+            <attribute name="angular0_indices" type="NX_INT"/>
+            <attribute name="angular1_indices" type="NX_INT"/>
+            <field name="energy" type="NX_NUMBER" units="NX_ENERGY">
+                <doc>
+                    Trace of the energy axis. Could be linked from the respective ``@reference``
```
PeterC-DLS, line 350:
```suggestion
                    Values on the energy axis. Could be linked from the respective ``@reference``
```

applications/NXmpes_arpes.nxdl.xml
```diff
@@ -0,0 +1,407 @@
...
+                    field.
+                </doc>
+                <attribute name="reference" recommended="true">
```
PeterC-DLS, line 353: I think we are settling on `@target`

applications/NXmpes_arpes.nxdl.xml
```diff
@@ -0,0 +1,407 @@
...
+                          @reference: '/entry/sample/transformations/sample_polar' for a manipulator angular scan
+                          @reference: '/entry/instrument/detector/sensor_y' for a 2D detector
+                          @reference: '/entry/instrument/collectioncolumn/deflector' for a deflector scan
+                          @reference: '/entry/process/calibration/angular1_calibration/calibrated_axis' for a preprocessed axis.
+                    </doc>
+                </attribute>
+            </field>
+            <field name="data" type="NX_NUMBER" units="NX_ANY">
+                <doc>
+                    Represents a measure of three-dimensional photoemission counts, where
```
PeterC-DLS, line 400:
```suggestion
                    Represents a measurement of three-dimensional photoemission counts, where
```
```suggestion
                    Represents a measure of three-dimensional photoemission counts, where
```

applications/NXxps.nxdl.xml
```diff
@@ -0,0 +1,732 @@
...
+            </enumeration>
+        </field>
+        <field name="method">
+            <doc>
+                A name of the experimental method according to `Clause 11`_ of
+                the `ISO 18115-1:2023`_ specification.
+                
+                Examples for XPS-related experiments include:
+                    * X-ray photoelectron spectroscopy (XPS)
+                    * angle-resolved X-ray photoelectron spectroscopy (ARXPS)
```
PeterC-DLS, line 41: How does this differ from ARPES?

applications/NXxps.nxdl.xml
```diff
@@ -0,0 +1,732 @@
...
+                            @depends_on=y_flip
+                            @transformation_type=rotation
+                            @vector=[0, 0, 1]
+                            @units=degree
+                            y_flip=180
+                            @depends_on=y_rotation
+                            @transformation_type=rotation
+                            @vector=[0, 1, 0]
+                            @units=degree
+                            y_rotation=beam_polar_angle_of_incidence
```
PeterC-DLS, line 126: Should this be `-polar_angle` so 1st net transformation is `R_y(pi - polar_angle)`

applications/NXxps.nxdl.xml
```diff
@@ -0,0 +1,732 @@
...
+                    If the z-axis is pointing in the direction of gravity (i.e., if the sample is mounted horizontally),
+                    the following transformations can be used for describing the XPS coordinate system
+                    with respect to the beam coordinate system (.):
+                    
+                    .. code-block::
+                    
+                        xps_coordinate_system:NXcoordinate_system
+                        depends_on=entry/geometries/xps_coordinate_system/coordinate_transformations/z_rotation
+                        coordinate_system_transformations:NXtransformations
+                            z_rotation=beam_azimuth_angle
```
PeterC-DLS, line 116: I think this azimuthal angle is relative to the `x`-axis so 2nd transformation is `R_z(azimuthal_angle)`

applications/NXxps.nxdl.xml
```diff
@@ -0,0 +1,732 @@
...
+                
+                .. _12.58: https://www.iso.org/obp/ui/en/#iso:std:iso:18115:-1:ed-3:v1:en:term:12.58
+            </doc>
+            <group name="source_probe" type="NXsource" recommended="true">
+                <field name="power" type="NX_FLOAT" units="NX_POWER" recommended="true"/>
+            </group>
+            <group name="beam_probe" type="NXbeam">
+                <field name="depends_on" recommended="true">
+                    <doc>
+                        Reference to the transformation describing the orientation of the beam
```
PeterC-DLS, line 148: This should point to `beam_polar_angle_of_incidence`

applications/NXxps.nxdl.xml
```diff
@@ -0,0 +1,732 @@
...
+                            the sample stage.
+                        </doc>
+                        <attribute name="transformation_type">
+                            <enumeration>
+                                <item value="rotation"/>
+                            </enumeration>
+                        </attribute>
+                        <attribute name="vector" type="NX_NUMBER">
+                            <enumeration>
+                                <item value="[0, -1, 0]"/>
```
PeterC-DLS, line 165: Does not match the entry's `NXtransformations` documentation

applications/NXxps.nxdl.xml
```diff
@@ -0,0 +1,732 @@
...
+                        </attribute>
+                        <attribute name="depends_on">
+                            <enumeration>
+                                <item value="beam_azimuth_angle"/>
+                            </enumeration>
+                        </attribute>
+                    </field>
+                    <field name="beam_azimuth_angle" type="NX_NUMBER" units="NX_ANGLE">
+                        <doc>
+                            Azimuthal rotation of the beam from the y-direction towards the operator, defined
```
PeterC-DLS, line 176: What operator? Where are they standing?

applications/NXxps.nxdl.xml
```diff
@@ -0,0 +1,732 @@
...
+                            The user is encouraged to use one of the options defined in the enumeration, but in case none of these fit
+                            (e.g., in the case of very complex line shapes), a different value for the ``function_type`` field
+                            can be used. In that case in particular, but also if one of the suggested values is used, the functional
+                            form of the peak should be given by the ``formula_description`` field. The user is also encouraged to use the
+                            ``description`` field for describing the fit function in a human-readable way.
+                        </doc>
+                        <enumeration open="true">
+                            <item value="Gaussian">
+                                <doc>
+                                    The Gaussian function models peaks with a symmetric shape, commonly arising from instrumental broadening
```
PeterC-DLS, line 333: It seems that there needs to be a formal way to map the parameters in these formulae to `fit_parameters`

applications/NXxps.nxdl.xml
```diff
@@ -0,0 +1,732 @@
...
+            <field name="depends_on" recommended="true">
+                <doc>
+                    Reference to the transformation describing the orientation of the sample
+                    relative to a defined coordinate system.
+                </doc>
+            </field>
+            <group name="transformations" type="NXtransformations" recommended="true">
+                <field name="sample_rotation_angle" type="NX_NUMBER" units="NX_ANGLE">
+                    <doc>
+                        Clockwise rotation about the sample normal.
```
PeterC-DLS, line 662: What is clockwise in a left-handed coordinate system?

base_classes/NXdata.nxdl.xml
```diff
@@ -309,6 +309,20 @@
 
 		</doc>
 	</attribute>
+	<attribute name="target">
```
PeterC-DLS, line 312: Why have this attribute attached to the group in addition to the one for `DATA`?

base_classes/NXcoordinate_system.nxdl.xml
```diff
@@ -0,0 +1,159 @@
...
+# Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+#
+# For further information, see http://www.nexusformat.org
+-->
+<definition xmlns="http://definition.nexusformat.org/nxdl/3.1" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" category="base" type="group" name="NXcoordinate_system" extends="NXobject" xsi:schemaLocation="http://definition.nexusformat.org/nxdl/3.1 ../nxdl.xsd">
+    <doc>
+        Base class to detail a coordinate system (CS).
+        
+        Whenever possible, an instance of :ref:`NXcoordinate_system` should be used as
+        a member in an :ref:`NXcoordinate_system_set` and the name of the instance
```
PeterC-DLS, line 29: Out of date reference to `NXcoordinate_system_set`?

base_classes/NXcoordinate_system.nxdl.xml
```diff
@@ -0,0 +1,159 @@
...
+            Coordinate system type.
+        </doc>
+        <enumeration>
+            <item value="undefined"/>
+            <item value="cartesian"/>
+        </enumeration>
+    </field>
+    <field name="handedness" type="NX_CHAR">
+        <doc>
+            Handedness of the coordinate system if it is a Cartesian.
```
PeterC-DLS, line 66:
```suggestion
            Handedness of the coordinate system if it is Cartesian.
```

base_classes/NXcoordinate_system.nxdl.xml
```diff
@@ -0,0 +1,159 @@
...
+        </doc>
+        <enumeration>
+            <item value="undefined"/>
+            <item value="right_handed"/>
+            <item value="left_handed"/>
+        </enumeration>
+    </field>
+    <field name="x_alias" type="NX_CHAR">
+        <doc>
+            Possibility to define an alias for the name of the x-axis.
```
PeterC-DLS, line 76:
```suggestion
            Opportunity to define an alias for the name of the x-axis.
```

base_classes/NXcoordinate_system.nxdl.xml
```diff
@@ -0,0 +1,159 @@
...
+        </doc>
+    </field>
+    <field name="x_direction" type="NX_CHAR">
+        <doc>
+            Human-readable field telling in which direction the x-axis points if that
+            instance of :ref:`NXcoordinate_system` has no reference to any parent and as such
+            is the mighty world reference frame.
```
PeterC-DLS, line 83:
```suggestion
            is the world reference frame.
```

base_classes/NXcoordinate_system.nxdl.xml
```diff
@@ -0,0 +1,159 @@
...
+            This axis is frequently referred to as the z-axis in real space and
+            the k-axis in reciprocal space.
+        </doc>
+        <dimensions rank="1">
+            <dim index="1" value="3"/>
+        </dimensions>
+    </field>
+    <field name="depends_on" type="NX_CHAR">
+        <doc>
+            This specificies the relation to another coordinate system by pointing to the last
```
PeterC-DLS, line 149:
```suggestion
            This specifies the relation to another coordinate system by pointing to the last
```

applications/NXxps.nxdl.xml
lukaspie: In [ARXPS](https://www.thermofisher.com/de/de/home/materials-science/learning-center/surface-analysis/angle-resolved-xps.html), you just change the sample-detector angle to be more or less grazing, but you don't actually resolve in k/angular space. So it's named rather similar, but these are completely different technique.

base_classes/NXcollectioncolumn.nxdl.xml
```diff
@@ -1,10 +1,10 @@
-<?xml version="1.0" encoding="UTF-8"?>
+<?xml version='1.0' encoding='UTF-8'?>
```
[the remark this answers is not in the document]
lukaspie: This is just the way that our [`nyaml`](https://github.com/FAIRmat-NFDI/nyaml) tool auto-generates the XML file based on the YAML file where we do the actual editing. If you prefer double quote, we can change that of course.

base_classes/NXmanipulator.nxdl.xml
```diff
@@ -0,0 +1,224 @@
...
+        <group name="value_log" type="NXlog">
+            <field name="value" type="NX_FLOAT" units="NX_TEMPERATURE">
+                <doc>
+                    In the case of an experiment in which the temperature changes and is recorded with time stamps,
+                    this is an array of length m of temperatures.
+                </doc>
+            </field>
+        </group>
+    </group>
+    <group name="sample_heater" type="NXactuator">
```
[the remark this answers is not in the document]
lukaspie: In our case, this is a cryostat that cools the sample _and_ the whole manipulator, so therefore we ommitted the ´sample_` prefix here (whereas the heater is really only heating the sample).

manual/source/classes/contributed_definitions/mpes-structure.rst
lukaspie: I added all application definitions and base classes in this file. We were planning to bring this to `manual/source/classes/applications/mpes-structure.rst` and `manual/source/classes/base_classes/mpes-structure.rst` at some point (see #1428), but this can be handled in that linked PR.

applications/NXmpes_arpes.nxdl.xml
```diff
@@ -0,0 +1,407 @@
...
+        photoelectron spectroscopy.
+    </doc>
+    <group type="NXentry">
+        <field name="definition">
+            <attribute name="version"/>
+            <enumeration>
+                <item value="NXmpes_arpes"/>
+            </enumeration>
+        </field>
+        <group name="arpes_geometry" type="NXcoordinate_system">
```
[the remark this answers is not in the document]
lukaspie: It is inherited as recommeded from `NXmpes`.

base_classes/NXelectronanalyzer.nxdl.xml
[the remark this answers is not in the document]
lukaspie: This is actually the voltage range of the power supply. I renamed it accordingly and updated the docs.

base_classes/NXfit.nxdl.xml
```diff
@@ -0,0 +1,199 @@
...
+            <dimensions rank="dimRank">
+                <doc>
+                    The ``input_independent`` field must have the same rank (``dimRank``)
+                    as the ``input_dependent`` field. Each individual dimension of ``input_independent``
+                    must have the same number of points as the corresponding dimension in
+                    the ``input_dependent`` field.
+                </doc>
+            </dimensions>
+        </field>
+        <field name="envelope" type="NX_NUMBER" units="NX_ANY">
```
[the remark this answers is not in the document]
lukaspie: This comes from my experience in XPS peak fitting, where the envelope represents the combined contribution of all peaks and background. So, basically, the sums all individual components (without the data points).

Would you call the envelope the sums all individual components _plus_ the data points? We can rename this field ro something more fitting, if you prefer.

base_classes/NXfit_background.nxdl.xml
```diff
@@ -0,0 +1,93 @@
...
+                </doc>
+            </dimensions>
+        </field>
+    </group>
+    <group name="function" type="NXfit_function">
+        <doc>
+            The functional form of the background.
+        </doc>
+    </group>
+    <attribute name="default">
```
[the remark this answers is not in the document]
lukaspie: removed

base_classes/NXelectronanalyzer.nxdl.xml
lukaspie: Yes, they are. Added a sentence that it should be in order of speed scan, but only if the axes describe the same physical quantity or different components of the same quantity.

base_classes/NXdeflector.nxdl.xml
lukaspie: Yes, it does use electrostatic fields (see the voltage field). We have not considered magnetic fields yet, I suggest this can be added if somebody actually needs it.

base_classes/NXfit.nxdl.xml
```diff
@@ -0,0 +1,199 @@
...
+                would be :math:`LLS(peak_1(p_1) + peak2(p_2) + backgr(p_3))`, where each :math:`p_x` describes the set
+                of parameters for one peak/background.
+                
+                It is however also possible to supply more involved formulas (e.g., in the case of constrained fits).
+            </doc>
+        </field>
+    </group>
+    <field name="figure_of_meritMETRIC" type="NX_NUMBER" nameType="partial" units="NX_UNITLESS">
+        <doc>
+            Figure-of-merit to determine the goodness of fit, i.e., how well the peak model
```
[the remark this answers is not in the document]
lukaspie: Correct, updated the docstring a bit to reflect this.

base_classes/NXelectronanalyzer.nxdl.xml
```diff
@@ -0,0 +1,279 @@
...
+        </doc>
+        <symbol name="nfa">
+            <doc>
+                Number of fast axes (axes acquired simultaneously, without scanning a
+                physical quantity)
+            </doc>
+        </symbol>
+        <symbol name="nsa">
+            <doc>
+                Number of slow axes (axes acquired scanning a physical quantity)
```
fsaizpoy, line 37: to use "while scanning" instead of "scanning"

base_classes/NXfit.nxdl.xml
```diff
@@ -0,0 +1,199 @@
...
+                The relative area can simply be derived by dividing the total_area by the
+                total area of all peaks or by a more complicated method (e.g., by
+                additionally dividing by the relative sensitivity factors). Details shall
+                be given in `global_fit_function`.
+            </doc>
+        </field>
+    </group>
+    <group name="backgroundBACKGROUND" type="NXfit_background" nameType="partial">
+        <doc>
+            One fitted background (functional form, position, and intensities) of the peak fit.
```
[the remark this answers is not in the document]
lukaspie: Done, added a link

base_classes/NXfit.nxdl.xml
```diff
@@ -0,0 +1,199 @@
...
+            </doc>
+        </field>
+        <field name="formula">
+            <doc>
+                For the optimization, the formula is any optimization process on the `global_fit_function` given above.
+                As an example, for a linear least squared algorithm on independent components, the formula of the error_function
+                would be :math:`LLS(peak_1(p_1) + peak2(p_2) + backgr(p_3))`, where each :math:`p_x` describes the set
+                of parameters for one peak/background.
+                
+                It is however also possible to supply more involved formulas (e.g., in the case of constrained fits).
```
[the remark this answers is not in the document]
lukaspie: Done, added the formula for the LLS in this case

base_classes/NXfit.nxdl.xml
```diff
@@ -0,0 +1,199 @@
...
+            <dimensions rank="dimRank">
+                <doc>
+                    The ``input_dependent`` field must have the same rank (``dimRank``)
+                    as the ``input_independent`` field. Each individual dimension of ``input_dependent``
+                    must have the same number of points as the corresponding dimension in
+                    the ``input_independent`` field.
+                </doc>
+            </dimensions>
+        </field>
+        <field name="input_independent" type="NX_NUMBER" units="NX_ANY">
```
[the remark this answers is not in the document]
lukaspie: See above

base_classes/NXmanipulator.nxdl.xml
```diff
@@ -0,0 +1,224 @@
...
+            Any additional actuator on the manipulator used to control an external
+            condition.
+        </doc>
+    </group>
+    <group type="NXsensor">
+        <doc>
+            Any additional sensors on the manipulator used to monitor an external condition.
+        </doc>
+    </group>
+    <group type="NXpositioner">
```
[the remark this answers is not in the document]
lukaspie: We didn't intend to describe for multi-stage manipulators so far, only multi-axis.

applications/NXmpes.nxdl.xml
lukaspie: In XPS, this is as far as I know always called iris.

applications/NXmpes.nxdl.xml
lukaspie: changed

applications/NXmpes.nxdl.xml
lukaspie: This is not the pixel size, but the pixel number.

applications/NXmpes_arpes.nxdl.xml
```diff
@@ -0,0 +1,407 @@
...
+            <field name="depends_on" type="NX_CHAR">
+                <doc>
+                    Link to transformations defining an APRES base coordinate system,
+                    which is defined such that the positive z-axis points towards the analyzer entry,
+                    and the x-axis lies within the beam/analyzer plane.
```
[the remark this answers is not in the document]
lukaspie: Yes

applications/NXmpes_arpes.nxdl.xml
lukaspie: @rettigl would you mind addressing this? I am not super familiar with the ARPES coordiante system.

applications/NXxps.nxdl.xml
lukaspie: We were proposing a formal mechanism some time ago (see e.g. https://github.com/nexusformat/definitions/pull/1271#discussion_r1415361365), but NIAC didn't want any math notation, so I don't see how such a formal mechanism can be implemented now.

applications/NXxps.nxdl.xml
lukaspie: (I think) it means essentially what I say [here](https://github.com/nexusformat/definitions/pull/1415#issuecomment-2657229558). If you think about a rotation around the z-axis from the perspective of viewing from the origin towards the positive end of the z-axis, in a left-handed CS the positive rotation direction is clockwise.

applications/NXxps.nxdl.xml
lukaspie: That was an old description, removed it.

applications/NXxps.nxdl.xml
lukaspie: I am afraid I don't understand this comment.

applications/NXxps.nxdl.xml
lukaspie: This doesn't make sense to me. Why is the use of depends_on different for beams compared to all other components in space?

applications/NXxps.nxdl.xml
lukaspie: That is what `@vector=[0, 0, 1]` means, right? Or do you imply to change this?

applications/NXxps.nxdl.xml
lukaspie: Do you mean to get rid of `y_flip` in that case? Not sure where the `-` sign is coming from if not.

applications/NXmpes_arpes.nxdl.xml
lukaspie: Just to be clear, the order should be
```
sample:NXsample
  depends_on="Reference to the end of the transformation chain,
              orienting the sample surface within the arpes_geometry coordinate system
              (``sample_azimuth`` or anything depending on it).
transformations:NXtransformations
  sample_azimuth
    @depends_on=offset_azimuth
  offset_azimuth
    @depends_on=sample_tilt
  sample_tilt
    @depends_on=offset_tilt
  offset_tilt
    @depends_on=sample_polar
  sample_polar
    @depends_on=offset_polar
  offset_polar
    @depends_on="Path to a transformation that places the sample surface into the origin of the arpes_geometry coordinate system."
```
Correct?

lukaspie: I implemented it like this now.

rettigl: yes, I think so, thanks.

base_classes/NXdeflector.nxdl.xml
PeterC-DLS: Add "Electrostatic deflectors..." to be specific

applications/NXmpes.nxdl.xml
PeterC-DLS: Then say it's the number not size and its units need fixing

applications/NXmpes.nxdl.xml
```diff
@@ -0,0 +1,1458 @@
...
+                                Calibrated binding energy axis.
+                                
+                                This concept is related to term `12.16`_ of the ISO 18115-1:2023 standard.
+                                
+                                .. _12.16: https://www.iso.org/obp/ui/en/#iso:std:iso:18115:-1:ed-3:v1:en:term:12.16
+                            </doc>
+                        </item>
+                    </enumeration>
+                </attribute>
+                <attribute name="reference" type="NX_CHAR" recommended="true">
```
PeterC-DLS, line 1322: Subject to #1410

applications/NXmpes_arpes.nxdl.xml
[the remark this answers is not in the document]
PeterC-DLS: But `NXxps` has a reduced listed in its "method" field

applications/NXxps.nxdl.xml
PeterC-DLS: A positive rotation about z-axis in a LH CS looks anti-clockwise when viewed from the origin along positive z - conversely, it looks clockwise looking to the origin from positive z. See https://en.wikipedia.org/wiki/Right-hand_rule

base_classes/NXcollectioncolumn.nxdl.xml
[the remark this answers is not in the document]
PeterC-DLS: Double quotes are used by all the current `nxdl` files in `base_classes` and `applications`

base_classes/NXfit.nxdl.xml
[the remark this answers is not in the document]
PeterC-DLS: Maybe `fit_total` or `fit_sum`?

base_classes/NXfit.nxdl.xml
```diff
@@ -0,0 +1,207 @@
...
+                - non-linear least squares
+                - Levenberg-Marquardt algorithm (damped least-squares)
+                - linear regression
+                - Bayesian linear regression
+            </doc>
+        </field>
+        <field name="formula_description">
+            <doc>
+                For the optimization, the formula is any optimization process on the ``global_fit_function`` given above.
+                As an example, for a linear least squared algorithm on independent components, the formula of the ``error_function``
```
PeterC-DLS, line 178:
```suggestion
                As an example, for a linear least squared algorithm on independent components, the formula of the ``error_function``
```
```suggestion
                As an example, for a least squares algorithm on independent components, the formula of the ``error_function``
                
```

base_classes/NXfit_function.nxdl.xml
```diff
@@ -0,0 +1,87 @@
...
+    <field name="formula_description">
+        <!--This should be a python parsable function. Here we should provide which keywords are available
+and a BNF of valid grammar.-->
+        <doc>
+            Description of the mathematical formula of the function, taking into account the
+            instances of ``TERM`` in ``fit_parameters``.
+        </doc>
+    </field>
+    <group name="fit_parameters" type="NXparameters">
+        <field name="TERM" nameType="any" units="NX_ANY">
```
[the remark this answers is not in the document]
PeterC-DLS: Yes, it would be more consistent than currently

base_classes/NXfit_background.nxdl.xml
```diff
@@ -0,0 +1,93 @@
...
+        </doc>
+        <symbol name="dimRank">
+            <doc>
+                Rank of the dependent and independent data arrays (for
+                multidimensional/multivariate fit.)
+            </doc>
+        </symbol>
+    </symbols>
+    <doc>
+        Description of the background for an NXfit model.
```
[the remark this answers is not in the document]
PeterC-DLS: This base class holds the fit results, values and function so maybe a name to reflect this.

base_classes/NXmanipulator.nxdl.xml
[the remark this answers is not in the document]
PeterC-DLS: Then update that group's `doc`

applications/NXxps.nxdl.xml
PeterC-DLS: Maybe this got displaced

base_classes/NXmanipulator.nxdl.xml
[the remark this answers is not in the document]
lukaspie: done

base_classes/NXfit_function.nxdl.xml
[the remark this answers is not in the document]
lukaspie: changed as discussed

base_classes/NXelectronanalyzer.nxdl.xml
lukaspie: I guess so.

base_classes/NXfit_background.nxdl.xml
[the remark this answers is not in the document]
lukaspie: It's that what a peak is in a fit model? I find the name describes it just fine.

base_classes/NXfit.nxdl.xml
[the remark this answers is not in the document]
lukaspie: changed to `fit_sum`, works for me

base_classes/NXcollectioncolumn.nxdl.xml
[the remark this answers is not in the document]
lukaspie: changed back to double quotes, also in all other new files

applications/NXxps.nxdl.xml
lukaspie: In general, I agree. Will make the necessary fixes once we have a stable version of #1415

applications/NXmpes_arpes.nxdl.xml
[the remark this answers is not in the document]
lukaspie: added such a reduced list here as well 👍

applications/NXmpes_arpes.nxdl.xml
```diff
@@ -0,0 +1,407 @@
...
+                <doc>
+                    Link to transformations defining an ARPES base coordinate system,
+                    which is defined such that the positive z-axis points towards the analyzer entry,
+                    and the x-axis lies within the beam/analyzer plane.
+                </doc>
+            </field>
+            <group type="NXtransformations">
+                <doc>
+                    Set of transformations, describing the orientation of the ARPES coordinate system
+                    with respect to the beam coordinate system (.).
```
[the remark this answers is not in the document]
lukaspie: beam coordinate system == default NeXus coordinate system in this case

applications/NXmpes.nxdl.xml
rettigl: yes, makes sense

applications/NXmpes_arpes.nxdl.xml
lukaspie: Will be added to the nxdl.xsd file as discussed in #1410, changed to target here

phyy-nx:
This will come up for vote on Thursday, April 17th!  Note its dependency on #1415.  It is assumed that if #1415 doesn't pass, then this will be held up, even it it does pass.

Also @PeterC-DLS, could you review your comments and see if they can be resolved?

applications/NXmpes.nxdl.xml
phyy-nx: @lukaspie @PeterC-DLS since #1410 is closed and we are waiting for a replacement, either `@target` fields need to be removed or the replacement PR will need to be reflected here.

lukaspie: Wait, I thought the idea was that you can then afterwards have a `@target` attribute attached to every field and group? That also means that I can put it in my application definition (because I want it to be recommended). Why would I need to remove it here?

See also https://github.com/nexusformat/definitions/pull/1410#issuecomment-2804333301, I don't think adding target to `linkType` and `groupType` allows me to add it to any field or group in any application definition because it would be an XML attribute, not a NeXus attribute of `attributeType`.

base_classes/NXdeflector.nxdl.xml
lukaspie: added a sentence to say it can be either electrostatic and electronmagnetic deflectors

applications/NXmpes.nxdl.xml
lukaspie: changed as suggested

applications/NXxps.nxdl.xml
lukaspie: It was agreed in #1415 that independent of the CS handedness (given by the determinant of its base vectors), the actual rotation is still right handed in that CS: https://github.com/FAIRmat-NFDI/nexus_definitions/blob/fairmat-2024-nxtransformations/base_classes/NXtransformations.nxdl.xml#L160. So basically, we _always_ go with the convention that when viewed from the origin along positive z, positive rotation is counterclockwise. 

For this CS, it means that the we can simply write "Rotation about the sample normal" and that should be clear enough.

applications/NXmpes.nxdl.xml
lukaspie: Another option: we keep `@reference` as originally suggested (to avoid conflicts with meaning of `@target`) to say what we originally wanted to convey: for a given `AXISNAME` like `angular1`, we want to record the corresponding transformations

We have 6 angles describing each electrons (3 angles for the sample and 3 for the detector) and then we write to angles `angular0` and `angular1`. Essentially, by this `@reference` attribute, we wanted to say that `angula0` corresponds to e.g. the tilt of the sample.  This can, but doesn't explicitly have to, be a link using `@target`, but basically just stating that this  `angula0` references the `sample_tilt_angle`.

applications/NXxps.nxdl.xml
lukaspie: I added a `beam_direction` with `@vector=[0,0,1]` now and changed the `depends_on` docs.

applications/NXmpes.nxdl.xml
lukaspie: I removed all such attributes for now so the vote can be started (see also discussion in #1410). We would like to emphaisze that this is a concept that we really need, hence we will look towards adding them later again if there is a better solution for `@reference` and `@target` attributes.

phyy-nx:
Hi all, as discussed in the Telco, we can now move this PR to an online vote. NIAC committee members please vote on this PR using emojis on this comment. 👍 for yes, 👎 for no, anything else (for example 👀) to abstain. We need 14 votes to hit quorum so please review and vote!

@PeterC-DLS, @lukaspie has responded to your comments, could you review and resolve them?  Thanks!

base_classes/NXfit_background.nxdl.xml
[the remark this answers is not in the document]
benajamin: In my experience with XPS, backgrounds can be a complicated mix of fitting and modelling and is sometimes done as a separate step from peak fitting. I think NXfit_background has value as a separate thing from NXfit.

phyy-nx:
Hi folks, as this vote ticks along, note that @lukaspie found a couple issues in this PR, which he has proposed addressing through https://github.com/FAIRmat-NFDI/nexus_definitions/pull/382, which would be merged into this one (#1424).  I don't believe anything there affects this vote, so I'm happy with merging it in, but I'll wait a day or so to see if there's any feedback.  Reasonable?  This seems a nice compromise between needing to start the vote over and with not moving too fast.  Thanks!

applications/NXxps.nxdl.xml
PeterC-DLS: I'm not sure you should be expressing this as a set of rotations as you are going from a right handed system to a left handed one.

applications/NXxps.nxdl.xml
PeterC-DLS: I think it's better to write the beam direction directly as direction cosines (if it's a static configuration) or as the result of (0,0,-1) being rotated by `beam_polar_angle` about `x` then rotated by `beam_azimuth_angle` about `z` - note both axes are in the left-handed `xps` coordinate system.

lukaspie: > as the result of (0,0,-1) being rotated by beam_polar_angle about x then rotated by beam_azimuth_angle about z

Isn't that what we the other transformations under `beam_probe` are doing. You rotate the beam in the CS that is defined by its basis vectors. Why do I need an additional rotation to describe the beam_direction.

applications/NXxps.nxdl.xml
lukaspie: I think you can still do that, given that the rotation convention is still the same (as discussed in #1415). You only change the handedness of the CS by using a basis vector set that has a determinant of -1.

base_classes/NXfit_background.nxdl.xml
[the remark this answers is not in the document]
PeterC-DLS: These two classes are recordings the results of some fitting of models. So the question is whether there's a gain in semantics by having separate classes or adding a label to a single fitted results class.

applications/NXxps.nxdl.xml
PeterC-DLS: I think this `NXtransformation` is unnecessary because it becomes circular when the beam direction is later defined in an `NXbeam`.

If you want to connect it to Mcstat, then `x=[0,0,1]`, `y=[-1,0,0]`, `z=[0,1,0]`

applications/NXxps.nxdl.xml
PeterC-DLS: A slight correction in my comment: it's `-beam_polar_angle` about `x`.  Are your transformations correct?

base_classes/NXfit_background.nxdl.xml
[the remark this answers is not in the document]
lukaspie: As I said, for me it is fine to keep both (as they are semantically different concepts) or to drop `NXfit_background` and just use a `background(NXpeak)` in `NXxps` instead. The advantage of the second approch is that we only need to make any changes in one class going forward.

Not sure if such a change is covered by the ongoing vote (which has enough votes to be accepted already.

lukaspie:
Remaining open conversations:
- Small changes in [FAIRmat-NFDI#382](https://github.com/FAIRmat-NFDI/nexus_definitions/pull/382) --> [link to comment](https://github.com/nexusformat/definitions/pull/1424#issuecomment-2825420133)
- Is `NXfit_background.nxdl.xml` needed? --> [link to comment](https://github.com/nexusformat/definitions/pull/1424/files#r1991558321)
- Docstring in coordinate system in `NXxps` --> [link to comment](https://github.com/nexusformat/definitions/pull/1424/files/2e61440681859ca1cf5a5ea7886cd0aa3fff5e84#r1998902053)
- `beam_direction` in `NXxps/NXentry/NXbeam` --> [link to comment](https://github.com/nexusformat/definitions/pull/1424/files/2e61440681859ca1cf5a5ea7886cd0aa3fff5e84#r1998922226)

phyy-nx:
Vote has passed!  Once the reviews have finished, comments have been resolved, an approval posted, this can be merged.

applications/NXxps.nxdl.xml
lukaspie: Wait, from [the figure](https://github.com/nexusformat/definitions/blob/2e61440681859ca1cf5a5ea7886cd0aa3fff5e84/applications/xps/xps_cs.png), when looking from the origin towards infinity, it is a clockwise rotation around z by `beam_azimuth_angle`, followed by a clockwise rotation around x (doesn't make a difference here, right) by `beam_polar_angle_of_incidence`.

Given that the default rotation convention has not changed (independent of left- or right-handed CS), positive rotation is equal to counterclockwise.

So the correct notation should be 
```
beam_probe:NXbeam
depends_on=transformations/beam_direction
transformations:NXtransformations
beam_direction
  @depends_on=beam_polar_angle_of_incidence
  @vector=[0, 0, 1]
beam_polar_angle_of_incidence
  @depends_on=beam_azimuth_angle
  @transformation_type=rotation
  @vector=[-1, 0, 0]
beam_azimuth_angle
  @depends_on=/entry/xps_coordinate_system
  @transformation_type=rotation
  @vector=[0, 0, -1]
```

Am I missing something?

PeterC-DLS: `beam_azimuth_angle` should have `@vector=[0,0,1]` and the initial direction is `[0,0,-1]`

lukaspie: I see what you mean. I changed it now to
```
beam_probe:NXbeam
depends_on=transformations/beam_direction
transformations:NXtransformations
beam_direction
  @depends_on=beam_polar_angle_of_incidence
  @vector=[0, 0, -1]
beam_polar_angle_of_incidence
  @depends_on=beam_azimuth_angle
  @transformation_type=rotation
  @vector=[-1, 0, 0]
beam_azimuth_angle
  @depends_on=/entry/xps_coordinate_system
  @transformation_type=rotation
  @vector=[0, 0, 1]
```

Is that correct?

lukaspie: > beam_azimuth_angle should have @vector=[0,0,1

Why is that? It is a clockwise rotation around z by `beam_azimuth_angle`, shouldn't it be `@vector=[0,0,-1]`, i.e., the same as for the clockwise rotation on around x by `beam_polar_angle_of_incidence`?

If you disagree, could you please write down exactly which transformations you expected, similar to my example above?

lukaspie: I am sorry, you are right. Looking down the positive z-axis, the beam azimuth is indeed a counter-clockwise rotation from the positive x-axis to the projection of the beam directiom, so the vector must be `@vector=[0,0,1]`.

lukaspie: The same must apply for the analyzer, of course, so I will change that as well.

PeterC-DLS: As I wrote earlier with an amendment: beam_dirn = [0,0,-1] depends on rotation of -polar about x and then depends on rotation of azi about z.

PeterC-DLS: Note you got to flip your beam direction

lukaspie: Yes, you are totally right, sorry for the confusion. The azimuth and polar angle for analyzer and sample are defined in the same way, of course. So in total, the transformations must be:

```
beam_probe:NXbeam
  depends_on=transformations/beam_direction
  transformations:NXtransformations
    beam_direction
      @depends_on=beam_polar_angle_of_incidence
      @vector=[0, 0, -1]
    beam_polar_angle_of_incidence
       @depends_on=beam_azimuth_angle
      @transformation_type=rotation
      @vector=[-1, 0, 0]
    beam_azimuth_angle
      @depends_on=/entry/xps_coordinate_system
      @transformation_type=rotation
      @vector=[0, 0, 1]

NXelectronanalyzer
  depends_on=transformations/analyzer_take_off_polar_angle
  transformations:NXtransformations
    analyzer_take_off_polar_angle
      @depends_on=analyzer_take_off_azimuth_angle
      @transformation_type=rotation
      @vector=[-1, 0, 0]
    analyzer_take_off_azimuth_angle
      @depends_on=/entry/xps_coordinate_system
      @transformation_type=rotation
      @vector=[0, 0, 1]

NXsample
depends_on=transformations/sample_rotation_angle
transformations:NXtransformations
  sample_rotation_angle
    @depends_on=sample_normal_polar_angle_of_tilt
    @transformation_type=rotation
    @vector=[-1, 0, 0]
  sample_normal_polar_angle_of_tilt
    @depends_on=sample_normal_tilt_azimuth_angle
    @transformation_type=rotation
    @vector=[-1, 0, 0]
  sample_normal_tilt_azimuth_angle
    @depends_on=/entry/xps_coordinate_system
    @transformation_type=rotation
    @vector=[0, 0, 1]
```
Do we agree on this?

> It is a positive rotation about z: the left hand grip/screw rule applies with left thumb upwards and fingers curling c/w when looking to origin.

I still don't understand this. Isn't the convention adopted trough #1415 now that right-hand rule _always_ applies for all rotations? That is, when viewed from the origin to inf along the axis, counterclockwise rotation is positive.

PeterC-DLS: Your comment (of February) with the matrices in #1415 is incorrect. The transformations are unchanged between left and right handed CSes.

lukaspie: Yes, the transformation matrices are in general independent from the handedness of the CS. You can define different **rotation conventions** (which are confusingly also sometimes called left- and right-handed), which then lead to different matrices. From what I understood from the literature, oftentimes people set these two conventions as the same, i.e., positive rotation is defined according to the coordinate system's handedness (i.e. left-handed CS -> left-handed rotations (clockwise rotation is positive when viewed along axis to infinity)). This is also what a lot of graphic APIs do (e.g., DirectX, OpenGL).

However, in the latest discussion surrounding #1415, it was agreed to always assume the convention above, i.e., when viewed from the origin to inf along the axis, counterclockwise rotation is positive. This is what people call right-handed rotations.  If you want to keep the same convention (right-hand rule for positive rotation) on a left-handed system, the rotation matrix formula doesn't change.

Hence my comment why the left hand rule would apply at all when following this convention.

PeterC-DLS: In the current state of #1415, there is no mention of rotation convention. What we call a positive rotation is determined by which hand rule we apply.

lukaspie: I think with #1563, the rotations here are corect using left-hand rule.

base_classes/NXfit_background.nxdl.xml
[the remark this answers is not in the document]
lukaspie: @PeterC-DLS what's your preference here? Shall we remove the extra class?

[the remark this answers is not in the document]
PeterC-DLS: My preference isn't too strong - if you want a separate class for background then just make it a subclass (or vice versa) to show they share the same structure.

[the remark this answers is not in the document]
lukaspie: The two classes were really a copy of each other, with the string "peak" replaced by "background" in various places. So when I try to subclass `NXfit_background` from `NXpeak`, the class is essentially empty. So, `NXfit_background` is essentially not needed at all. I have removed it now and replaced its use by `NXpeak`. If there is a need for a dedicated class for backgrounds in the future, we can always add it again.

lukaspie:
> @PeterC-DLS is there anything blocking us from merging this PR? All comments have been addressed and CI/CD is passing.

@PeterC-DLS thanks for the thumbs-up, can you then also please approve the PR? I think it needs at least one formal approval before I can merge it.

PeterC-DLS (approved):
I'm happy with this

lukaspie:
@PeterC-DLS I only realized after merging that this branch had 1281 commits on top of main, which are now all part of the repo history. While these are all valid commits (they are basically all from the work that we have been doing over at https://github.com/FAIRmat-NFDI/nexus_definitions), I am not so sure that this was the best way to go. Is there any standard practice within this repo what one should do when merging? Should commits from a PR be squashed?

If having these large number of PRs in the history is problematic, maybe it would be good to revert the merging of this PR and then bring the same contents again (this time using a squash commit).

We will have the same problems for #1422, #1423, and #1425 by the way.

PR #1426, opened 2024-09-24
Fairmat 2024: new common base classes

lukaspie:


prjemian:
Any new classes, whether intended to be a base class or an application definition, must go into contributed definitions until they are ratified by the NIAC.

prjemian (changes requested):
Any new classes must be add to `contributed_definitions/`, not directly to `base_classes/` or `application_definnitions/`.

PR #1428, opened 2024-09-24
Fairmat 2024: changes to dev_tools, groupings in manual

lukaspie:


dev_tools/utils/nxdl_utils.py
```diff
@@ -304,10 +308,13 @@ def belongs_to(nxdl_elem, child, name, class_type=None, hdf_name=None):
         return True
     if not hdf_name:  # search for name fits is only allowed for hdf_nodes
         return False
-    try:  # check if nameType allows different name
-        name_any = bool(child.attrib["nameType"] == "any")
-    except KeyError:
-        name_any = False
+    name_any = ("nameType" in child.attrib.keys()
+        and child.attrib["nameType"] == "any"
```
lukaspie: Can we do `(child.attrib["nameType"].get("nameType") == "any")`?

dev_tools/utils/nxdl_utils.py
```diff
@@ -304,10 +308,13 @@ def belongs_to(nxdl_elem, child, name, class_type=None, hdf_name=None):
...
+    ) or (
```
lukaspie: Can we do
```suggestion
    name_partial =  (
        child.attrib["nameType"].get("nameType") == "any"
    ) or (

File: dev_tools/utils/nxdl_utils.py [hunk 3]
```diff
@@ -316,12 +323,20 @@ def belongs_to_capital(params):
     """Checking continues for Upper case"""
     (act_htmlname, chk_name, name_any, nxdl_elem, child, name) = params
     # or starts with capital and no reserved words used
+    name_partial =  (
+        "nameType" in child.attrib.keys()
+        and child.attrib["nameType"] == "partial"
+    ) or (
```
lukaspie: Similarly,
```suggestion
    name_partial =  (
        child.attrib["nameType"].get("nameType") == "partial"
) or (
```

mkuehbach:
The more I think about the changes in this PR the less I find them useful:
- OK: Having a grouping of application definitions and contributed_definitions that are used likely within a certain field or special usage (for contributed) is useful. The main purpose is chunkify the long flat list of definitions.
- ??: Having a grouping of base classes I feel is i) ackward and ii) repetitive:
i) Base classes should be reused over new ones created so if it fits base_classes should be used in multiple appdefs and base classes, so how to really group them more than technical skeleton (NXobject, NXdata, NXentry), assets (NXuser, NXsample, NXinstrument), many base classes inherit from NXcomponent. ii) If I want to know what is part of an application definition I click on the application definition and there is the list of all NXclasses of all groups used. Only value I see is to convey additional conceptual or design informative messages to humans that have not found a place into the definition itself.
- ??: Having a collection of common base classes is tricky to maintain. Say if certain communities edit their grouping page and somebody from another community now uses one of the classes from their field, this class would be used in appdefs from two different domains, i.e. change state from original community used only to common. Means those who use it need to reevaluate if their usage triggers that the base class should be considered a common or a one used by appdefs of only one domain.

My suggestions, keep grouping pages in applications and contributed, remove groupings in base_classes. Do not repeat listing base classes, use groupings mainly only to convey additional information that is not part of the class definition itself, i.e. additional contextualization, how the design came about, addressing specific user groups if required.

phyy-nx:
From @woutdenolf, what about adding labeling to the .xml files, such that the groupings are automatic?  This would help with long term maintenance of these groupings.

mkuehbach:
This does not provide feedback to the PR https://github.com/FAIRmat-NFDI/nexus_definitions/pull/412
that was set up to isolate the complete restructuring of the grouping.
Adding a label aka tag to classes (base classes and noteworthy also appdefs) would enable automating the grouping that 412 I added by hand.

However, there are two significant traits of the FAIRmat proposal wrt to group classes instead of/or in addition to a long flat list: i) help people navigate, ii) provide additional contextualization that is not part of the NXDL.XML file thus the grouping rst files may not be completely autogenerateable unless they represent really only a moving of the class / header docstring into individual topic-specific rst files. Both feature branches @lukaspie for this @mkuehbach (412) PR suggests parts of the grouping written manually, bridging between pure nxdl.xml docstrings and documentation meant for humans.

So what to do with this PR?

phyy-nx:
Telco: accept as is, but interested in a 'grouping' tag or some such in the future.  @lukaspie will create an issue to work on in the next milestone.

We like the flat list moved into own list better.

phyy-nx (approved):

PR #1476, opened 2024-09-28
Remove nameType=any from reference_plane as this name is used literally

### LINKED ISSUE #1475: Devise way for validators to better understand substitutions in the definitions
A validator currently has problems understanding the relationships between the definition names and the actual object names used in a file. This is also an issue when dealing with string substitutions for object names, especially when there are multiple similar names. 

Edit: search from @prjemian:
https://github.com/search?q=repo%3Anexusformat%2Fdefinitions%20nameType&type=code

PeterC-DLS:
Fixes #1475 (partially)

### DIFF
--- a/base_classes/NXbeam.nxdl.xml
+++ b/base_classes/NXbeam.nxdl.xml
@@ -308,7 +308,7 @@
                 </doc>
             </attribute>
         </field>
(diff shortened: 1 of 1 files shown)


prjemian (approved):
I agree.  Thanks!

PR #1482, opened 2024-09-29
Greatly simplify the NX_CHAR documentation

### LINKED ISSUE #1441: scalar or list values for Fields/Attributes, but does it work for NX_CHAR, too?
Note that NX_CHAR seems to have a special restriction:  
https://manual.nexusformat.org/nxdl-types.html

phyy-nx:
Idea is to state that NX_CHAR is a string of characters, and remove the other documentation that doesn't seem as necessary nowadays.

Fixes #1441


prjemian (approved):
Great.  I, too, believe the additional text is no longer necessary.

PR #1484, opened 2024-09-29
fix CLI doc

woutdenolf:
This was copy&pasted from impatient_app but not modified. manual_app is for testing/building the NeXus documentation.


PeterC-DLS (approved):
LGTM

PR #1485, opened 2024-09-29
UPPERCASE to indicate flexible names

### LINKED ISSUE #1436: UPPER case notation

prjemian:
- close #1436

### editing checklist

- [x] add `partial` and its description
- [x] Define use of `nameType` in fields, groups, and attributes
- [ ] #1490
- [x] Revise a couple NXDL files for development here
- [x] Document all substitutable letters in italics.  This will make situations like "Q" and "I" in NXcanSAS more readable, as they will be in non-italics. (Change the Python code that renders the documentation.)
- [x] update the change history with this PR

### DIFF
--- a/applications/NXcanSAS.nxdl.xml
+++ b/applications/NXcanSAS.nxdl.xml
@@ -278,7 +278,7 @@
         </doc>
       </attribute>
 
-      <field name="Q" type="NX_NUMBER" units="NX_PER_LENGTH">
+      <field name="Q" type="NX_NUMBER" units="NX_PER_LENGTH" nameType="specified">
         <!-- http://www.cansas.org/formats/canSAS2012/1.0/doc/basics.html#definition-of -->
         <doc>
           .. index:: NXcanSAS (applications); Q
@@ -425,7 +425,7 @@
           </doc>
         </attribute>
       </field>
-      <field name="I" type="NX_NUMBER">
+      <field name="I" type="NX_NUMBER" nameType="specified">
         <!-- http://www.cansas.org/formats/canSAS2012/1.0/doc/basics.html#definition-of-intensity -->
 		<doc>
           .. index:: NXcanSAS (applications); I
@@ -1187,7 +1187,7 @@
         ============================
       -->
     
-    <group name="TRANSMISSION_SPECTRUM" type="NXdata" minOccurs="0">
+    <group name="TRANSMISSION_SPECTRUM" type="NXdata" minOccurs="0" nameType="any">
       <doc>
         The *SAStransmission_spectrum* element
         
--- a/nxdl.xsd
+++ b/nxdl.xsd
@@ -1,5 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
+<!-- Here's a procedure to validate this XSD Schema file:
+https://stackoverflow.com/a/48980995/1046449 -->
+
 <xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
 	targetNamespace="http://definition.nexusformat.org/nxdl/3.1"
 	xmlns:nx="http://definition.nexusformat.org/nxdl/3.1"
@@ -164,6 +167,32 @@
 			<xs:simpleType>
 				<xs:restriction base="xs:string">
 					<xs:pattern value=".*(\w+).*" />
+				</xs:restriction>
+			</xs:simpleType>
+		</xs:attribute>
+	</xs:attributeGroup>
+
+	<xs:attributeGroup name="nameTypeAttributeGroup">
+		<xs:attribute name="nameType" use="optional" default="specified">
+			<xs:annotation>
+				<xs:documentation>
+					This interprets the ``name`` attribute as:
+
+					* ``"specified"`` = Exactly this name.
+					* ``"any"`` = Any name not already used in group.
+					* ``"partial"`` = The capital letters are substitutable
+					  (empty string allowed) and the lower case letters
+					  (and other allowed symbols, such as ``_``) are not.
+
+					In each case, all names in a NeXus data file must
+					follow the naming :ref:`rules&lt;Design-Naming&gt;`.
+				</xs:documentation>
+			</xs:annotation>
+			<xs:simpleType>
+				<xs:restriction base="nx:NX_CHAR">
+					<xs:enumeration value="specified" />
+					<xs:enumeration value="any" />
+					<xs:enumeration value="partial" />
 				</xs:restriction>
 			</xs:simpleType>
 		</xs:attribute>
@@ -486,6 +515,7 @@
 			</xs:annotation>
 		</xs:attribute>
 		<xs:attributeGroup ref="nx:deprecatedAttributeGroup"/>
+		<xs:attributeGroup ref="nx:nameTypeAttributeGroup"/>
 	</xs:complexType>
 
 	<xs:group name="groupGroup">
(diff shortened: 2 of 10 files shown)


prjemian:
> Document all substitutable letters in italics

@phyy-nx Do you interpret this to mean _Change the Python code that renders the documentation_?

prjemian:
> check the sources for instances where the above rules are broken

I'll list any such instances here and open new issue for each NXDL file separately.

prjemian:
I'll modify the NXcanSAS and NXdata NXDL files as part of this PR to test documentation of substitutions for the variations of `nameType`: `"specified"`, `"any"`, and `"partial"`.

prjemian:
In application definitions, `nameType="specified"` is assumed throughout, unless provided.  In NXcanSAS, in various lines such as https://github.com/nexusformat/definitions/blob/ab5e65df8b787d472be233449d285702b145df7d/applications/NXcanSAS.nxdl.xml#L144
the attribute name has upper case but the "SAS" part cannot be replaced by the user.

prjemian:
The only case in NXcanSAS that requires a non-default `nameType` is here: https://github.com/nexusformat/definitions/blob/ab5e65df8b787d472be233449d285702b145df7d/applications/NXcanSAS.nxdl.xml#L1190

This situation is `nameType="any"`, as described in the documentation text.

prjemian:
@woutdenolf Can you help out with this PR?  

I'm stuck on the step _Document all substitutable letters in italics._  You understand the Python code now better.  Can you try this out on NXdata & NXcanSAS? Together, they represent all the `nameType=` options.

prjemian:
I added the `high priority` label since other FAIRmat PRs are waiting on this.

phyy-nx:
> > Document all substitutable letters in italics
> 
> 
> 
> @phyy-nx Do you interpret this to mean _Change the Python code that renders the documentation_?
> 
> 
> 


Yup, that's right

woutdenolf:
I added a commit for: Document all substitutable letters in italics.

I did not find a way to do **bolditalic**.

image

woutdenolf:
In Nxinstrument I see `<group type="NXtransformations" name="DIFFRACTOMETER" />`

Is this name variable or fixed?

The current formatting makes it fixed since **specified** is the default.

image

prjemian:
> I did not find a way to do bolditalic for substitutable characters

I agree.  A quick web search shows _**bolditalic**_ is not possible in rst.

prjemian:
Should be: `<group type="NXtransformations" name="DIFFRACTOMETER" nameType="any" />`

lsal93:
> I did not find a way to do _**bolditalic**_ for substitutable characters. Maybe someone can fix that. Currently it is just italic (and bold non-italic for fixed characters).

It should be possible to use html directly, as in this quick google find
https://kallimachos.github.io/rst/_sources/bolditalic.rst.txt

I tried the raw html workaround from above link using sphinx 7.3.7 and it seems to work.
`.. raw:: html`
`   <b><i>FIELDNAME</i>_scaling_factor</b>`

Though perhaps italics only is easier to tell from bold than bolditalics
<b><i>FIELDNAME</i>_scaling_factor</b>
vs
<i>FIELDNAME</i><b>_scaling_factor</b>

sanbrock:
Note that the new convention on substitutable names also effects how the method 'get_first_parent_ref' shall work. It is using [get_nx_namefit](https://github.com/nexusformat/definitions/blob/f301a2af86d181fe67765605bcc57332b4d102cf/dev_tools/utils/nxdl_utils.py#L107), but it only checks for UPPERCASE prefixes.
In one of our PRs, we have already implemented an improved version which checks for multiple UPPERCASE runs:
https://github.com/nexusformat/definitions/pull/1428
but this one is not yet checking 'nameType' according to the new rules.

woutdenolf:
`.. raw:: html` it needs to be an inline directive though.

prjemian:
Is there any way to format the _entire_ line in HTML?  Might not be if we also want to generate the proper anchors for cross-referencing.

woutdenolf:
Why would we need the RST formatting in `get_first_parent_ref` ?

For example

```
.. _/NXcanSAS/ENTRY/DATA-group:

    *DATA*: (required) :ref:`NXdata` :ref:`⤆ </NXentry/DATA-group>`
```

`/NXentry/DATA-group>` does not need formatting (I checked and the link still works as expected)

image

woutdenolf:
https://kallimachos.github.io/rst/bolditalic.html

```rst
:bolditalic:`This is bold and italic using the :bolditalic: role.`

.. raw:: html

   <b><i>This is bold and italics using raw html.</i></b>
```

Seems like a sphinx extension? https://github.com/kallimachos/rst/blob/9249dbd4916aece34d27d09fc2c287cadaa2b644/doc/conf.py#L20

woutdenolf:
Seems to work:

image

An example of the RST formatting of a partial attribute:

```
  **@**\ :bolditalic:`AXISNAME`\ **_indices**: (optional) :ref:`NX_INT <NX_INT>` 
```

sanbrock:
nameType [specified(default), any, or partial] and its meaning is now handled by the dev_tools/utils/nxdl_utils.py in the PR #1428 
Please note that the PR mentioned above is also implementing an example for providing detailed documentation for groups of AppDefs and BaseCalsses while also listing all definitions.

woutdenolf:
To be clear, the commits I added to this PR support nameType based RST formatting of the field/attribute/group/link name.

For example 

```
  .. _/NXdata@AXISNAME_indices-attribute:

  .. index:: AXISNAME_indices (file attribute)

  **@**\ :bolditalic:`AXISNAME`\ **_indices**: (optional) :ref:`NX_INT <NX_INT>`
```

Which results in

image

Anchor and index is not affected.

prjemian:
@nexusformat/developers - This is ready for review.  We should move promptly since other work (#1419, #1428, #1490, #1491) depends on this PR.

applications/NXcanSAS.nxdl.xml
```diff
@@ -278,7 +278,7 @@
         </doc>
       </attribute>
 
-      <field name="Q" type="NX_NUMBER" units="NX_PER_LENGTH">
+      <field name="Q" type="NX_NUMBER" units="NX_PER_LENGTH" nameType="specified">
```
sanbrock, line 281: Note that this is not necessarily needed, because 'specified' is the default option for nameType.  (Only iff a name is not specified for a group nameType becomes by default 'any')

applications/NXcanSAS.nxdl.xml
```diff
@@ -1187,7 +1187,7 @@
         ============================
       -->
     
-    <group name="TRANSMISSION_SPECTRUM" type="NXdata" minOccurs="0">
+    <group name="TRANSMISSION_SPECTRUM" type="NXdata" minOccurs="0" nameType="any">
```
sanbrock, line 1190: Yes, it is needed here.

nxdl.xsd
```diff
@@ -169,6 +172,32 @@
...
 	</xs:attributeGroup>
 
+	<xs:attributeGroup name="nameTypeAttributeGroup">
+		<xs:attribute name="nameType" use="optional" default="specified">
+			<xs:annotation>
+				<xs:documentation>
+					This interprets the ``name`` attribute as:
+
+					* ``"specified"`` = Exactly this name.
+					* ``"any"`` = Any name not already used in group.
```
sanbrock, line 182: Note that NeXus convention is not fully in line with the above statement default="specified".
To express that we may add here the following:  
Note that if no name and nameType are provided for a group, nameType="any" becomes the default.

sanbrock (changes requested):
Small suggestion for the xsd file.

applications/NXcanSAS.nxdl.xml
prjemian: Rather, since upper case has been the NeXus indicator that a name is flexible, let's leave this as a reminder, even though it is not needed, lest someone not associated with the canSAS think it to be flexible.

nxdl.xsd
prjemian: done per f9ceb0b

sanbrock (approved):

Issue #1451, opened 2024-09-26
NXenvironment


sanbrock:
- Can it be used as /ENTRY/INSTRUMENT/ENVIRONMENT or only /ENTRY/SAMPLE/ENVIRONMENT?
- can an ENVIRONMENT list all independent controllers (for a scan) or only a single one?
- can an ENVIRONMENT contain measurement sensors?
- Also support parameters which may not be measured, like temperature could be simply RT (Room Temperature) without it being actually measured by a sensor.

sanbrock:
PR #1414

sanbrock:
Proposal:
- use of @default
  A) add to each group manually as proposed
  B) no need to add, because it is a NeXus conventions that all groups can have this attribute
  C) add this to NXobject and let all group inherit it
- NXidentifier
  A) use as a Base class which we add wherever needed
  B) make the Fields to Attributes and use this base class as super class. This allows attaching a single identifier to any group  inheriting from this.
  C) same as B, but use nameType=partial, like identifier_serviceNAME, identifierNAME, and identifier_is_persistantNAME to support multiple IDs to add. E.g. identifier_experiment, identifier_collection, identifier_orcid, identifier_pdb
- NXfabrication
  features: vendor, model(@version), serial_number, construction_year, capability, IDENTIFIER
- NXsensor
  added 'current' to the enum list of measurement types  
  features FABRICATION
- NXactuator
  features: (short)name, type, FABRICATION, physical_quantity, pid, OUTPUT, TRANSFORMATIONS
- NXenvironment:
  added ACTUATOR
  added 'value' to support the case where no actual SENSOR, ACTUATOR is attached

prjemian:
Check the nxdl.xsd Schema for the default attribute. If it is there for all
fields and groups, it is not necessary to add the NXobject. Lots of work is
done in the nxdl.xsd Schema.

PR #1486, opened 2024-09-30
add identifierNAME to NXobject

lukaspie:
Implements vote in https://github.com/nexusformat/definitions/issues/1451#issuecomment-2382523273.
Depends on https://github.com/nexusformat/definitions/pull/1485 -> done

Any implementation should consider comments in https://github.com/nexusformat/definitions/pull/1416

### DIFF
--- a/base_classes/NXobject.nxdl.xml
+++ b/base_classes/NXobject.nxdl.xml
@@ -129,7 +129,7 @@
 					<doc>
 						Linking ISSN
 						
-						The linking The ISSN, or ISSN-L, is a specific ISSN that groups the different media of the same serial publication.
+						The linking ISSN, or ISSN-L, is a specific ISSN that groups the different media of the same serial publication.
 						
 						Syntax: NNNN-NNNC, where N a decimal digit character (i.e., in the set {0,1,2,...,9}), and C is in {0,1,2,...,9,X}
 						Example: 1234-5678 or 1234-567X
--- a/base_classes/NXobject.nxdl.xml
+++ b/base_classes/NXobject.nxdl.xml
@@ -193,19 +193,7 @@
 						Example: urn:isbn:0000000000000
 					</doc>
 				</item>
-				<item value="Other">
-					<doc>
-						If the identifier comes from a service that is not listed in the enumeration, ``type`` shall be set to "Other""
-						and specified in ``type_other``.
-					</doc>
-				</item>
 			</enumeration>
-		</attribute>
-		<attribute name="type_other">
-			<doc>
-				If the identifier comes from a different service, set type="Other" and specify
-				here.
-			</doc>
 		</attribute>
 	</field>
 </definition>
--- a/base_classes/NXobject.nxdl.xml
+++ b/base_classes/NXobject.nxdl.xml
@@ -48,7 +48,7 @@
 
 				It is recommended to use the most specific type when describing the identifier.
 				
-				For example, all IGSNs are DOIs and all DOIs are Handles; however, an IGSN should have type IGSN (and not DOI or Hdl).
+				For example, all IGSNs (see below) are DOIs and all DOIs are Handles; however, an IGSN should have type IGSN (and not DOI or Hdl).
 				Similarly, an ARK, Purl, ORCID and ROR identifiers should have their corresponding types and should not use the more generic URL identifier.
 			</doc>
 			<enumeration>
--- a/base_classes/NXobject.nxdl.xml
+++ b/base_classes/NXobject.nxdl.xml
@@ -189,11 +189,5 @@
 				here.
 			</doc>
 		</attribute>
-		<attribute name="is_persistent" type="NX_BOOLEAN">
-			<doc>
-				 True if the identifier is persistent (i.e., unique and available indefinitely),
-				 False otherwise.
-			</doc>
-		</attribute>
 	</field>
 </definition>
--- a/base_classes/NXobject.nxdl.xml
+++ b/base_classes/NXobject.nxdl.xml
@@ -31,23 +31,167 @@
 	</doc>
         <!--attribute name="name"><doc>name of instance</doc></attribute-->
 	<field name="identifierNAME" type="NX_CHAR" nameType="partial">
-        <doc>
-             The unique code, IRI or hash to resolve this reference.
-             Typically, this is stated by the service which is considered a complete
-             identifier, e.g., for a DOI it's something of the form `10.1107/S1600576714027575`
-             or `https://doi.org/10.1107/S1600576714027575`, which are both resolvable.
-        </doc>
-	    <attribute name="service" type="NX_CHAR">
+		<doc>
+			An identifier for a (persistent) resource.
+
+			An identifier, provided by some authority, that has been assigned to the
+			real-world object described by this ``NXobject``. To be useful, the identifier
+			must not be reassigned to a different real-world object. It is typical for
+			there to be some mechanism to resolve an identifier, obtaining metadata
+			about the object. Identifiers for which some guarantees exist regarding
+			this resolution process are called persistent identifiers.
+			Persistent identifiers are also known as PIDs.
+		</doc>
+		<attribute name="type" type="NX_CHAR">
+			<enumeration>
+				<item value="ARK">
+					<doc>
+						Archival Resource Key.
+						
+						An ARK is a Uniform Resource Identifier (URI) designed to support long-term access to a variety of information objects.
+						
+						Syntax: https://NMA/ark:/NAAN/Name[Qualifier]. Brackets indicate optional elements.
+						Example: https://example.org/ark:/12345/abcde
+					</doc>
+				</item>
+				<item value="DOI">
+					<doc>
+						Digital Object Identifier.
+						
+						A DOI is a unique alphanumeric string used to identify digital content. It consists of a prefix and a suffix, separated by a slash.
+						
+						Syntax: 10.XXXX/XXXXXX
+						Example: 10.1107/S1600576714027575
+					</doc>
+				</item>
+				<item value="Hdl">
+					<doc>
+						A handle is a unique identifier that consists of a prefix indicating the naming authority and a suffix representing the local name 
+						of a resource, cts. 
+						
+						A handle is a unique identifier used to facilitate the identification and management of digital objects. It is composed of a prefix that 
+						indicates the naming authority and a suffix that specifies the resource's local name. 
+						
+						This refers specifically to an ID in the Handle system operated by the Corporation for National Research Initiatives (CNRI).
+						
+						Syntax: hdl:prefix/identifier
+						Example: hdl:123456789/abc123
+					</doc>
+				</item>
+				<item value="IGSN">
+					<doc>
+						International Generic Sample Number.
+						
+						The IGSN is a unique identifier assigned to a specific sample or specimen in the context of scientific research.
+						
+						Syntax: https://igsn.org/{IGSN-ID}
+						Example: https://igsn.org/IGSN:XYZ123456
+					</doc>
+				</item>
+				<item value="ISNI">
+					<doc>
+						ISNI is an ISO standard to uniquely identify individuals and organizations involved in creative work, including pseudonyms
+						and other public personas.
+						
+						An ISNI is made up of 16 digits, the last character being a check character. The check character may be either a decimal digit
+						or the character “X”. 
+						
+						Syntax: https://isni.org/isni/{ISNI-ID}
+						Example: https://isni.org/isni/0000000121032683
+					</doc>
+				</item>
+				<item value="ISSN">
+					<doc>
+						International Standard Serial Number
+						
+						An ISSN is an 8-digit unique identifier used to distinguish a serial publication, whether in print or electronic form.
+						
+						Syntax: ISSN XXXX-XXXX
+						Example: ISSN 1234-5678
+					</doc>
+				</item>
+				<item value="ISSN-L">
+					<doc>
+						Linking ISSN
+						
+						The linking The ISSN, or ISSN-L, is a specific ISSN that groups the different media of the same serial publication.
+						
+						Syntax: ISSN-L XXXX-XXXX
+						Example: ISSN-L 1234-5678
+					</doc>
+				</item>
+				<item value="ORCID">
+					<doc>
+						Open Researcher and Contributor identifier.
+						
+						ORCID provides a free and persistent identifier that uniquely distinguishes authors and contributors in scientific research.
+						
+						Example: https://orcid.org/XXXX-XXXX-XXXX-XXXX
+						Syntax: https://orcid.org/0000-0002-1825-0097
+					</doc>
+				</item>
+				<item value="PURL">
+					<doc>
+						Persistent Uniform Resource Locator.
+						
+						A Persistent Uniform Resource Locator (PURL) is a type of URL designed to provide a stable, long-term reference to a web 
+						resource by using a resolver to redirect users to the resource's current location, even if it moves over time.
+						
+						A PURL has three parts: (1) a protocol, (2) a resolver address, and (3) a name.
+						
+						Syntax: https://purl.org/foo/bar
+						Example: https://purl.org/dc/elements/1.1/title
+					</doc>
+				</item>
+				<item value="ROR">
+					<doc>
+						Research Organization Registry
+						
+						A ROR ID is a globally unique identifier for research organizations, enabling unambiguous linking of institutions across systems.
+						
+						Syntax: https://ror.org/{ROR-ID}
+						Example: https://ror.org/052gg0110
+					</doc>
+				</item>
+				<item value="URL">
+					<doc>
+						Uniform Resource Locator
+						
+						Also known as web address, a URL is the address used to access a resource on the internet, 
+						specifying its location and the protocol to retrieve it.
+						
+						Syntax: scheme://domain:port/path?query_string#fragment_id
+						Example: https://www.example.com/about
+					</doc>
+				</item>
+				<item value="URN">
+					<doc>
+						Uniform Resource Name
+						
+						A URN is a unique, persistent identifier for a resource regardless of where it is stored.
+						
+						Syntax: urn:&lt;namespace&gt;:&lt;namespace-specific-string&gt;. The leading urn: sequence is case-insensitive.
+						Example: urn:isbn:0000000000000
+					</doc>
+				</item>
+				<item value="Other">
+					<doc>
+						If the identifier comes from a service that is not listed in the enumeration, ``type`` shall be set to "Other""
+						and specified in ``type_other``.
+					</doc>
+				</item>
+			</enumeration>
+		</attribute>
+		<attribute name="type_other">
 			<doc>
-				The service by which the resource can be resolved.
-				
-				Examples: doi, urn, hdl, purl, orcid, iso, url
+				If the identifier comes from a different service, set type="Other" and specify
+				here.
 			</doc>
-    	</attribute>
+		</attribute>
 		<attribute name="is_persistent" type="NX_BOOLEAN">
 			<doc>
-				True if the identifier is persistent (i.e., unique and available indefinitely),
-				False otherwise.
+				 True if the identifier is persistent (i.e., unique and available indefinitely),
+				 False otherwise.
 			</doc>
 		</attribute>
 	</field>
--- a/base_classes/NXobject.nxdl.xml
+++ b/base_classes/NXobject.nxdl.xml
@@ -30,7 +30,7 @@
 		This is the base object of NeXus
 	</doc>
         <!--attribute name="name"><doc>name of instance</doc></attribute-->
-	<field name="identifierNAME" type="NX_CHAR" name_Type="partial">
+	<field name="identifierNAME" type="NX_CHAR" nameType="partial">
         <doc>
              The unique code, IRI or hash to resolve this reference.
              Typically, this is stated by the service which is considered a complete
--- a/base_classes/NXobject.nxdl.xml
+++ b/base_classes/NXobject.nxdl.xml
@@ -141,8 +141,8 @@
 						
 						ORCID provides a free and persistent identifier that uniquely distinguishes authors and contributors in scientific research.
 						
-						Example: https://orcid.org/XXXX-XXXX-XXXX-XXXX
-						Syntax: https://orcid.org/0000-0002-1825-0097
+						Syntax: https://orcid.org/XXXX-XXXX-XXXX-XXXX
+						Example: https://orcid.org/0000-0002-1825-0097
 					</doc>
 				</item>
 				<item value="PURL">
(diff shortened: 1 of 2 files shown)


base_classes/NXobject.nxdl.xml
```diff
@@ -30,5 +30,25 @@
 		This is the base object of NeXus
 	</doc>
         <!--attribute name="name"><doc>name of instance</doc></attribute-->
+	<field name="identifierNAME" type="NX_CHAR" nameType="partial">
```
prjemian, line 33: This looks like it fits in the list of [reserved prefixes](https://manual.nexusformat.org/datarules.html#reserved-prefixes).  Add it to that list.

prjemian: The other prefixes are only defined in the documentation.  Should we make them available to XML by adding them here, similar to `"identifierNAME"`?  Or do this in a separate issue/branch/PR?

lukaspie: I added `identifier` to the reserved prefixes.

lukaspie: > The other prefixes are only defined in the documentation. Should we make them available to XML by adding them here, similar to `"identifierNAME"`? Or do this in a separate issue/branch/PR?

I think it is good to add them to `NXobject` directly. That still doesn't automatically make them reserved (it is still technically possible to overwrite them somewhere else), but at least the docs can be clearly defined and then inherited. 

I suggest to do it in a differen PR since for the identifier change we already have a NIAC vote (not sure if the other changes would need a separate one).

rayosborn:
I thought that, in the end, we decided against using "identiferNAME", instead of an identifier attribute, just called "identifier". Obviously, Sandor has the official record.

prjemian:
If it is an attribute, then it would be added to the `nxdl.xsd` schema.

rayosborn:
Isn't it possible also to add attributes to the NXobject class at least for groups?

prjemian:
It gets messy there. Can be very precise in the schema.

lukaspie:
I added some of the feedback provided by @paulmillar in https://github.com/nexusformat/definitions/pull/1416 with respect to the type of identifier. He made a strong point that `service` is not the right word, but rather something like `type`.

Note that here I am still adding `identifierNAME` as a _field_ rather than an attribute. My argument for this is that if we were to use an attribute and still keep the `service` and `is_persistent` (as was agreed if I recall correctly), we would need to add three attributes like this: `identifierNAME`, `identifier_serviceNAME`,  `identifier_is_persistentNAME`, which would be rather messy. Since `identifier` will be added to the reserved prefixes anyway, why not make it a field with the other two concepts as attributes? Maybe @sanbrock can also chime in with his opinion.

> I thought that, in the end, we decided against using "identiferNAME", instead of an identifier attribute, just called "identifier". Obviously, @sanbrock has the official record.

My understanding was that it shall be possible to have more than one identifier (e.g., from different services) for one object. I think this is also what @sanbrock noted. Even if identifier will become an attribute, I would suggest to use `<attribute name="identifierNAME" type="NX_CHAR" nameType="partial">`.

prjemian:
We have been reducing use of `type` to a few limited situations because its meaning varies with context.  XML Schema, XML, NXDL, numeric, data, units, ... (I'm certain I've missed a few.)  All of these want to use this noun.  Quickly, it became bewildering.  At the end of the day, the XML Schema and the NXDL data type win.  Here's an example: https://github.com/nexusformat/definitions/blob/67d8519b26601db921d1902957b5eb46e95afcbc/nxdl.xsd#L47

Please pick a noun which is not `type`.

rayosborn:
At the NIAC, I recall that people (including, I believe, @phyy-nx) thought the "is_persistent" tag unnecessary. We need @sanbrock to confirm, but I also believe that the attributes would just be "identifier" and "service." There is no need to add partial names here, which I agree become very messy. I have no idea whether we specifically voted on any of these issues.

sanbrock:
We have not voted on it but agreed that a workgrouo shall conclud with a reasonable solution in a form of a PR which can then be voted on.

sanbrock:
Indeed, we qgreef to use type instead of service.

sanbrock:
There was no real conclusion on is_persistant. Indeed, its necessity was debated, although I argued that it is acually good to know if an Idebtifier is a PID or not.
Ww may state that the type holds this information. E.g. doi or orcid are good examples of PIDs.

sanbrock:
Note that we did agree that our solution shall allow attaching multiple identifiers  (e.g. orcid, linkdin to a USER).
- identifierNAME with nameType=partial would do that job.
- identifier as a reserved prefix would also do that (but this solution is rather a convention /only appears in the documentation and/or in the xsd descriptuon of the NeXus Definiion Language/ rather than an asserted statement written in NXDL in a definition. While the former requires hard coded implementation in all NEXUS tools in any programming languages used, the later would be inferred automatically by any NEXUS tools which has implemented the interpretation of NeXus definitions written in NXDL. Because of this, I would prefer @identifierNAME with @identifier_typeNAME declared in NXobject over the actual simpler solution of listing identifier among the reserved prefixes.
- identifierNAME as a deckared Field in NXobject would also do the same, but it would allow adding @type to it. This seems to be the most elegant solution. Knowing that only a few udentifiers will be provided in practice for a given group, I do not think it has disadvantages compared to using attributes.

sanbrock:
Note that another alternative is 
@identifierTYPE - this allows the use of identifier_orcid, identifier_url, identifier_iri, etc. in the data file, and would make the @identifier_typeNAME (or identifier/@type) as an extra attribute unnecessary.

sanbrock:
NIAC suggestions  on the revwrite of NXidentifirler were recorded in #1451

sanbrock:
These  have been also voted for in Session J. See https://www.nexusformat.org/content/NIAC2024_minutes/

base_classes/NXobject.nxdl.xml
```diff
@@ -30,5 +30,164 @@
...
+						Example: hdl:123456789/abc123
+					</doc>
+				</item>
+				<item value="IGSN">
+					<doc>
+						International Generic Sample Number.
+						
+						The IGSN is a unique identifier assigned to a specific sample or specimen in the context of scientific research.
+						
+						Syntax: https://igsn.org/{IGSN-ID}
```
paulmillar, line 88: I'm not convinced this syntax is correct.

Per [the Wikipedia article](https://en.wikipedia.org/wiki/International_Generic_Sample_Number#History), IGSNs are now issued by DataCite.

Also, from this wikipedia article, an example of a DateCite issued IGSN is `10.58052/SSH000SUA`.  Note that this is also a DOI.

Per [the partnership agreement](https://datacite.org/blog/igsn-and-datacite-partnership/):

    Existing IGSN ID handles will now be registered IGSN ID
    DOIs and the handles aliased to the DOIs to ensure that
    these continue to resolve. 

I take this to mean that there are now DataCite-issued DOIs for all IGSNs, including those historical IGSNs issued before the partnership.

Therefore, I believe NeXus can use the same syntax for DOIs and IGSNs.

base_classes/NXobject.nxdl.xml
```diff
@@ -30,5 +30,164 @@
...
+				</item>
+				<item value="ISNI">
+					<doc>
+						ISNI is an ISO standard to uniquely identify individuals and organizations involved in creative work, including pseudonyms
+						and other public personas.
+						
+						An ISNI is made up of 16 digits, the last character being a check character. The check character may be either a decimal digit
+						or the character “X”. 
+						
+						Syntax: https://isni.org/isni/{ISNI-ID}
```
paulmillar, line 100: Do we really need the `https://isni.org/isni/` prefix?

Couldn't we store the ISNI-ID number and document that a URL may be derived from the ISNI by adding the prefix?

i.e., 
    Syntax: 16 base-10 digits stored without any spaces.

Apparently, specifying the lack of spaces [is important](https://en.wikipedia.org/wiki/International_Standard_Name_Identifier#ISNI_format).

base_classes/NXobject.nxdl.xml
```diff
@@ -30,5 +30,164 @@
...
+						Example: https://isni.org/isni/0000000121032683
+					</doc>
+				</item>
+				<item value="ISSN">
+					<doc>
+						International Standard Serial Number
+						
+						An ISSN is an 8-digit unique identifier used to distinguish a serial publication, whether in print or electronic form.
+						
+						Syntax: ISSN XXXX-XXXX
```
paulmillar, line 110: As above,  I don't think the prefix `ISSN ` in the syntax adds anything useful as the `type` attribute already identifies the value as an ISSN.

Also (borrowing from Wikipedia) the syntax might be better expressed as `NNNN-NNNC`

where N is in the set {0,1,2,...,9}, a decimal digit character, and C is in {0,1,2,...,9,X};

base_classes/NXobject.nxdl.xml
```diff
@@ -30,5 +30,164 @@
...
+			An identifier, provided by some authority, that has been assigned to the
+			real-world object described by this ``NXobject``. To be useful, the identifier
+			must not be reassigned to a different real-world object. It is typical for
+			there to be some mechanism to resolve an identifier, obtaining metadata
+			about the object. Identifiers for which some guarantees exist regarding
+			this resolution process are called persistent identifiers.
+			Persistent identifiers are also known as PIDs.
+		</doc>
+		<attribute name="type" type="NX_CHAR">
+			<doc>The type of identifier used.</doc>
```
paulmillar, line 46: I think it would make sense to include a comment about using the most specific type when describing the identifier.

For example, all IGSNs are DOIs and all DOIs are Handles; however, an IGSN should have type `IGSN` (and not `DOI` or `Hdl`).

Similarly, an ARK, Purl, ORCID and ROR identifiers should have their corresponding types and should not use the more generic URL identifier.

base_classes/NXobject.nxdl.xml
```diff
@@ -30,5 +30,164 @@
...
+						Also known as web address, a URL is the address used to access a resource on the internet, 
+						specifying its location and the protocol to retrieve it.
+						
+						Syntax: scheme://domain:port/path?query_string#fragment_id
+						Example: https://www.example.com/about
+					</doc>
+				</item>
+				<item value="URN">
+					<doc>
+						Uniform Resource Name
```
paulmillar, line 170: We probably should mention that identifiers with more specific `type` attribute values should not be stored as a URN, even when this is valid.

The two example that springs to mind are DOI and ISSN, but there may be others with valid URN namespaces.

The `doi` URN namespace [has been registered](https://www.iana.org/assignments/urn-formal/doi), so the URN `doi:10.1107/S1600576714027575` is a valid URN-based representation for the DOI `10.1107/S1600576714027575`.

Similarly, the `ISSN` URN namespace [has been registered](https://www.iana.org/assignments/urn-formal/issn), so the URN `URN:ISSN:1234-1231` is a valid URN that refers to the ISSN `1234-1231`.

However, I believe we shouldn't allow DOIs (and IGSNs) and ISSNs to be written using the `URN` type, but rather with their more specific types.

manual/source/datarules.rst
```diff
@@ -198,20 +198,22 @@ the prefixes reserved by NeXus.
...
-``DECTRIS_``  attributes, fields  reserved for use by Dectris                   https://www.dectris.com
-``IDF_``      attributes          reserved for use by pulsedTD Muon definition  https://www.isis.stfc.ac.uk/Pages/nexus-definition-v27924.pdf
-``NDAttr``    attributes          reserved for use by EPICS area detector       https://github.com/areaDetector
-``NX``        NXDL class          for the class names used with NeXus groups    https://www.nexusformat.org
-``NX_``       attributes          reserved for use by NeXus                     https://www.nexusformat.org
-``PDBX_``     attributes          reserved for the US protein data bank         https://www.rcsb.org
-``SAS_``      attributes          reserved for use by canSAS                    https://www.cansas.org
-``SILX_``     attributes          reserved for use by silx                      https://www.silx.org
-============  ==================  ============================================  =============================================================
+    reserved prefixes; identifier
```
paulmillar, line 201: It looks like there are some formatting changes.

IMHO, it would be better to separate out layout/formatting changes from content-modifying changes.  This patch should add the `identifier` row, but not modify the existing rows (if possible).

base_classes/NXobject.nxdl.xml
```diff
@@ -30,5 +30,164 @@
...
+					<doc>
+						A handle is a unique identifier that consists of a prefix indicating the naming authority and a suffix representing the local name 
+						of a resource, cts. 
+						
+						A handle is a unique identifier used to facilitate the identification and management of digital objects. It is composed of a prefix that 
+						indicates the naming authority and a suffix that specifies the resource's local name. 
+						
+						This refers specifically to an ID in the Handle system operated by the Corporation for National Research Initiatives (CNRI).
+						
+						Syntax: hdl:prefix/identifier
```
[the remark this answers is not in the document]
lukaspie: agreed, changed

base_classes/NXobject.nxdl.xml
lukaspie: I agree, this was the older syntax. I added some text describing this change:
```
Since 2021, IGSNs are issued by DataCite, meaning that hat there are now DataCite-issued DOIs for all IGSNs, including those historical IGSNs issued beforehand. Therefore, the syntax is the same as for DOIs.

Syntax: 10.XXXX/XXXXXX
Example: 10.1107/S1600576714027575
```

base_classes/NXobject.nxdl.xml
lukaspie: Sounds good. I used
```
A URL can be generated from the ISNI ID by combining it with the prefix https://isni.org/isni/, resulting in
https://isni.org/isni/{ISNI-ID}.
						
Syntax: 16 base-10 digits stored without any spaces.
Example: 0000000121032683
```

lukaspie:
CI/CD failing because multi-line doc handling is not implemented (see #1491)

phyy-nx:
Hey all, can anyone remember why we decided to use `identifierNAME` instead of a new group, NXidentifier, as was originally proposed?  Thanks!

base_classes/NXobject.nxdl.xml
phyy-nx: `hat` to `that`

phyy-nx:
I took the liberty of resolving a bunch of @paulmillar's comments, Paul feel free to re-review?

In the meantime, @lukaspie and @sanbrock, as discussed in a recent Telco (if my memory serves), the NIAC didn't really like how after this change, the web page for `NXobject` became dominated by identifier descriptions.  We considered two alternatives:

1. Revert to an NXidentifier object as originally proposed which can contain all this documentation
2. Move nearly all the identifier documentation to a separate webpage, linked to from `NXobject`

I think we'd be happy with either option, but regardless it would need a NIAC vote.  I note that this is a blocker for #1419.  Is it possible to get that in before the next NIAC telco?

phyy-nx:
Also a blocker for #1420

sanbrock:
While a separate page can be added on the use of Identifiers, during the discussion in the Telco it was also appreciated that the extensive documentation string is indeed collapsed in the HTML webpage.   
Hence, only because of the documentation, it is not needed to go back to our original proposal of creating NXidentifier as a new base class, and we can keep the conclusion of the NIAC meeting implementing it as a field.

base_classes/NXobject.nxdl.xml
```diff
@@ -30,5 +30,182 @@
...
+					</doc>
+				</item>
+				<item value="ORCID">
+					<doc>
+						Open Researcher and Contributor identifier.
+						
+						ORCID provides a free and persistent identifier that uniquely distinguishes authors and contributors in scientific research.
+						
+						Example: https://orcid.org/XXXX-XXXX-XXXX-XXXX
+						Syntax: https://orcid.org/0000-0002-1825-0097
```
paulmillar, line 145: "Syntax" and "Example" labels seem to be the wrong way around.  The "Example" label is the general format, while the "Syntax" label is before a specific example.

base_classes/NXobject.nxdl.xml
```diff
@@ -30,5 +30,182 @@
...
+				<item value="URN">
+					<doc>
+						Uniform Resource Name
+						
+						A URN is a unique, persistent identifier for a resource regardless of where it is stored.
+
+						It is recommended that identifiers with more specific type attribute (such as DOI or ISSN) values should not be stored as a URN,
+						even when this is valid. As an example, the URN doi:10.1107/S1600576714027575 is a valid URN-based representation for the DOI
+						10.1107/S1600576714027575, but it is strongly recommended to use type="DOI" in this case.
+
```
paulmillar, line 191: "strongly recommended" : is "strongly" needed?

prjemian: Not quite my choice of words but the goal when this was written was to emphasize this recommendation.  A wordsmith is most welcome to propose an improvement.

phyy-nx:
> While a separate page can be added on the use of Identifiers, during the discussion in the Telco it was also appreciated that the extensive documentation string is indeed collapsed in the HTML webpage. 

Totally forgot.  In that case I agree this can go to a vote in our next telco on Monday, once @paulmillar's new comments are addressed.  We'll follow the normal procedure of kicking off a two week voting period, unless we happen to get quorum!

base_classes/NXobject.nxdl.xml
lukaspie: You are right. I changed the labels.

base_classes/NXobject.nxdl.xml
lukaspie: I agree, "strongly" is probably a bit too much. I just changed it to "recommended".

base_classes/NXobject.nxdl.xml
lukaspie: This was already addressed in the NXDL file itself.

phyy-nx:
@lukaspie please remove type_other and Other from type (see [comment](https://github.com/nexusformat/definitions/pull/1407#issuecomment-2528429712) on #1407).   Once done, as discussed on the Telco, we will start an online vote for this PR.  Thanks!

phyy-nx:
Hi all, as discussed in the Telco, now that the above points have been addressed (thanks @lukaspie), we can now move this PR to an online vote.  NIAC committee members please vote on this PR using emojis. 👍 for yes, 👎 for no, anything else (for example 👀) to abstain.  We need 14 votes to hit quorum so please review and vote!

base_classes/NXobject.nxdl.xml
```diff
@@ -30,5 +30,170 @@
...
+						A DOI is a unique alphanumeric string used to identify digital content. It consists of a prefix and a suffix, separated by a slash.
+						
+						Syntax: 10.XXXX/XXXXXX
+						Example: 10.1107/S1600576714027575
+					</doc>
+				</item>
+				<item value="Hdl">
+					<doc>
+						A handle is a unique identifier that consists of a prefix indicating the naming authority and a suffix representing the local name 
+						of a resource, cts.
```
PeterC-DLS, line 78: What's `cts`? Perhaps, `etc`?

base_classes/NXobject.nxdl.xml
```diff
@@ -30,5 +30,170 @@
...
+			this resolution process are called persistent identifiers.
+			Persistent identifiers are also known as PIDs.
+		</doc>
+		<attribute name="type" type="NX_CHAR">
+			<doc>
+				The type of identifier used.
+
+				It is recommended to use the most specific type when describing the identifier.
+				
+				For example, all IGSNs are DOIs and all DOIs are Handles; however, an IGSN should have type IGSN (and not DOI or Hdl).
```
PeterC-DLS, line 51: What's an `IGSN`?

base_classes/NXobject.nxdl.xml
lukaspie: I think this was just a copying error. We can simply remove everything after the comma.

base_classes/NXobject.nxdl.xml
```diff
@@ -30,5 +30,170 @@
 		This is the base object of NeXus
 	</doc>
         <!--attribute name="name"><doc>name of instance</doc></attribute-->
-</definition>
+	<field name="identifierNAME" type="NX_CHAR" nameType="partial">
+		<doc>
+			An identifier for a (persistent) resource.
+
+			An identifier, provided by some authority, that has been assigned to the
+			real-world object described by this ``NXobject``. To be useful, the identifier
```
PeterC-DLS, line 38: How `real-world` should it be? Is a digital asset real?

base_classes/NXobject.nxdl.xml
lukaspie: These are the [International Generic Sample Numbers](https://ev.igsn.org/about-igsns). They are defined as another type for identifier in the enumeration.

base_classes/NXobject.nxdl.xml
lukaspie: I think a digitial asset is just as real as some kind of physical object, so they should be included here. Do you prefer a different description?

base_classes/NXobject.nxdl.xml
lukaspie: I removed it.

base_classes/NXobject.nxdl.xml
PeterC-DLS: Perhaps add "(see below)"

base_classes/NXobject.nxdl.xml
```diff
@@ -30,5 +30,170 @@
...
+						
+						Syntax: NNNN-NNNC, where N a decimal digit character (i.e., in the set {0,1,2,...,9}), and C is in {0,1,2,...,9,X}
+						Example: 1234-5678 or 1234-567X
+					</doc>
+				</item>
+				<item value="ISSN-L">
+					<doc>
+						Linking ISSN
+						
+						The linking The ISSN, or ISSN-L, is a specific ISSN that groups the different media of the same serial publication.
```
PeterC-DLS, line 132:
```suggestion
						The linking ISSN, or ISSN-L, is a specific ISSN that groups the different media of the same serial publication.
```

lukaspie: thanks for checking the spelling, implemented

base_classes/NXobject.nxdl.xml
lukaspie: ok, done

base_classes/NXobject.nxdl.xml
PeterC-DLS: Perhaps simply omit real-world:
```
An identifier, provided by some authority, that has been assigned to an object described by this ``NXobject``. To be useful, the identifier must not be reassigned to a different object
```

PeterC-DLS (approved):
LGTM

phyy-nx:
Vote has passed with 16 votes.  Thanks all!

@lukaspie feel free to merge

lukaspie:
> Vote has passed with 16 votes. Thanks all!
> 
> @lukaspie feel free to merge

Great, thanks!

PR #1497, opened 2024-10-15
show how to validate an XML file

### LINKED ISSUE #1380: validating XML files for area detector HDF5 writer
Add steps to the [existing example](https://manual.nexusformat.org/examples/epics/index.html#epics-area-detector-examples) about using the EPICS area detector HDF5 file writer.

To validate an XML file, you'll need the associated XML Schema (xsd) file, then use this command:
 
```bash
xmllint --noout --schema path/to/local/schema/file.xsd  file_to_validate.xml
```

- [attributes](https://github.com/areaDetector/ADCore/blob/master/XML_schema/NDAttributes.xsd) file
- [layout](https://github.com/areaDetector/ADCore/blob/master/XML_schema/hdf5_xml_layout_schema.xsd) file

_originally: https://github.com/BCDA-APS/bluesky_training/issues/296 by @prjemian_

prjemian:
- close #1380

### DIFF
--- a/manual/source/examples/epics/index.rst
+++ b/manual/source/examples/epics/index.rst
@@ -199,6 +199,25 @@ were arrays rather than single values.  That was due to an error in the original
 This has been fixed in the ``attributes.xml`` file presented here.
 
 
+.. _EPICS_Area_Detector.examples.validate-xml:
+
+Validate the XML files
+**********************
+
+To validate an XML file, you'll need the associated ``.xsd`` (XML Schema) file,
+then use this command::
+
+   xmllint --noout --schema path/to/local/schema/file.xsd  file_to_validate.xml
+
+These are the XML Schema files to use:
+
+===================  ===================================
+to validate ...      ... use this XML Schema file
+===================  ===================================
+``attributes.xml``   `NDAttributes.xsd <https://github.com/areaDetector/ADCore/blob/master/XML_schema/NDAttributes.xsd>`_
+``layout.xml``       `hdf5_xml_layout_schema.xsd <https://github.com/areaDetector/ADCore/blob/master/XML_schema/hdf5_xml_layout_schema.xsd>`_
+===================  ===================================
+
 .. _EPICS_Area_Detector.examples.python:
 
 Python code to store an image in a NeXus file
--- a/manual/source/examples/epics/index.rst
+++ b/manual/source/examples/epics/index.rst
@@ -216,6 +216,9 @@ to validate ...      ... use this XML Schema file
 ===================  ===================================
 ``attributes.xml``   `NDAttributes.xsd <https://github.com/areaDetector/ADCore/blob/master/XML_schema/NDAttributes.xsd>`_
 ``layout.xml``       `hdf5_xml_layout_schema.xsd <https://github.com/areaDetector/ADCore/blob/master/XML_schema/hdf5_xml_layout_schema.xsd>`_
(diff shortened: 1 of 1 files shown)


prjemian:
New documentation looks like this:
image

manual/source/examples/epics/index.rst
```diff
@@ -199,6 +199,25 @@ were arrays rather than single values.  That was due to an error in the original
...
+
+Validate the XML files
+**********************
+
+To validate an XML file, you'll need the associated ``.xsd`` (XML Schema) file,
+then use this command::
+
+   xmllint --noout --schema path/to/local/schema/file.xsd  file_to_validate.xml
+
+These are the XML Schema files to use:
```
PeterC-DLS, line 212: We should also mention `nxdl.xsd`

prjemian: Would you agree that the approach described in this [SO post](https://stackoverflow.com/a/48980995) is a good fit?

In this context, https://www.w3.org/2012/04/XMLSchema.xsd applies to both, with the same command instruction.

prjemian: Validating `nxdlTypes.xsd` this way shows 33 "Schemas validity error" reports. 

> Character content other than whitespace is not allowed because the content type is 'element-only'.

Fails for each unit example element, like this one: https://github.com/nexusformat/definitions/blob/c1225135b59f4b75002551ca5311e2ea3c4b5a6c/nxdlTypes.xsd#L93

This issue should add validation instructions.  Any change in `nxdlTypes.xsd` should come in a separate issue.  It will involve changes to the Python software that creates documentation from this file.

When these .xsd files were written years ago, I created and revised them with a licensed XML-aware editor.  Now, that license has lapsed and I have edited them in VSCode.  These examples of units have missed this type of validation.

prjemian: Also adding instruction for any NXDL file (uses `nxdl.xsd`), for completeness.

PeterC-DLS: > Would you agree that the approach described in this [SO post](https://stackoverflow.com/a/48980995) is a good fit?
> 
> In this context, https://www.w3.org/2012/04/XMLSchema.xsd applies to both, with the same command instruction.

Yes, we should include this in the CI too.

PeterC-DLS (commented):
LGTM

BTW layout.xml doesn't have schema metadata

prjemian:
Can you explain?

PeterC-DLS:
> Can you explain?

```
$ head -3 manual/source/examples/epics/layout.xml 
<?xml version="1.0" standalone="no" ?>
<hdf5_layout>
  <group name="entry">
```
cf
```
$ head -6 manual/source/examples/epics/attributes.xml
<?xml version="1.0" standalone="no" ?>
<!-- Attributes -->
<Attributes
  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:noNamespaceSchemaLocation="https://github.com/areaDetector/ADCore/blob/master/iocBoot/NDAttributes.xsd"
  >                                            
```
So there's a reference to the schema in the latter file

prjemian:
The metadata about [XML Schema location](https://www.w3schools.com/XML/schema_example.asp) is [not necessary](https://stackoverflow.com/a/15597057) to validate the XML layout files.  The schema location was not added in [ADCore](https://github.com/areaDetector/ADCore) since it was [mentioned](https://github.com/areaDetector/ADCore/issues/38#issuecomment-77470573) as an addition for layout files.

PeterC-DLS (approved):

Issue #1494, opened 2024-10-07
Inheritance of the NXroot class from NXobject


rayosborn:
At the recent NIAC, it was officially approved to allow base classes to inherit from the NXobject base class, in order to formalize sections of the documentation that allow certain classes, such as NXnote and NXcollection, to be added to any group. This is an important improvement in formalizing the status of these special groups, but it leads to a problem with the NXroot base class. At the moment, this only allows NXentry groups. With inheritance, suddenly a whole raft of additional groups, potentially including a number with `nameType=partial`, e.g., "FIELDNAME_errors", would be allowed. 

To prevent this, I would like to suggest that the "extends=NXobject" tag be removed from the NXroot "definition" tag, i.e., the NXroot class would not inherit anything. This is clearly a backward-compatible change because currently the NXobject class is empty, and I don't believe that it would have any additional side-effects, but I would welcome comments before issuing a PR.

Issue #1493, opened 2024-10-05
Specification of the NX_class attribute in the NXroot NXDL file


rayosborn:
NXroot is the only base class whose NXDL file explicitly added the `NX_class` attribute as a group attribute, with 'NXroot' as the sole enumerated value. The addition of the `NX_class` attribute to the HDF5 group attributes is obviously required by all base classes in order to define their class, but this is an implementation detail that is handled by the applications reading the file. The attribute is not even referenced in any of the other base classes or application definitions, apart from an anomalous comment in `NXsubentry`, which should also be changed for consistency.

I suspect that it was added because the `NXroot` class is not an HDf5 group, but the file-level object. Its presence here implies that the `NX_class` attribute is required as a file-level attribute by all NeXus files, but I don't believe this is (or should be) correct. In the nexusformat API, the root-level group is automatically assigned to the `NXroot` class if the attribute is not specified, but it is not added when writing a file, since its presence is completely redundant.

prjemian:
> it was added because the `NXroot` class is not an HDf5 group, but the file-level object.

See https://github.com/nexusformat/definitions/issues/149, which states

> To avoid invalidating existing NeXus data files, this attribute will be added as optional to NXroot.

prjemian:
If ever it was changed from optional, I believe that would have been a mistake.  The source does not say the attribute is required and this is indicated directly in the published documentation: https://github.com/nexusformat/definitions/blob/67d8519b26601db921d1902957b5eb46e95afcbc/base_classes/NXroot.nxdl.xml#L30-L39

![image](https://github.com/user-attachments/assets/60f068f8-c957-40bc-8fe9-8b65f2253ab7)

prjemian:
I'm removing the `bug` label.  The documentation is clear this is optional.

rayosborn:
There is nothing in the NXDL file to state that it is optional. That means that a strict validator will deem every NeXus file without it to violate the standard. At the very least, an optional tag needs to be added. I still think it's a mistake though.

prjemian:
Base class content is not optional?

rayosborn:
In principle, they are all optional, but when a new user reads the documentation, they are likely to assume it's required, ironically reinforcing the confusion that triggered this change in the first place. It was caused by people writing their own applications directly in h5py or the HDF5 API and tripping up because they didn't find the 'NX_class' attribute at the file level. In my view, the fix to #149 made this worse.

In #149, @stuartcampbell wrote: 
[...]

Issue #149, opened 2014-05-05
add attribute NX_class="NXroot" to NXroot NXDL


stuartcampbell:
**Original reporter**: _[prjemian](https://github.com/prjemian)_

From a discussion in the code repository, there is an inconsistency with the definition of `NXroot`.  It seems that NeXus data files need an attribute at the root level declaring the root node **is** `NXroot`.  

To avoid invalidating existing NeXus data files, this attribute will be added as optional to NXroot.

See:
- [http://trac.nexusformat.org/code/ticket/238]
- [http://lists.nexusformat.org/pipermail/nexus-developers/2010/000855.html]

PR #1505, opened 2024-11-04
Remove-NXroot-inheritance

rayosborn:
This PR addresses Issues #1493 and #1494. It contains two changes to the NXroot NXDL file,

* The first ensures that the NXroot base class does not inherit from the NXobject base class. When group classes that are valid in any base class are added to the NXobject base class, it is important to exclude NXroot to enforce the rule that only NXentry groups are permitted.
* The second removes the `NX_class` attribute definition from root attributes. This is the only example in the entire repository where the mechanism for defining classes is conflated with the contents of that class. Its presence here implies that the attribute is required, whereas it is completely redundant.


PeterC-DLS:
This was added to address #149

tacaswell:
I am 👍🏻 on removing the inheritance from `NXobject` (makes the definitions match the prose), but I am very 👎🏻 on removing `NX_class` attribute (and would support a proposal to make it required in the future).

rayosborn:
There is already an extended discussion of the `NX_class` issues in #1493, so please read that before making too many extra comments. In particular, it explains a little of the history behind its addition. I am strongly against making it required (which would require a NIAC vote), although I would be interested to hear from @tacaswell why he thinks it is necessary.

base_classes/NXroot.nxdl.xml
```diff
@@ -26,7 +26,15 @@
...
     type="group">
-    <doc>Definition of the root NeXus group.</doc>
+    <doc>
+        The root of a NeXus file.
+
+        In the NeXus standard, only NXentry groups are allowed at the root
+        level of a file, although it is permitted to include additional groups
+        that are not part of the NeXus tree and will not be validated by
+        NeXus tools. NeXus defines a number of file-level attributes that can
+        be used to annotate the file.
```
tacaswell, line 36:
```suggestion
        In the NeXus standard, only NXentry groups are allowed at the root
        level of a file, although it is permitted to include additional groups
        that are not part of the NeXus tree and will not be validated by
        NeXus tools. NeXus defines a number of root-level attributes that can
        be used to annotate the root group.
```

prjemian:
additional groups and datasets

tacaswell (approved):

tacaswell:
> Is that also true of the HDF5 C-API, i.e., that a path of '/' refers to a HDF5 group, or is that just in h5py?

My understanding is that this is also true in the c API

mkuehbach:
https://manual.nexusformat.org/nxdl_desc.html#definitiontype
@extends
The extends attribute allows this definition to subclass from another NXDL, otherwise extends="NXobject" should be used.

phyy-nx:
From Telco: bug-fix, doesn't need a vote.  After @mkuehbach's comment is resolved ok to merge!

rayosborn:
> @mkuehbach Does "allows" means that the attribute "@extends" is an optional one for an NX_class?

@mkuehbach, I have added a sentence to the `<doc>` tag explaining why the NXroot class does not inherit from the NXobject class. I think that should be sufficient, since the referenced web page in intended to give instructions for defining new classes, where inheritance is normally required. The NXroot class really is special, but adding a comment to the web page would, in my view, add an unnecessary complication to otherwise clear advice.

Issue #1506, opened 2024-11-04
Inheritance from the NXobject base class


rayosborn:
It was decided at the last NIAC meeting in September that the NXobject base class should be used to define objects that are valid in all other base classes. This needs to be implemented by adding the appropriate classes to the NXobject NXDL file.

According to the documentation, I believe that the following classes should be added:     `NXcollection`, `NXdata`, `NXgeometry`, `NXlog`, `NXnote`, `NXparameters`, `NXtransformations`.

I believe it also makes sense to add the partial name patterns that have reserved suffixes, so that these can also be validated in any base class: `GROUPNAME_log`, `FIELDNAME_set`, `FIELDNAME_errors`, `FIELDNAME_weights`, `FIELDNAME_mask`.

I will submit a PR to implement this policy.

phyy-nx:
For reference, specific NIAC vote was here:
https://github.com/nexusformat/definitions/issues/1442#issuecomment-2381282900

prjemian:
The `NXobject` class was introduced to be the parent of all other classes.  I agree that NXroot is exceptional since it never falls under NXentry.  I amend as follows:

The `NXobject` class is the parent class of all classes except `NXroot`.  

Further: It is intended that `NXobject` never be assigned to a `NX_class="NXobject"` attribute in a NeXus file.

prjemian:
As I recall, the `NXroot` class was introduced as a place where common file-level attributes could be provided with standardized names.  Oldest to newest, these are relevant issues and PRs:

- #149
- #189
- #466
- #783
- #784
- #789
- #836

PR #1507, opened 2024-11-04
Implement-NXobject-inheritance

rayosborn:
* Adds base classes that are allowed to be in any other base class.
* Adds groups and fields with partial name patterns that use the reserved suffixes.

This implements the proposal in #1506.


rayosborn:
If you viewed this earlier, you might have seen changes to the NXroot NXDL file. This is the subject of another PR and the changes have now been removed to keep them separate.

base_classes/NXobject.nxdl.xml
```diff
@@ -29,6 +29,36 @@
 	<doc>
 		This is the base object of NeXus
```
phyy-nx, line 30: Add some new documentation here at the top of the NXobject doc string:

- Note that nearly all base classes should derive from NXobject directly or indirectly through subclassing
- Note that all of the below groups and fields are explicitly allowed to be part of any NXobject derived object at any time.  In other words, if for example you have an NXdetector, it is always allowed to at NXnote to it.
- If FIELDNAME_XXX is added, it is expected that FIELDNAME also exists.  (Note to developers, validators should check for that)

The Telco agreed this doesn't need a vote.  So if you weren't on the Telco, now's the time to let us know if you think this needs a vote, otherwise, this be merged with an approving review.

phyy-nx (changes requested):

base_classes/NXobject.nxdl.xml
PeterC-DLS: Also, thinking about the application definition case a little more after the Telco: inheritance may mean that these new groups and fields become required unless you explicitly set them as optional.

phyy-nx: Adding `minOccurs=0` seems reasonable to me

phyy-nx (approved):

base_classes/NXobject.nxdl.xml
```diff
@@ -27,8 +27,43 @@
...
 	</doc>
-        <!--attribute name="name"><doc>name of instance</doc></attribute-->
+    <group type="NXcollection" minOccurs="0" />
+    <group type="NXdata" minOccurs="0" />
+    <group type="NXgeometry" minOccurs="0" />
+    <group type="NXlog" minOccurs="0" />
+    <group type="NXnote" minOccurs="0" />
+    <group type="NXparameters" minOccurs="0" />
+    <group type="NXtransformations" minOccurs="0" />
+    <group name="GROUPNAME_log" type="NXlog" nameType="partial">
```
PeterC-DLS, line 44: Should these be partial nameTypes optional too as their presence is not required if the instances are present.

base_classes/NXobject.nxdl.xml
```diff
@@ -27,8 +27,43 @@
...
+        respectively) that exists within the same group.
 	</doc>
-        <!--attribute name="name"><doc>name of instance</doc></attribute-->
+    <group type="NXcollection" minOccurs="0" />
+    <group type="NXdata" minOccurs="0" />
+    <group type="NXgeometry" minOccurs="0" />
+    <group type="NXlog" minOccurs="0" />
+    <group type="NXnote" minOccurs="0" />
+    <group type="NXparameters" minOccurs="0" />
+    <group type="NXtransformations" minOccurs="0" />
```
PeterC-DLS, line 43: Do we also need a `depends_on` attribute?

base_classes/NXobject.nxdl.xml
rayosborn: I think all partial names are automatically optional. The whole point is that they are an optional supplement any other field in the same group. If an application definition wants to require one, it needs to specify the name exactly.

base_classes/NXobject.nxdl.xml
rayosborn: I'm not sure why we would need to add it, since attributes are usually optional. Even the NXtransformations class appears in many application definitions without one.

base_classes/NXobject.nxdl.xml
sanbrock: Being in a base class actually also means that it is optional (by default). As Ray said, if needed, it can be made required in an AppDef. Note that optionality is not connected to nameType that only specifies the naming convention rule for inheritance. As an AppDef can require INSTRUMENT with nameType=any, it can also require GROUPNAME_log with nameType=partial. That would only mean that NXlog instances of xxx_log, yyy_log, etc. all belong to this concept.

phyy-nx: From Telco: add minoccurs=0 to these partial fields

(explicit is better than implicit)

base_classes/NXobject.nxdl.xml
phyy-nx: (from Telco) Nope, will be part of NXcomponent

phyy-nx:
Note for @rayosborn to resolve the conflicts :)

phyy-nx:

base_classes/NXobject.nxdl.xml
```diff
@@ -27,8 +27,41 @@
...
+    <group type="NXnote" minOccurs="0" />
+    <group type="NXparameters" minOccurs="0" />
+    <group name="GROUPNAME_log" type="NXlog" nameType="partial" minOccurs="0">
+        <doc>
+            NXlog group containing logged values of GROUPNAME.
+        </doc>
+    </group>
+    <field name="FIELDNAME_set" type="NX_NUMBER" nameType="partial" minOccurs="0">
+        <doc>
+            Target values of FIELDNAME.
```
paulmillar, line 49: Perhaps:

  Desired or target values of FIELDNAME.  In control theory, this is referred to as the setpoint.

base_classes/NXobject.nxdl.xml
```diff
@@ -27,8 +27,41 @@
...
+        </doc>
+    </group>
+    <field name="FIELDNAME_set" type="NX_NUMBER" nameType="partial" minOccurs="0">
+        <doc>
+            Target values of FIELDNAME. 
+        </doc>
+    </field>
+    <field name="FIELDNAME_errors" type="NX_NUMBER" nameType="partial" minOccurs="0">
+        <doc>
+            Uncertainties of FIELDNAME values.
```
paulmillar, line 54: Perhaps:

Uncertainties of FIELDNAME measurements.  Uncertainty values should be one sigma for normal errors.  Twice the uncertainty value provides the 68% confidence interval.

phyy-nx (approved):
The vote has passed. Thanks all. @lukaspie please merge at your convenience.

@paulmillar please be sure any issues you raised that you'd like to pursue are moved to a new issue. Note I resolved all of your comments but I'd like you to rescue your feedback from them! Thanks.

lukaspie:
> The vote has passed. Thanks all. @lukaspie please merge at your convenience after resolving conflicts.
> 

This is not my PR, so @rayosborn should do the merge :-)

rayosborn:
> > The vote has passed. Thanks all. @lukaspie please merge at your convenience after resolving conflicts.
> 
> This is not my PR, so @rayosborn should do the merge :-)

I would be happy to merge, but apparently, I don't have the privilege to do so. It's obviously a long time since I merged a PR in this repository. Do I have to be added to an approvers' list?

lukaspie:
> > > The vote has passed. Thanks all. @lukaspie please merge at your convenience after resolving conflicts.
> > 
> > 
> > This is not my PR, so @rayosborn should do the merge :-)
> 
> I would be happy to merge, but apparently, I don't have the privilege to do so. It's obviously a long time since I merged a PR in this repository. Do I have to be added to an approvers' list?

I think you have to rebase this on top of the main branch (or just merge in the main branch), so that the conflicts go away. Then you should be able to merge.

PR #1514, opened 2024-11-23
XDS: clarify the definition of the optional attribute in relation to minOccurs for fields and groups

woutdenolf:
optional is only a synonym for minOccurs=0 when optional=true.

When optional is false, this must mean the opposite so minOccurs>0 since nonNegativeUnbounded (not specifying the exact value of minOccurs, we only know it is greater than 0).

### DIFF
--- a/nxdl.xsd
+++ b/nxdl.xsd
@@ -502,7 +502,7 @@ https://stackoverflow.com/a/48980995/1046449 -->
 		<xs:attribute name="optional" use="optional" type="nx:NX_BOOLEAN" default="false" >
 			<xs:annotation>
 				<xs:documentation>
-					The value "true" is equivalent to minOccurs=0 while the value
+					A value of "true" is equivalent to minOccurs=0 while a value of
 					"false" is equivalent to minOccurs>0.
 				</xs:documentation>
 			</xs:annotation>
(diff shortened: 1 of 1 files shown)


nxdl.xsd
```diff
@@ -502,7 +502,8 @@ https://stackoverflow.com/a/48980995/1046449 -->
 		<xs:attribute name="optional" use="optional" type="nx:NX_BOOLEAN" default="false" >
 			<xs:annotation>
 				<xs:documentation>
-					A synonym for minOccurs=0.
+					The value "true" is equivalent to minOccurs=0 while the value
+					"false" is equivalent to minOccurs>0.
```
prjemian, line 506: Are the choices as large as `minOccurs>0`?  Or the only choice when `false` is `minOccurs=1`?

woutdenolf: > Are the choices as large as minOccurs>0?

Any nonNegativeUnbounded greater than zero.

nxdl.xsd
```diff
@@ -502,7 +502,8 @@ https://stackoverflow.com/a/48980995/1046449 -->
 		<xs:attribute name="optional" use="optional" type="nx:NX_BOOLEAN" default="false" >
 			<xs:annotation>
 				<xs:documentation>
-					A synonym for minOccurs=0.
+					The value "true" is equivalent to minOccurs=0 while the value
```
PeterC-DLS, line 505:
```suggestion
					A value of "true" is equivalent to minOccurs=0 while a "false" value
```

PeterC-DLS (approved):
LGTM

phyy-nx (approved):
LGTM

Issue #1513, opened 2024-11-23
XSD: questions about occurences


woutdenolf:
```xml
<xs:attribute name="minOccurs" use="optional" default="0" type="nx:nonNegativeUnbounded">
	<xs:annotation>
		<xs:documentation>
			Minimum number of times this ``group`` is allowed to be present within its
			parent group.  Note each ``group`` must have a ``name`` attribute
			that is unique among all ``group`` and ``field``
			declarations within a common parent group.
		</xs:documentation>
	</xs:annotation>
</xs:attribute>
<xs:attribute name="recommended" use="optional" type="nx:NX_BOOLEAN" default="false" >
	<xs:annotation>
		<xs:documentation>
			A synonym for optional, but with the recommendation that this
			``group`` be specified.
		</xs:documentation>
	</xs:annotation>
</xs:attribute>
<xs:attribute name="optional" use="optional" type="nx:NX_BOOLEAN" default="false" >
	<xs:annotation>
		<xs:documentation>
			A synonym for minOccurs=0.
		</xs:documentation>
	</xs:annotation>
</xs:attribute>
<xs:attribute name="maxOccurs" use="optional" type="nx:nonNegativeUnbounded">
	<xs:annotation>
		<xs:documentation>
			Maximum number of times this ``group`` is allowed to be present within its
			parent ``group``.  Note each ``group`` must have a ``name`` attribute
			that is unique among all ``group`` and ``field``
			declarations within a common parent ``group``.
		</xs:documentation>
	</xs:annotation>
</xs:attribute>
```

1. `maxOccurs` does not have a default. So what does it mean when not provided?
1. Does `optional="true"` imply `minOccurs=0`? Does it says something about `maxOccurs`?
1. Does `optional="false"` imply `minOccurs=1` or `minOccurs>=1`? Does it says something about `maxOccurs`?
1. Does `recommended="true"` imply `optional="true"`?
1. What does `recommended="false"` imply?

prjemian:
> 1. maxOccurs does not have a default. So what does it mean when not provided?

In this case, infinity (`type="nx:nonNegativeUnbounded"`).  https://github.com/nexusformat/definitions/blob/5b6daa03d50374ad25cd99df73bed5b4621371a9/nxdl.xsd#L509

prjemian:
> 3. Does optional="false" imply minOccurs=1 or minOccurs>=1? Does it says something about maxOccurs?

Was this answered for question 2?

[...]

PR #1515, opened 2024-11-23
maxOccurs is optional which means it has a default.

woutdenolf:
The default of maxOccurs for groups seems to "unbounded": https://github.com/nexusformat/definitions/issues/1513#issuecomment-2495512150. I'm not sure where this comes from though.

The default of maxOccurs for fields seems to be 1.

So the default is different from groups and fields which surprises me.

### DIFF
--- a/nxdl.xsd
+++ b/nxdl.xsd
@@ -506,7 +506,7 @@ https://stackoverflow.com/a/48980995/1046449 -->
 				</xs:documentation>
 			</xs:annotation>
 		</xs:attribute>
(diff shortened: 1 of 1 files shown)


prjemian:
> seems to "unbounded"

The XSD file uses `nx:nonNegativeUnbounded`.  This is unbounded.  No question to it.

woutdenolf:
> No question to it.

How is that? `nx:nonNegativeUnbounded` is just the type. How does the type determine what the default value is?

PR #1517, opened 2024-12-05
Remove limitation on how entries are named

### LINKED ISSUE #1438: does entry(NXentry) allow entry1, entry2, entry3 etc.
- Or should it be defined as entryID(NXentry) instead ?
- ENTRY(NXentry) is not strict enough but entryID (aka mixed capitalization) is also not because what does the character sequence ID tell?
- Only from the allowed characters as defined in NeXus, two of characters or multiple i.e.
also entrymyfamous(NXentry) is allowed?
- Why at all to start from 1 when 0 is an allowed character? Is the best practice counting defined in the documentation?

phyy-nx:
- Change all instances of <group type="NXentry" name="entry"> to <group type="NXentry"> (24 instances)
- Deprecate all instances of <attribute name="entry"> (6 instances)

Fixes #1438

This was voted on in [NIAC 2024](https://www.nexusformat.org/content/NIAC2024_minutes/#session-e-sept-28th-1330-utc) so this only needs a review from @nexusformat/developers

### DIFF
--- a/applications/NXarchive.nxdl.xml
+++ b/applications/NXarchive.nxdl.xml
@@ -37,7 +37,7 @@
 	        a mechanism to link all aspects of ISIS research from 
 	        proposal through to publication. 
     </doc>
-    <group type="NXentry" name="entry">
+    <group type="NXentry">
       <attribute name="index"></attribute>
       <field name="title" />
       <field name="experiment_identifier" type="NX_CHAR">
--- a/applications/NXarpes.nxdl.xml
+++ b/applications/NXarpes.nxdl.xml
@@ -33,12 +33,6 @@
     It has been drawn up with hemispherical electron analysers in mind. 
   </doc>
   <group type="NXentry">
-    <attribute name="entry" deprecated="See https://github.com/nexusformat/definitions/issues/1438">
-      <doc>
-        NeXus convention is to use "entry1", "entry2", ... 
-        for analysis software to locate each entry.
-      </doc>
-    </attribute>
     <field name="title" type="NX_CHAR"/>
     <field name="start_time" type="NX_DATE_TIME"/>
     <field name="definition">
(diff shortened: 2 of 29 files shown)


rayosborn:
I support the goals of this PR but I am puzzled by those application definitions, which have an attribute called "entry", e.g.,
```
<attribute name="entry" deprecated="See https://github.com/nexusformat/definitions/issues/1438">
```
Did the people writing the application definition think this was the way that the entry would be named, rather than by putting the name in the group tag? Using an attribute like that has no meaning in the NeXus standard. 

I'm not sure it should be deprecated; it should just be removed. Do we deprecate something that was never part of the standard in the first place?

phyy-nx:
I'm ambivalent.  In the meeting we said we'd deprecate it, which I think is the default course of action when removing something.  For reference, here are the application definitions with this attribute:

- NXxas.nxdl.xml
- NXarpes.nxdl.xml
- NXiqproc.nxdl.xml
- NXsqom.nxdl.xml
- NXxasproc.nxdl.xml
- NXsastof.nxdl.xml

I checked the first one and it's been there since 5f52e3a5479ca137c52ea1a5fcbf19cbe2fbfbf1 in 2010 when the definition was created.

rayosborn:
I suspect that the older examples reflected some confusion about how NXDL files should be written, not long after the standard was produced. I agree that we should be careful when making a change that could break things, but I'm not sure what would be broken here. If someone actually used the "entry" attribute in their file, it would serve no function.

prjemian:
I agree this was added likely by misunderstanding. My tendency is to yank it.  Is there any harm in deprecating at this time?  That's a more conservative approach.

phyy-nx:
Thinking more, I agree that removing it is fine.  It doesn't do anything.  I'll make that commit and we can think about it a bit more.

PeterC-DLS (approved):
I'm in favour of just removing the extraneous attribute

PR #1518, opened 2024-12-06
Permit simple inheritance for base classes

### LINKED ISSUE #1442: interface inheritance at base classes

phyy-nx:
- Adds section on "Inheritance in NeXus" to design.rst
- Changes "In contrast to NeXus base classes, NeXus supports inheritance in application definitions." to "Like base classes, NeXus supports inheritance in application definitions."
- Rewrites the FAQ question on inheritance

Notably, I left this paragraph alone in design.rst, as I didn't think that discussion needed to be cluttered by inheritance and that it didn't contradict the use of inheritance:

> NeXus base classes are not proper classes in the same sense as used in object oriented programming languages. In fact the use of the term classes is actually misleading but has established itself during the development of NeXus. NeXus base classes are rather dictionaries of field names and their meanings which are permitted in a particular NeXus group implementing the NeXus class. This sounds complicated but becomes easy if you consider that most NeXus groups describe instrument components. Then for example, a NXmonochromator base class describes all the possible field names which NeXus allows to be used to describe a monochromator.

Also noteably, I found that the definition of [`@extends`](https://www.nexusformat.org/content/NIAC2024_minutes/#session-g-sept-29th-1000-utc) didn't need changing:

> The `extends` attribute allows this definition to subclass from another NXDL, otherwise `extends="NXobject"` should be used.

My grepping and googling found these instances of specific discussion of inheritance but I might have missed some.  If anyone else spots more, let me know!

This is as voted on in [NIAC 2024](https://www.nexusformat.org/content/NIAC2024_minutes/#session-g-sept-29th-1000-utc) and so only needs a review by @nexusformat/developers.

Closes #1442

### DIFF
--- a/manual/source/design.rst
+++ b/manual/source/design.rst
@@ -524,13 +524,14 @@ annotate or in a ``NXcollection``. All of the base classes are documented in the
 Inheritance in NeXus
 --------------------
 
-Simple inheritance is supported in NeXus for both base classes and application defintions (which are
+Single inheritance is supported in NeXus for both base classes and application definitions (which are
 described below).  Extending a base class or application definition inherits the properties and objects of the
-parent class or definiton, such as groups, fields, attributes, and symbol tables. These properties and objects
-of the parent can then be overriden by the subclass.  Only single inheritance is allowed (you can only inherit
+parent class or definition, such as groups, fields, attributes, and symbol tables. These properties and objects
+of the parent can then be overridden by the subclass.  Only single inheritance is allowed (you can only inherit
 from a single parent).
 
-Use the ``@extends`` attribute of a definition to indicate which definition is being subclassed.
+Use the ``@extends`` attribute of a definition to indicate which definition is being subclassed. Base classes
+should extend from ``NXobject`` unless they are being subclassed.
 
 .. _NXdata-facilitates-TheDefaultPlot:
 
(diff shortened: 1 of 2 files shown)


phyy-nx:
Yay I fixed the build!  I believe this just needs an approval, but we can also review it on the next Telco

yayahjb (approved):
This should be useful and does no harm.  It should be merged.

manual/source/design.rst
```diff
@@ -519,6 +519,19 @@ But there are some base classes which have special uses which need to be mention
...
 
+.. _inheritance:
+
+Inheritance in NeXus
+--------------------
+
+Simple inheritance is supported in NeXus for both base classes and application defintions (which are
+described below).  Extending a base class or application definition inherits the properties and objects of the
+parent class or definiton, such as groups, fields, attributes, and symbol tables. These properties and objects
+of the parent can then be overriden by the subclass.  Only single inheritance is allowed (you can only inherit
```
PeterC-DLS, line 530:
```suggestion
of the parent can then be overridden by the subclass.  Only single inheritance is allowed (you can only inherit
```

manual/source/design.rst
```diff
@@ -519,6 +519,19 @@ But there are some base classes which have special uses which need to be mention
...
+from a single parent).
+
+Use the ``@extends`` attribute of a definition to indicate which definition is being subclassed.
```
PeterC-DLS, line 533: Could note that you should(?) use `"NXobject"` if it does not inherit

PR #1519, opened 2024-12-09
Move `NXlens_em` to base classes

lukaspie:
This implements what was discussed in https://github.com/nexusformat/definitions/pull/1407#issuecomment-2528429712. It moves `NXlens_em` from the contributed definitions to the base classes and adds it to `NXsource` as an renameable group. The difference between the `NXlens_em` from the contributed definitions and in base classes comes down to formatting, but no new terms have been added.

Note that `NXlens_em` itself is an extension of `NXcomponent` (which is meant as a basis for describing any components of an instrument).  `NXcomponent` is also making use of  `NXcircuit` (used for any sort of electronic circuit devices), so this is brought in here as well. These were originally to be discussed in #1424, but I think it's best to bring them in here.


phyy-nx:
Comment from Telco:
- NXcomponent seems like a big change.  And that many base classes should derive from it.  The NIAC requests a survey of base classes to identify those which should and those which should not derive from NXcomponent.  Examples include NXdetector which should, and NXdata which should not.
- As a corollary, adding NXcomponent as a new parent class to classes such as NXdetector, NXbeam, etc., would be a significant change, so move NXcomponent to a new PR, and make this one dependent on that new PR

Alternate proposal: instead of NXcomponent, add the necessary new fields to NXobject.  This would imply that NXobject has NXfabrication and NXcircuit, but that's ok.  Better to have more possibilities on NXobject than to add more base classes in general.

Reviewed in the NIAC: took a straw poll and it was almost evenly split between the above two options.  This will need more discussion.  @nexusformat/developers please chime in here!  Thanks!

lukaspie:
> Comment from Telco:
> 
> * NXcomponent seems like a big change.  And that many base classes should derive from it.  The NIAC requests a survey of base classes to identify those which should and those which should not derive from NXcomponent.  Examples include NXdetector which should, and NXdata which should not.
> * As a corollary, adding NXcomponent as a new parent class to classes such as NXdetector, NXbeam, etc., would be a significant change, so move NXcomponent to a new PR, and make this one dependent on that new PR
> 
> Alternate proposal: instead of NXcomponent, add the necessary new fields to NXobject. This would imply that NXobject has NXfabrication and NXcircuit, but that's ok. Better to have more possibilities on NXobject than to add more base classes in general.
> 
> Reviewed in the NIAC: took a straw poll and it was almost evenly split between the above two options. This will need more discussion. @nexusformat/developers please chime in here! Thanks!

#1525 implements option 1 here and adds a list of all classes for which it would likely make sense to extend `NXcomponent`. This is more meant as an exploration/survey rather than a votable proposal.

lukaspie:
@phyy-nx @sanbrock @mkuehbach given that discussions around NXcomponent have not been settled yet (there is a new for this anyway, #1525), I moved it out this PR and just copied all relevant concepts to `NXlens_em` for now, so we can at least move forward with that one.

I think this PR is then ready to be discussion in one of the next Telcos.

mkuehbach:
> Note that `NXlens_em` itself is an extension of `NXcomponent` (which is meant as a basis for describing any components of an instrument).  `NXcomponent` is also making use of  `NXcircuit` (used for any sort of electronic circuit devices), so this is brought in here as well. These were originally to be discussed in #1424, but I think it's best to bring them in here.

NXcomponent was removed NXlens_em extends NXobject

lukaspie:
> > Note that `NXlens_em` itself is an extension of `NXcomponent` (which is meant as a basis for describing any components of an instrument).  `NXcomponent` is also making use of  `NXcircuit` (used for any sort of electronic circuit devices), so this is brought in here as well. These were originally to be discussed in #1424, but I think it's best to bring them in here.
> 
> NXcomponent was removed NXlens_em extends NXobject

Yes exactly, this is now implemented.

base_classes/NXlens_em.nxdl.xml
```diff
@@ -0,0 +1,134 @@
...
+    <field name="description" type="NX_CHAR">
+        <doc>
+            Ideally, use instances of ``identifierNAME`` to point to a resource
+            that provides further details.
+
+            If such a resource does not exist or should not be used, use this free text,
+            although it is not recommended.
+        </doc>
+    </field>
+    <field name="value" type="NX_NUMBER" units="NX_ANY">
```
phyy-nx, line 55:
```suggestion
    <field name="power_setting" type="NX_NUMBER" units="NX_ANY">
```

base_classes/NXlens_em.nxdl.xml
```diff
@@ -0,0 +1,110 @@
...
+    <field name="description" type="NX_CHAR">
+        <doc>
+            Ideally, use instances of ``identifierNAME`` to point to a resource
+            that provides further details.
+
+            If such a resource does not exist or should not be used, use this free text,
+            although it is not recommended.
+        </doc>
+    </field>
+    <field name="power_setting" type="NX_NUMBER" units="NX_ANY">
```
paulmillar, line 55: I have no experience with EM, but an observation as an outsider ...

I'm wondering whether this value is continuous (accepting a very large number of different values) or whether this value accepts only a modest number of values.

If it's the latter, then it might be more flexible to consider the values as arbitrary labels that just coincidentally look like numbers (i.e., `type="NX_CHAR"`).  This would allow the field to capture non-numerical values, should EM manufacturers start to use them.

base_classes/NXlens_em.nxdl.xml
```diff
@@ -0,0 +1,110 @@
...
+             Descriptor for the operation mode of the lens when other details are not
+             directly controllable as the control software of the microscope
+             does not enable or is not configured to display these values.
+             
+             Like value, the mode can only be interpreted for a specific microscope
+             but can still be useful to guide users as to how to repeat the measurement.
+        </doc>
+    </field>
+    <!--control level perspective-->
+    <field name="voltage" type="NX_NUMBER" units="NX_VOLTAGE">
```
paulmillar, line 80: Would it make sense to consider measured vs set points?  E.g., field `voltage` records the measured voltage while `voltage_set` describes the set point.

paulmillar: Another point: the name is somewhat unfortunate -- just mirroring the units.

Would something like `name="control_voltage"` make sense?

base_classes/NXsource.nxdl.xml
```diff
@@ -235,6 +235,11 @@
 			 The size and position of an aperture inside the source.
 		</doc>
 	</group>
+	<group type="NXlens_em">
+		<doc>
+			Individual electromagnetic lenses inside the source.
```
paulmillar, line 240: Is the order of the `NXlens_em` significant?

base_classes/NXlens_em.nxdl.xml
paulmillar: OK, thanks for the explanation.  I think we can leave this until there's an actual need for it.

base_classes/NXlens_em.nxdl.xml
paulmillar: OK.  Thanks for the feedback.

base_classes/NXsource.nxdl.xml
paulmillar: OK, thanks for the feedback.

mkuehbach:
> Is there another link for details of electro-magnetic lenses in the literature see e.g. `L. Reimer https://doi.org/10.1007/978-3-540-38967-5 ? It forwards to a whole book with a price of 245,-€ . This might be frustrating, if you don't have the book already and if you don't know what scanning elctron microscopy is, you sure don't have the book.

Possibly one could also point to
* https://www.globalsino.com/EM/ or more specifically to an example like
* https://www.globalsino.com/EM/page1985.html

In terms of scientific depth, coverage, and succinctness though these references
are weaker than is Reimer. Cheaper than Reimer and also useful e.g.
P. Hawkes, "Magnetic Electron Lenses", ISBN/EAN: 9783642815188

phyy-nx (changes requested):
Vote has failed. We will discuss the review in tomorrow's telco.

sanbrock:
After fixing the reference, we can run the vote again.

lukaspie:
> After fixing the reference, we can run the vote again.

Added references to all the resources given above, so the NeXus user has some choices.

@sanbrock @phyy-nx please feel free to initiate another vote.

phyy-nx:
Hi all, as discussed in the Telco, we try again as an online vote. NIAC committee members please vote on this PR using emojis on this comment. 👍 for yes, 👎 for no, anything else (for example 👀) to abstain. We need 14 votes to hit quorum so please review and vote!

tacaswell:
Was there any discussion of applying this to the elements in a storage ring?

phyy-nx:
> Was there any discussion of applying this to the elements in a storage ring?

Interesting.  Do you mean like describing the upstream optical focusing elements at something like a synchrotron?

tacaswell:
> Do you mean like describing the upstream optical focusing elements at something like a synchrotron?

Yes.  However I don't know enough about how the accelerator folks describe the lattice to know if it sensible to try and interoperate with the electron microscope lens descriptions beyond the superficial (N-pole magnets focusing / steering electron beams).

phyy-nx:
> > Do you mean like describing the upstream optical focusing elements at something like a synchrotron?
> 
> Yes. However I don't know enough about how the accelerator folks describe the lattice to know if it sensible to try and interoperate with the electron microscope lens descriptions beyond the superficial (N-pole magnets focusing / steering electron beams).

Likewise, but I like the idea...

fsaizpoy:
Hi Lukas,
Is NXlens_em also intended to be used to describe PEEM microscopes?

lukaspie:
> Hi Lukas, Is NXlens_em also intended to be used to describe PEEM microscopes?

We have not considered it for PEEM yet. Mostly `NXlens_em` was designed for electron microscopes (SEM, TEM) and for the electromagnetic lenses in photoelectron spectrometers. We have a draft for an application definition for PEEM (https://github.com/FAIRmat-NFDI/nexus_definitions/pull/147) together with @PeterC-DLS, but there has been not much work in this direction for some time. I think most concepts in `NXlens_em` would be applicable for PEEMs as well though.

phyy-nx (approved):
The vote has passed, thanks everyone.  @lukaspie feel free to merge.

PR #1524, opened 2024-12-19
Correct reference to character limit

PeterC-DLS:


manual/source/datarules.rst
```diff
@@ -93,7 +93,7 @@ described by the following :index:`rules <rules; naming>`:
        ^[a-zA-Z0-9_]([a-zA-Z0-9_.]*[a-zA-Z0-9_])?$
 
    The length should be limited to no more than
-   63 characters (imposed by the HDF5 :index:`rules <rules; HDF5>` for names).
+   63 characters (imposed by the HDF4 :index:`rules <rules; HDF>` for names).
```
woutdenolf, line 96: I found [this](https://docs.unidata.ucar.edu/netcdf-c/4.9.2/netcdf_8h.html#af9d98edb32fe7feb7bf9b45480812b05)

> The max size of an SD dataset name in HDF4 (from HDF4 documentation) is 64.
> 
> But in in the wild we have encountered longer names. As long as the HDF4 name is not greater than NC_MAX_NAME, our code will be OK. 

I could not find it in the HDF4 specs: https://doi.org/10.5281/zenodo.13310722

woutdenolf (approved):

manual/source/datarules.rst
PeterC-DLS: Thanks, there's no such limit in HDF5.

PeterC-DLS: HDF4 library truncates to `VSNAMELENMAX` (a preprocessor symbol that is 64 by default) for in-memory names. See [hlimits.h#L55-L64](https://github.com/HDFGroup/hdf4/blob/ae94aece1220dfa2393be65d0f44fd3234307625/hdf/src/hlimits.h#L55-L64)

tacaswell (approved):
The guidance is only "should".  Even if we no longer maintain hdf4 bindings, it has historically been a supported target and it does little harm to acknowledge that.

manual/source/datarules.rst
tacaswell, line 96:
```suggestion
   63 characters (historically imposed by the HDF4 :index:`rules <rules; HDF>` for names).
```

PR #1408, opened 2024-09-23
Fairmat 2024: additions and clarifications in NXbeam

lukaspie:
With this PR, the idea is:
- to clarify how the `incident_energy` field in `NXbeam` is to be used. Specfically, for polychromatic beams, a single field `incident_energy` is not sufficient, but rather one needs information about the `incident_energy_spread`. In addition, for a polychromatic beam, it is useful to incidate the `incident_energy_weights`, i.e., how much each each of energies contributes to the overall intensity.
-  add some fields about pulsed beams and allow for descriptions of FROG (frequency-resolved optical gating) setups.
- add a field `previous_device` to indicate that this beam originate from a given beam device. This can be useful to describe the chain of devices and beams in the description of the beam paths (e.g., in an optical spectroscopy experiment).

PR #1525, opened 2024-12-19
`NXcomponent` as a parent base class

lukaspie:
This implements the discussion items in https://github.com/nexusformat/definitions/pull/1519#issuecomment-2551840081.

`NXcomponent` is meant as a parent base class that many base classes describing devices/instrument components should derive from. As requested in the comment above, this is meant as a survey of base classes to identify those which should and those which should not derive from `NXcomponent`.

These are the base classes that are now inheriting from NXcomponent:
in `base_classes`:
  - NXaperture
  - NXattenuator
  - NXbeam_stop
  - NXbending_magnet
  - NXcapillary
  - NXcollimator
  - NXcrystal
  - NXdetector
  - NXdisk_chopper
  - NXelectron_detector (through NXdetector)
  - NXfermi_chopper
  - NXfilter
  - NXflipper
  - NXfresnel_zone_plate
  - NXgrating
  - NXguide
  - NXinsertion_device
  - NXmirror
  - NXmoderator
  - NXmonitor
  - NXmonochromator
  - NXpinhole
  - NXpolarizer
  - NXpositioner
  - NXsample
  - NXsample_component
  - NXsensor
  - NXslit
  - NXvelocity_selector
  - NXxraylens

in `contributed_definitions`:
  - NXaperture_em
  - NXbeam_splitter
  - NXchamber
  - NXcollectioncolumn
  - NXcontainer
  - NXcorrector_cs
  - NXdeflector
  - NXebeam_column
  - NXelectronanalyser
  - NXenergydispersion
  - NXibeam_column
  - NXlens_em
  - NXlens_opt
  - NXmagnetic_kicker
  - NXmanipulator
  - NXpid
  - NXpolarizer_opt
  - NXpulser_apm
  - NXpump
  - NXquadrupole_magnet
  - NXreflectron
  - NXscanbox_em
  - NXseparator
  - NXsolenoid_magnet
  - NXspin_rotator
  - NXspindispersion
  - NXstage_lab
  - NXwaveplate

In addition, the `default` attribute is added to `NXobject` so that it doesn't have to be repeated in every class. As a consequence, it can be removed  This affects almost all of the classes above, but additionally the default attribute is removed from these classes (which are still inheriting from NXobject, not from NXcomponent):
- NXbeam
- NXcite
- NXdetector_group
- NXdetector_module
- NXevent_data
- NXgeometry
- NXinstrument
- NXlog
- NXnote
- NXoff_geometry
- NXorientation
- NXparameters
- NXprocess
- NXreflections
- NXshape
- NXtransformations
- NXtranslation
- NXuser

further changes:
- NXcylindrical_geometry: rename "beamline component" to "component" in docs
- NXoff_geometry: rename "beamline component" to "component" in docs

Open questions:
- Should any of these also inherit from NXcomponent?
  - NXbeam
  - NXdetector_module
  - NXinstrument


mkuehbach:
Wrt to the open questions of all the classes mentioned in that question only NXdetector_module should also inherit from NXcomponent

lukaspie:
Quick comment that I missed to mention in today's telco: are there any tools that would expect these classes to extend `NXobject` and not `NXcomponent`? For example, conditional checks like `if group.attribute(NX_class) == "NXobject"`.

We had a similar problem before when we tried to subclass `NXdata` (e.g. to `NXdata_mpes`) that some of the visualization tools (H5web in particular) are_only_ checking for `NXdata` and not for subclassed versions, so the default visualization breaks.

Since base class inheritance is now supported, such checks will **anyway** break on a smaller scale, e.g. for `NXelectron_detector` extending `NXdetector` (already part of the standard), so maybe it would be even good to do this once for many of the fundamental classes (like here) so that such errors are seen in most places and can be corrected immediately.

rayosborn:
The question of whether there are tools to implement inheritance is impossible to answer. I can modify the [`nexusformat package`](https://nexpy.github.io/nexpy) to explicitly include this inheritance, since all the base classes are really classes (i.e., not just objects with an `NX_class` attribute). However, inheritance is meaningless if someone is using, say, `h5py` or `H5web` directly, since HDF5 groups have no class. The [`NXvalidate` package](https://github.com/nexpy/nxvalidate) explicitly extend the `NXobject` class and will, I believe, automatically extend the `NXcomponent` class once it is approved, but again that is up to the package developer. 

That is one reason I have argued in the past for a package like `NeXpy` to be officially supported or at least recommended, i.e., so that there is a generic NeXus viewer that is guaranteed to follow changes in the standard, but I lost that battle years ago.

base_classes/NXobject.nxdl.xml
```diff
@@ -196,4 +196,17 @@
 			</enumeration>
 		</attribute>
 	</field>
+	<attribute name="default">
+        <doc>
+            .. index:: plotting
+
+            Declares which child group contains a path leading
+            to a :ref:`NXdata` group.
```
lukaspie, line 204: For supporting base class inheritance, should we say
```suggestion
            to a :ref:`NXdata` group or a group extending :ref:`NXdata`.
```

lukaspie: Changed to 
```
Declares which child group contains a path leading to a :ref:`NXdata` group or a group using a base class extending :ref:`NXdata`.
```

lukaspie:
As suggested in [#1407](https://github.com/nexusformat/definitions/pull/1408#issuecomment-2596241636), here is a flow diagram with the new inheritance (not containing changes to the contributed definitions).

NXcomponent_inheritance

lukaspie:
I also added **`inputs`** and  **`outputs`** to `NXcomponent`: lists of other components (NX_CHAR, rank of 0 or 1 (single or multiple)). This is to be used to chain components and/or beams together, as discussed originally in #1408.

lukaspie:
Removed `NXcircuit` from `NXcomponent` as discussed in today's Telco.

base_classes/NXcomponent.nxdl.xml
```diff
@@ -0,0 +1,92 @@
...
+            although it is not recommended.
+        </doc>
+    </field>
+    <field name="inputs" type="NX_CHAR">
+        <doc>
+            Instance or list of instances of ``NXcomponent`` (or base classes
+            extending ``NXcomponent``) or ``NXbeam`` that act as input(s) to this 
+            component.
+
+            Each input should point to the path of the group acting as input.
```
paulmillar, line 53: This description leaves me with a number of questions:

Where is the format for a path defined?  Can the path be absolute, can it be relative?

How are the paths within the list concatenated into a single string?  (space-separated list? comma-separated list? colon-separated list?) If paths have separator character, is white space allowed (or required) around separator character?

Is there an escaping mechanism?  One of the paths in the list might contain a path-separator character.  Without escaping, this could lead to ambiguity.

Is there any significance placed on the order of paths within this list?

The approach of serialising paths seems to be reimplementing HDF5's symlink feature.  Why not just use symlinks, instead?

base_classes/NXcomponent.nxdl.xml
```diff
@@ -0,0 +1,92 @@
...
+            the beam path in an experiment.
+        </doc>
+    </field>
+    <field name="outputs" type="NX_CHAR">
+        <doc>
+            Instance or list of instances of ``NXcomponent`` (or base classes
+            extending ``NXcomponent``) or ``NXbeam`` that act as output(s) of this
+            component.
+
+            For more information, see :ref:`inputs &lt;/NXcomponent/inputs-field&gt;`.
```
paulmillar, line 65: The description leaves me with the same questions as with the `inputs` field.

base_classes/NXcomponent.nxdl.xml
```diff
@@ -0,0 +1,92 @@
...
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+#
+# For further information, see http://www.nexusformat.org
+-->
+<definition xmlns="http://definition.nexusformat.org/nxdl/3.1" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" category="base" type="group" name="NXcomponent" extends="NXobject" xsi:schemaLocation="http://definition.nexusformat.org/nxdl/3.1 ../nxdl.xsd">
+    <doc>
+         Base class for components of an instrument - real ones or simulated ones.
```
paulmillar, line 26: I would suggest also mentioning that a NXcomponent has some well-defined position in space.

paulmillar:
Just to be clear: I really like the idea behind this PR and would like to see this go into NeXus.

However, I'd like the descriptions to be improved; in particular, the comments for the "inputs" field.

base_classes/NXcomponent.nxdl.xml
phyy-nx: The specification of paths using NX_CHAR is well precedented and I don't think it needs documentation here.  See for example `@depends_on` in NXtransfromations:

https://manual.nexusformat.org/classes/base_classes/NXtransformations.html#nxtransformations-axisname-depends-on-attribute

base_classes/NXcomponent.nxdl.xml
phyy-nx: Isn't this implied with all NeXus files?  That they are specifying things in space?

base_classes/NXcomponent.nxdl.xml
paulmillar: OK, fair point.  The use here matches the existing use in `@depends_on`.

To me, it looks like paths are currently suboptimally specified in NeXus.

However, give the existing precedent, this comment doesn't belong against this PR, but rather as a more general issue.  I'll open an issue and try to describe my concerns there.

Sorry for the noise.

phyy-nx: Agree they could use some clarification.  I searched for path in the docs and here are a couple that seem relevant:
https://manual.nexusformat.org/preface.html#class-path-specification
https://manual.nexusformat.org/design.html#index-35
Thanks!

base_classes/NXcomponent.nxdl.xml
paulmillar: I think there are several base classes that do not (in of themself) correspond to a specific physical object, located in 3D space; e.g.,
  * `NXprocess`
  * `NXnote`
  * `NXuser`
  * `NXdata`
  * `NXparameters`
  * `NXcite`
  * `NXroot`
  * `NXcylindrical_geometry`
  * ...

For example, within NeXus, I cannot provide locational information when including some NXdata group.  It just doesn't make sense.

For me, one of the interesting aspects of `NXcomponent` is that all subclasses inherit the `depends_on` field and `NXtransformations` group.  This gives us the ability to describe any subclass' physical location and orientation.  With this PR, any base class that is not a subclass of `NXcomponent` will lack this ability.

So, I believe a general statement about `NXcomponent` (inc. all subclasses) are that it describes things that exist in 3D space and have some sense of a location and orientation.  Further, I think it is helpful to include this observation in the description of the class, as it might help the reader develop a better intuition about this class.

Does that make sense?

base_classes/NXcomponent.nxdl.xml
lukaspie: With respect to the concatenation of paths within the list, the idea was to use the same as what is used for the [`auxiliary_signals` attribute](https://manual.nexusformat.org/classes/base_classes/NXdata.html#nxdata-auxiliary-signals-attribute) in `NXdata`. 

We could write 
```
String or array of strings holding the paths to instances of ``NXcomponent`` (or base classes extending ``NXcomponent``) or ``NXbeam`` that act as input(s) to this component.
```
An example would be:
```
entry:NXentry
    instrument:NXinstrument
      detector1:NXdetecor
      detector2:NXdetecor
    beam:NXbeam
    my_component:NXcomponent
	  @inputs = "entry/beam"
	  @outputs = ["entry/instrument/detector1", "entry/instrument/detector2"] 
```
Would that be better?

base_classes/NXcomponent.nxdl.xml
```diff
@@ -0,0 +1,92 @@
...
+    </doc>
+    <field name="applied" type="NX_BOOLEAN">
+        <doc>
+            Was the component used?
```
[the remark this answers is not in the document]
lukaspie: The idea was that you could have an instrument which contains a list of components. The `used` field tells you if the component was used in the experiment, e.g. was a lamp turned on, was a pump running, etc.

It's basically a short overview for the user to understand the parts of the instruments that were "active" during the measurements. Of course, actual information about the component at the time the experiment was performed can only be extracted from its other (meta)data.

base_classes/NXcomponent.nxdl.xml
paulmillar: I happy with pretty much whatever solution you provide.  My concern is that this solution (whatever it is) should be well-documented.

After reading the description, the person should be able to write down a valid entry.

Referring to other fields that have solved the same problem is OK, but it introduces the risk that (in some future version of NeXus) this other field value's syntax/format is modified.

But referring to `NXdata/auxiliary_signals` is fine.

In terms of practical steps, it wouldn't hurt to update the docstring and add a reference to an existing field with the same format/syntax (either as a quick fix that doesn't disrupt the voting, or as an easy post-fix change) .  We can also open an issue on the (apparent) lack of good definition of paths in NeXus and fix that more globally.

base_classes/NXcomponent.nxdl.xml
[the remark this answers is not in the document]
paulmillar: OK.  That's certainly logically consistent.  I would suggest updating the description to make this clearer.

On a higher (more philosophical) level, what is the benefit in including components that are switched off?

If they are not active and contributing to the experiment, couldn't they simply be omitted from the metadata?  A beamline could have lots of equipment lying around and that isn't being used in a particular experiment.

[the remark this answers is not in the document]
lukaspie: In general I agree that not every single component should be included if it doesn't contribute to the experiment. The idea is to indicate the presence of some components that are present and nominally turned off, yet may influence the measurement (think a component blocking parts of the beam, for example). This may help users to understand the data they obtain, also with respect to components that were not really used in the sense that they were actively part of the experimental procedure. Which components to include of course depends on the individual experiment, users will probably know which "unused" components can influence their data for a given setup.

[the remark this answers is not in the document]
paulmillar: Thanks for the explanation; that certainly sounds reasonable.

I think the PR would be beneficial if the doc included this information: that `applied` `false` implies the component is somehow influencing the measurement but is switched off.  To me, this isn't clear from "Was the component used?".

Since this field is optional, shouldn't the doc string also describe what are the semantics if `applied` field is no specified?  For example, does the absence of `applied` indicate the value is unknown, or that the component doesn't have an on/off state (so the field doesn't apply), or that some default value applies?

I imagine having a component that is influencing the data but is switched off is something of an edge case. In most cases, the components being described are switched on (insofar as they have an on/off state).

Perhaps the semantics when omitting the `applied` field could be equivalent to `applied=true`,  and that components that have no on/off state should omit this field.

In any case, I think it would be good if the doc string was updated to mention the intended semantics if the `applied` field is omitted.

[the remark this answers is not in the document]
lukaspie: I agree with all your points here, glad that we agree about the use of this field.

For me, applied would probably really only _need_ to be set if it was false, i.e. if the component is a bystander. In case it is actually used, you _can_ set `applied=true`, but you can also just just skip it and then `applied=true` is assumed. 

Essentially, this is just a case where the documentation needs to be more precise. But again, not sure how that works with the voting here. My suggestion would be to add all this documentation changes (also for the `input` field etc.) to this PR after voting has passed since it does not really change the meaning of the vote.

phyy-nx (approved):
The vote has passed.  Thanks all.  @lukaspie please merge at your convenience.

@paulmillar please be sure any issues you raised that you'd like to pursue are moved to a new issue.  Note I resolved all of your comments but I'd like you to rescue your feedback from them!  Thanks.

PR #1534, opened 2025-01-27
add base class NXcircuit

lukaspie:
`NXcircuit` is a base class for electronic circuit devices. We imagine it to be quite useful in classes where the actual implementation of the electronic circuit is of great importance for the measurements, e.g. for scanning probe microscopy or for certain electron detectors. 

Since  `NXcircuit`  is supposed to be used in many places, it was planned to be part of `NXcomponent`. It was requested that it be removed from #1525, so it is readded here.

Should be listed within `NXcomponent` once #1525 has been approved.


yayahjb:
I don't seem to find a circuit diagram in here.  Did I miss something.

lukaspie:
This PR will be covered via the EM proposal. Hence, I am closing here.

PR #1535, opened 2025-01-27
NXmx: make NXdetector_channel optional

### LINKED ISSUE #1511: NXmx: should NXdetector_channel in NXdetector be required?
This field probably doesn't make sense in many types of detector so would it be better to make it optional to avoid adding meaningless values or giving a validation error?

phyy-nx:
Fixes #1511

The vibe from the telco where this was discussed was that this doesn't need a vote since it's a bugfix.

### DIFF
--- a/applications/NXmx.nxdl.xml
+++ b/applications/NXmx.nxdl.xml
@@ -809,7 +809,7 @@
                         </doc>
                     </field>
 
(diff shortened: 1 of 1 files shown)


PeterC-DLS (approved):
LGTM

PR #1543, opened 2025-02-25
Multiple changes in NXoptical_specroscopy, NXellipsometry, NXraman

RonHildebrandt:
- Includes relaxation of URL requirement in NXraman
- Introducing the nameType=partial
- Changing environment in optical spectroscopy for the same do be directly dereived from NXsample and not from generic environment.
- Relaxing requirements of detector type


RonHildebrandt:
chosed wrong branch...

PR #1545, opened 2025-02-27
Bug fix: `NXcomponent` notation in `NXsample_component`

lukaspie:
This is a fix for a bug introduced in #1525.  `NXsample_component` extends `NXcomponent`, but the notation was wrong (`Nx` instead of `NX`). Should be an easy merge with no vote needed.

I also did a `git grep`, this is the only place where this shows up.


lukaspie:
@phyy-nx would you be able to review this bug fix? Should be an easy merge with no vote needed. Thanks!

PR #1546, opened 2025-02-28
`nameType="partial"` fixes for some base classes

lukaspie:
Fixes the `nameType`  for four different concepts:
- NXtransformations/AXISNAME_end -> was nameType="any", should be nameType="partial"
- NXtransformations/AXISNAME_increment -> was nameType="any", should be nameType="partial"
- NXhistory/identifierName -> has no nameType, should be nameType="partial"
- NXcalibration/identifierName -> has no nameType, should be nameType="partial"

For the fields in NXtransformations, this was probably a copy-paste error. For NXhistory and NXcalibration, which were recently added by me, it was simply forgotten to add the `nameType`.

### DIFF
--- a/base_classes/NXcalibration.nxdl.xml
+++ b/base_classes/NXcalibration.nxdl.xml
@@ -169,7 +169,7 @@
              :ref:`/NXdata/FIELDNAME_offset &lt;/NXdata/FIELDNAME_offset-field&gt;`.
         </doc>
     </field>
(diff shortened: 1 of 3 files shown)


lukaspie:
@phyy-nx would you be able to review this bug fix? Should be an easy merge with no vote needed. Thanks!

woutdenolf (approved):

PR #1550, opened 2025-03-12
nxdata stricter suggestion on when axisname indices can be omitted + clarify 3D or higher

woutdenolf:


base_classes/NXdata.nxdl.xml
```diff
@@ -75,11 +75,12 @@
...
-		* alternative axes: instead of a single ``x`` axis you can have several axes, one of which being the default.
-		* signals with more than one dimension: ``data`` could be 2D with axes ``x`` and ``y`` along each dimension.
-		* axes with more than one dimension: ``data`` could be 2D with axes ``x`` and ``y`` also being 2D, providing a
-		  unique ``(x, y)`` coordinate for each ``data`` point.
+		* Histogram data. For example ``x`` has one more value than ``data``.
+		* Alternative axes. For example instead of a single ``x`` axis you can have several axes, one of which being the default.
+		* Signals with more than one dimension. For example ``data`` could be 2D with axes ``x`` and ``y`` along each dimension.
+		  Equivalently 3- or higher-dimensional data is allowed.
+		* Axes with more than one dimension. For example ``data`` could be 2D with axes ``x`` and ``y`` also being 2D, providing a
+		  unique ``(x, y)`` coordinate for each ``data`` point. Equivalently 3- or higher-dimensional axes are allowed.
```
woutdenolf, line 83: Addresses @rayosborn comment https://github.com/nexusformat/definitions/pull/1396#pullrequestreview-2679270523

base_classes/NXdata.nxdl.xml
```diff
@@ -141,8 +142,8 @@
 		   field name in the :ref:`axes &lt;/NXdata@axes-attribute&gt;` attribute are used.
 
 		4. When :ref:`AXISNAME_indices &lt;/NXdata@AXISNAME_indices-attribute&gt;` is the same as the indices of "AXISNAME" in the
-		   :ref:`axes &lt;/NXdata@axes-attribute&gt;` attribute, there is no need to provide
-		   :ref:`AXISNAME_indices &lt;/NXdata@AXISNAME_indices-attribute&gt;`.
+		   :ref:`axes &lt;/NXdata@axes-attribute&gt;` attribute and :ref:`AXISNAME &lt;/NXdata/AXISNAME-field&gt;` is one-dimensional,
+		   the :ref:`AXISNAME_indices &lt;/NXdata@AXISNAME_indices-attribute&gt;` attribute can be omitted.
```
woutdenolf, line 146: Addresses @PeterC-DLS comment https://github.com/nexusformat/definitions/pull/1396#discussion_r1722968603

woutdenolf: For reference: for higher dimensional axes the AXISNAME_indices attribute can never be omitted unambiguously by using the indices in `@axes` instead. For example `@axes=["x", "x", "z"]` could be `@x_indices=[0,1]` or `@x_indices=[1,0]`.

base_classes/NXdata.nxdl.xml
woutdenolf: For reference: the examples could give the impression NXdata only support 1D and 2D data.

Solve by

* Clearly prepend with "For example"
* Add "Equivalently 3- or higher-dimensional data is allowed." and "Equivalently 3- or higher-dimensional axes are allowed"

base_classes/NXdata.nxdl.xml
woutdenolf: Sure it can be omitted for nD if it can be derived "unambigously". But then to define what "unambigously" means in this context would be rather convoluted. For example it won't work when `signal[23,23,17]`.

 So your original suggestion to omit AXISNAME_indices only for 1D still makes sense imo.

Issue #771, opened 2020-04-28
Make NeXus terminology machine resolvable


hgoerzig:
Make NeXus terminology machine resolvable so that NeXus terms can be mapped and integrated with different vocabularies.
Proposed starting point was https://manual.nexusformat.org/genindex.html

prjemian:
It might be easy enough to create a first cut, based on how language entries in the manual are created.  The language is defined in XML Schema and NeXus-specific NXDL (XML) files.  These files are parsed for content and documentation before rendering into reST.  The code that does the rendering could generate the ontology.

prjemian:
This would extract the NeXus [_lexicon_](https://en.wikipedia.org/wiki/Lexicon) from the NeXus XML files (both `.xsd` and `.nxdl`), which contain both _lexicon_ and _grammar_.

prjemian:
@hgoerzig : Could you send me your email (by email to jemian at my work place anl.gov)?  I've got the contact we discussed in the last telco.

prjemian:
Examples from NXentry (there are 39 such entries to the lexicon from NXentry) in json format:

```
[
  {
    "parent": "definition",
    "category": "attribute",
    "name": "default",
    "index_name": "default",
    "kind": "file",
    "optional": "(optional) ",
    "type": "NX_CHAR",
    "units": "",
    "base": "base_classes/NXentry.nxdl.xml",
    "sourceline": 33,
    "deprecated": false
  },
  {
    "parent": "definition",
    "category": "attribute",
    "name": "IDF_Version",
    "index_name": "IDF_Version",
    "kind": "file",
    "optional": "(optional) ",
    "type": "NX_CHAR",
    "units": "",
    "base": "base_classes/NXentry.nxdl.xml",
    "sourceline": 90,
    "deprecated": false
  },
  {
    "parent": "definition",
    "category": "field",
    "name": "title",
    "index_name": "title",
    "optional": "(optional) ",
    "type": "NX_CHAR",
    "units": "",
    "base": "base_classes/NXentry.nxdl.xml",
    "sourceline": 94,
    "deprecated": false
  },
  {
    "parent": "definition",
    "category": "field",
    "name": "experiment_identifier",
    "index_name": "experiment_identifier",
    "optional": "(optional) ",
    "type": "NX_CHAR",
    "units": "",
    "base": "base_classes/NXentry.nxdl.xml",
    "sourceline": 97,
    "deprecated": false
  },
  {
    "parent": "definition",
[...]

Issue #112, opened 2014-05-05
Merge CIF coordinates into NeXus


stuartcampbell:
**Original reporter**: *[prjemian](https://github.com/prjemian)*

Merge the method used by CIF to define coordinates into the NeXus coordinate system.

zjttoefs:
This is done, when #273 is done.

PR #1558, opened 2025-04-17
add `target` attribute to `NXobject` using new field super-type

lukaspie:
Follow up to #1410.

This is an initial draft for how attributes that are always allowed could be added. The idea is that a new generic field (with the new super-dtype `NX_DTYPE`) is added to `NXobject` that contains these default attributes.

For now, only the attribute `@target` was added (see comparable discussion in #1410).

In general, I believe the issue is a bit bigger. The `nxdl.xsd` file [defines](https://github.com/nexusformat/definitions/blob/main/nxdl.xsd#L692) multiple `<xs:attribute>` for the `fieldType` (like `units`, `long_name`, `signal`). However, some of these are used as XML attributes (i.e., writing something like 
```
<field name="my_field" type="NX_NUMBER" units=NX_ANY"/>
```
Then again, some other of these attributes are (to my understanding) to be used like this 
```
<field name="my_field">
    <attribute name="long_name"/> # same for axes, signal, primary, etc.
</field>
```

This may be due to me misunderstanding XML design, but I think with the definitions in the nxdl.xd file only the first use case (as XML attributes) is defined, whereas the second use case is not. To illustrate, this passes the schema validation:
```
<field name="my_field" long_name="a_longer_name" primary=0 signal=1</field>
```
I think this is not intended and rather these attributes should be defined somewhere else as attributes that each field can have. The mechanism given in this PR using a generic field in `NXobject` would be one way to define these.

### DIFF
--- a/base_classes/NXobject.nxdl.xml
+++ b/base_classes/NXobject.nxdl.xml
@@ -97,7 +97,7 @@
     </field>
     <attribute name="target" type="NX_CHAR">
 		<doc>
-			This attribute can be used to describe that this field is connected to another field.
+			This attribute can be used to describe that this group is connected to another group.
 			For links, this must be set to the path of the target of the link.
 
 			See :ref:`FIELDNAME/@target &lt;//NXobject/FIELDNAME@target-attribute&gt;` for more information
--- a/base_classes/NXobject.nxdl.xml
+++ b/base_classes/NXobject.nxdl.xml
@@ -64,6 +64,46 @@
             Boolean mask of FIELDNAME values. The value is masked if set to 1.
         </doc>
     </field>
+    <field name="FIELDNAME" type="NX_DTYPE" minOccurs="0">
+        <doc>
+            This is a generic field that defines the attributes that all fields can have.
+        </doc>
+        <attribute name="target" type="NX_CHAR">
+            <doc>
+                This attribute can be used to describe that this field is connected to another field.
+                For links, this must be set to the path of the target of the link.
+
+                Declares the absolute path of an existing field or group.
+
+                Matching regular expression::
+
+                    (/[a-zA-Z_][\w_]*(:[a-zA-Z_][\w_]*)?)+
+
+                For example, given a ``/entry/instrument/detector/polar_angle`` field which is to be
+                connected/linked to /entry/data/polar_angle`` in an ``NXdata`` group, this would be the
+                NeXus data file structure::
+
+                    /: NeXus data file
+                        /entry:NXentry
+                            /data:NXdata
+                                /polar_angle:NX_NUMBER
+                                    @target="/entry/instrument/detector/polar_angle"
+                            /instrument:NXinstrument
+                                /detector:NXdetector
+                                    /polar_angle:NX_NUMBER
+                                        @target="/entry/instrument/detector/polar_angle"
+            </doc>
+        </attribute>
+    </field>
+    <attribute name="target" type="NX_CHAR">
+		<doc>
+			This attribute can be used to describe that this field is connected to another field.
+			For links, this must be set to the path of the target of the link.
+
+			See :ref:`FIELDNAME/@target &lt;//NXobject/FIELDNAME@target-attribute&gt;` for more information
+			how to use this attribute.
+		</doc>
+	</attribute>
 	<field name="identifierNAME" type="NX_CHAR" nameType="partial">
 		<doc>
 			An identifier for a (persistent) resource.
--- a/nxdlTypes.xsd
+++ b/nxdlTypes.xsd
@@ -454,6 +454,7 @@
 			<xs:documentation>any valid NeXus field or attribute type</xs:documentation>
 		</xs:annotation>
 		<xs:union memberTypes="
+			nxdl:NX_DTYPE
 			nxdl:ISO8601
 			nxdl:NX_BINARY
 			nxdl:NX_BOOLEAN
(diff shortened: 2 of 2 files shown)


prjemian:
As described in a [previous comment](https://github.com/nexusformat/definitions/issues/771#issuecomment-620860468), the NeXus NXDL is a a set of XML specifications, each of which adhere to an XML Schema (`nxdl.xsd`).  For certain attributes such as `target`, it is the XML Schema that defines when and where the attribute appears.  In the case of `target`, the schema [defines](https://github.com/nexusformat/definitions/blob/2aede967caebc84f3335d50f1344f9b3ed2e8603/nxdl.xsd#L1082) it as a required attribute of a [link](https://github.com/nexusformat/definitions/blob/2aede967caebc84f3335d50f1344f9b3ed2e8603/nxdl.xsd#L1071).  Your proposal is to generalize the `target` attribute.

If NeXus is to continue use of `nxdl.xsd`, then refactor `target`'s definition into: https://github.com/nexusformat/definitions/blob/2aede967caebc84f3335d50f1344f9b3ed2e8603/nxdl.xsd#L607

prjemian:
FWIW: I've been adding the `target` attribute to [fields](https://github.com/BCDA-APS/apstools/blob/fd1b8c44db65af46505e3c6da04c118e4aa3088b/apstools/callbacks/nexus_writer.py#L504) and [groups](https://github.com/BCDA-APS/apstools/blob/fd1b8c44db65af46505e3c6da04c118e4aa3088b/apstools/callbacks/nexus_writer.py#L204) in NeXus/HDF5 files written from the `apstools` `NXWriter` code.  These additions are in anticipation of their use as a link.

While the NeXus standard defines `target` only for links, it does not preclude similar use throughout.  That's the basis for my use as shown.

lukaspie:
> If NeXus is to continue use of `nxdl.xsd`, then refactor `target`'s definition into:
> 
> https://github.com/nexusformat/definitions/blob/2aede967caebc84f3335d50f1344f9b3ed2e8603/nxdl.xsd#L607

Am I missing something on how XSD works?

To my understanding, with `<xs:attribute`, you define the attributes that you can use _within_ the XML element of a given `complexType`. So, if I were to bring `<xs:attribute name="target" use="required" type="nx:validTargetName">` to `<xs:complexType name="basicComponent">` (as you suggest), I could do
```
<field name="my_field" target="some_other_field/>
```
in an nxdl.xml file afterwards (this is what's described e.g. here: https://www.w3schools.com/xml/schema_simple_attributes.asp).

However, with that you **don't** define
```
<field name="my_field">
    <attribute name="target"/> # this is of complexType `attributeType
</field>
```
everywhere. You still need to define that attribute within some NXDL file to state that any field can have a `@target` attribute.

Am I missing something about how XSD works in this case? Or is the second case just convention (from the documentation) and not explicitly defined through the XSD file. If so, why not add it explicitly to `NXobject`?

lukaspie:
This is generally confusing to me, see how the `xs:attributes` of `fieldType` are used within the NXDL files, e.g. [`long_name`](https://github.com/search?q=repo%3Anexusformat%2Fdefinitions+long_name&type=code) vs. [`type`](https://github.com/search?q=repo%3Anexusformat%2Fdefinitions+type&type=code&p=4). One is used as a NeXus attribute (i.e., of type `attributeType`), whereas the other is used as an XML attribute of `fieldType`.

lukaspie:
> Recently, some have found it interesting to [add](https://github.com/nexusformat/definitions/blame/main/base_classes/NXobject.nxdl.xml?rgh-link-date=2025-04-22T17%3A12%3A22Z) to the `NXobject` definition.
> 
> In my opinion, these additions contribute towards your confusion. Again, the standard is based on the XML Schema. Also, I agree that XML Schema is not particularly easy to read or write.
> 

I don't think that necessarily is the thing that confuses me. These additions (one of which came from me, by the way) were made exactly for the same reason why I made this PR: to define groups/fields/attributes that can be used in every group (as every group inherits from `NXobject`). This also effectively blocks part of the namespace, as e.g. any field that ends on `_set` should follow the [definition](https://github.com/nexusformat/definitions/blob/main/base_classes/NXobject.nxdl.xml#L47) in `NXobject`. So far so good. 

What I wanted to achieve through this PR is similar: to define what it means to have the `@target` attribute in any field or group. 

> `NXobject` has been in the NXDL suite for a long time. During most of that time, it has been a [bare-bones definition](https://github.com/nexusformat/definitions/issues/112#issuecomment-42242455), inheriting any and all of its contents, rules, etc from the XSD Schema (nxdl.xsd). The NXDL is not an object-oriented standard, yet `NXobject` was added so that all (other) NXDL definitions would have a common starting point (its addition pre-dates the NeXus GitHub repo). Still, again, the standard is based on the XML Schema. Attributes (and other content) defined by the XML Schema are allowed and are not necessary to write into each and every NXDL file. The `target` attribute is a good example.
> 

My confusion is exactly in the last two sentences here. I just don't understand how that can be achieved through the XML Schema (i.e., the `nxdl.xsd`). Again, if I just move the definition of `<xs:attribute name="target" ...` to `basicComponent`, from my understanding of XML that means you can use "target" as an **XML attribute** for fields/group, but that doesn't mean that it is defined as a **NeXus attribute** (i.e., of type `attributeType`), see my [description above](https://github.com/nexusformat/definitions/pull/1558#issuecomment-2821494762).

What would help me greatly is how to understand why by adding it to the XSD, the latter case is achieved and why I don't need to define in the NXDL files anymore (which is what you are saying, if I understand you correctly).

PR #1559, opened 2025-04-30
Update NXprocess documentation

rayosborn:
A NIAC Telco discussion indicated that it was common practice to store the results of data analysis within the NXprocess group describing the operation. On reflection, I believe that this is a very valuable way of storing processed data, although I haven't used NXprocess groups in this way before. There are many advantages to encapsulating both the description of the process (program name, data, input parameters) with the resulting data.

This PR does not change anything substantively. It modifies the base class documentation to make this usage more evident to non-specialists and explicity adds NXparameter and NXdata groups to store input parameters and the resulting data. Strictly speaking, both these group additions are redundant because their base classes have been added to the NXobject base class and are therefore already allowed. However, I think it is useful to add them here to show typical ways in which NXprocess groups are structured. As with groups in all base classes, their actual inclusion in a NeXus file is optional, unless an application definition explicitly requires them.

### DIFF
--- a/base_classes/NXprocess.nxdl.xml
+++ b/base_classes/NXprocess.nxdl.xml
@@ -58,4 +58,14 @@
             The name will be numbered to allow for ordering of steps.
         </doc>
     </group>
+    <group type="NXparameters">
+        <doc>
+            Parameters used in performing the data analysis.
+        </doc>
+    </group>
+    <group type="NXdata">
+        <doc>
+            The data resulting from the operation.
+        </doc>
+    </group>
 </definition>
--- a/base_classes/NXprocess.nxdl.xml
+++ b/base_classes/NXprocess.nxdl.xml
@@ -26,7 +26,12 @@
     xsi:schemaLocation="http://definition.nexusformat.org/nxdl/3.1 ../nxdl.xsd"
     name="NXprocess" 
 	type="group" extends="NXobject">
-    <doc>Document an event of data processing, reconstruction, or analysis for this data.</doc>
+    <doc>
+        The :ref:`NXprocess` class describes an operation used to
+        process data as part of an analysis workflow, providing
+        information such as the software used, the date of the
+        operation, the input parameters, and the resulting data. 
+    </doc>
     <field name="program" type="NX_CHAR">
         <doc>Name of the program used</doc>
     </field>
--- a/base_classes/NXprocess.nxdl.xml
+++ b/base_classes/NXprocess.nxdl.xml
@@ -37,8 +37,8 @@
     </field>
     <field name="sequence_index" type="NX_POSINT">
         <doc>
-            Sequence index of processing, 
-            for determining the order of multiple **NXprocess** steps.  
+            Sequence index of processing, for determining the order of
+            multiple **NXprocess** steps.  
             Starts with 1.
         </doc>
     </field>
@@ -50,10 +50,10 @@
     </field>
     <group type="NXnote">
         <doc>
-            The note will contain information about how the data was processed
-            or anything about the data provenance. 
-            The contents of the note can be anything that the processing code 
-            can understand, or simple text.
+            The note will contain information about how the data was
+            processed or anything about the data provenance. The
+            contents of the note can be anything that the processing
+            code can understand, or simple text.
             
             The name will be numbered to allow for ordering of steps.
         </doc>


base_classes/NXprocess.nxdl.xml
```diff
@@ -45,12 +50,35 @@
     </field>
     <group type="NXnote">
         <doc>
-            The note will contain information about how the data was processed
-            or anything about the data provenance. 
-            The contents of the note can be anything that the processing code 
-            can understand, or simple text.
+            The note will contain information about how the data was
+            processed or anything about the data provenance. The
```
mkuehbach, line 54: Indeed that is exactly how we are using it also, could mention to folks using NXnote here for provenance.

base_classes/NXprocess.nxdl.xml
```diff
@@ -45,12 +50,35 @@
...
+        <doc>
+            Parameters used in performing the data analysis.
+        </doc>
+    </group>
+    <group type="NXdata">
+        <doc>
+            The data resulting from the operation.
+        </doc>
+    </group>
+    <attribute name="default">
```
mkuehbach, line 71: This is not required anymore because default has been added to NXobject, NXprocess inherits from NXobject

mkuehbach (changes requested):
Thanks for the initiative lgtm, one important suggestion for improval

base_classes/NXprocess.nxdl.xml
rayosborn: Thanks for pointing that out. I'm happy to remove it.

PeterC-DLS:
Can you comment on how this relates to https://manual.nexusformat.org/applying-nexus.html#processed-data? It seems to me that the original usage was to place one or more `NXprocess` groups in an `NXentry` to document how the processed data in that entry was derived. This is how it was used in Dawn's [processing](https://diamondlightsource.atlassian.net/wiki/spaces/DT/pages/1378476/Batch+2D+Powder+Diffraction+Data+Reduction+-+Processing+Perspective+DAWN+1.7.1) ([video is here](https://dawnsci.org/using-dawn/videos/#processing)) to write the output NeXus files.

sanbrock:
In fact, the prescrition of having separate entry foreach data processing step also suggests that if you want to specify restrictions/requirement on those steps, one must create seoarate Application definition for each of them.
While this can be useful, one can also imagine the use case where a single Application definition prescribes multiple data processing steps. Hence, I suggest to clarify in the documentation that multiple processing steps can be organised in multiple entries, or sequenced in a single entry.

PeterC-DLS:
> Multiple processing steps must have a separate entry each.
Seems overly restrictive and should be relaxed!

mkuehbach (changes requested):
I am okay with this PR getting merged before #1422 but then this branch needs to be updated with ``main`` first.

base_classes/NXprocess.nxdl.xml
```diff
@@ -45,12 +50,22 @@
     </field>
     <group type="NXnote">
         <doc>
-            The note will contain information about how the data was processed
-            or anything about the data provenance. 
-            The contents of the note can be anything that the processing code 
-            can understand, or simple text.
+            The note will contain information about how the data was
+            processed or anything about the data provenance. The
+            contents of the note can be anything that the processing
```
mkuehbach, line 55: "the processing code", you mean the program that is will be written under ``program`` if so might be good to mak a reference in the docstring to this ``program`` concept ?

base_classes/NXprocess.nxdl.xml
```diff
@@ -26,14 +26,19 @@
...
+        operation, the input parameters, and the resulting data. 
+    </doc>
     <field name="program" type="NX_CHAR">
         <doc>Name of the program used</doc>
     </field>
     <field name="sequence_index" type="NX_POSINT">
         <doc>
-            Sequence index of processing, 
-            for determining the order of multiple **NXprocess** steps.  
+            Sequence index of processing, for determining the order of
```
mkuehbach, line 40: NXprocess essentially details a step of processing along a workflow. Multiple NXprocess instances built the workflow, now what is the recommendation when in an appdef we have NXprocess instances distributed all over the tree? Does the sequence_index refer only to the sequence of those NXprocess instances that are siblings, i.e.
entry/process1,
entry/process2,
entry/process3
entry/some_more_involved_stuff/process1
entry/some_more_involved_stuff/process2

Is there any connection between sequence_index in entry and those in entry/some_more_involved_stuff or not.

Maybe an alternative is to introduce a concept:
identifier_workflow_node which would inherit from NXobject/identifierNAME and could be used to assign each NXprocess instance in an appdef a node UUID.
And then to add a concept previous and next, similar like we have it in

mkuehbach:
```previous(NX_CHAR):
     doc: |
       A list of identifier_workflow_node values that specifies the workflow graph whereby
       this NXprocess is connected to other instances of NXprocess.
       Use "." if there is no dependency.
       
       Example Y-junction on the workflow graph, process1 and process2 feeding into process3:
       process1:NXprocess
          identifier_workflow_node = "uuid1"
          previous = ["."]
       process2:NXprocess
          identifier_workflow_node = "uuid2"
          previous = ["."]
       process3:NXprocess
          identifier_workflow_node = "uuid3"     
          previous = ["uuid1", "uuid2"]    
```
This is a depends_on like for workflows, sequence_index should be preferred choice to use if workflow is linear throughout.

base_classes/NXprocess.nxdl.xml
mkuehbach, line 1: Agree with @PeterC-DLS that NeXus documentation should be added, it should be okay to have a bunch of NXprocesses dispersed in branches below an entry, although, indeed not every case needs this

mkuehbach:
telco today @phyy-nx: @rayosborn please edit this paragraph and add to this PR:
Also in [elsewhere](https://manual.nexusformat.org/applying-nexus.html#processed-data) in the manual

rayosborn:
I have suggested changes to the manual, which better reflect the ways in which the NXprocess group could be used. One thing that we didn't discuss just now in the Telco is how the NXdata group produced by the NXprocess is identified if it is not stored within the NXprocess group. There is no field or link within the NXprocess group that specifies which NXdata group was generated. In the NXtomoproc application definition, the only way to tell that the NXdata group is the result of the action described in the NXprocess group is in the tag descriptions, which are, of course, not stored in the file itself. In this case, there is only one NXdata group, but there could be more.

I don't want to delay this PR any further, but I think in the future, we should consider making a recommendation that, if an application definition wants the NXdata group to be outside the NXprocess group, there should either be a link to the NXdata group within the NXprocess group (or vice versa) or that the path to the NXdata group should be stored as a field in the NXprocess group.

base_classes/NXprocess.nxdl.xml
mkuehbach: Was discussed at the 2025/06/16 telco and edited

base_classes/NXprocess.nxdl.xml
mkuehbach: Agreed, good suggestion

mkuehbach (approved):

mkuehbach (changes requested):
Rebase again nexusformat/definitions.git main branch. Other than this, approved from my side.

PeterC-DLS (approved):
LGTM

phyy-nx:
Hi folks, as discussed on the Telco, this PR can now me moved to an online vote.  Please vote using emojis (👍  for yes, 👎  for now, anything else e.g. 👀  for abstain).  Voting will close in 2 weeks.  Thanks!

rayosborn:
@PeterC-DLS, you suggested the possibility of adding some text about adding a "default" attribute to the NeXus manual text, but I decided in the end that it made more sense to add the attribute explicitly to the NXprocess group itself. Please could you confirm that my change to the PR is syntactically correct as well as appropriate. There have already been a few votes concerning this PR, so I will revert the change if anyone thinks it has been added too late.

phyy-nx:
The vote has passed, thank you.  @mkuehbach has requested a rebase, then it can be merged

rayosborn:
The rebase has been completed.

PR #1560, opened 2025-05-01
Update NXparameter documentation

rayosborn:
The name of this base class suggests that it would be the ideal repository to store the results of any parameter optimization, such as least-squares fitting. To facilitate its use in this context, it is helpful to standardize a number of attributes for each parameter, which are common to most, if not all, optimizaiton procedures. For a number of years, NeXpy has been using NXparameter groups to define the parameters for a specific function, such as a Gaussian, in conjunction with the LMFIT package (https://lmfit.github.io/lmfit-py/), which calls them "models." If multiple functions are fitted simulaneously to a data set, multiple NXparameter groups are stored in a NXprocess group, which also defines the program used to produce the optimization.

An example of a NXparameter group is given here:
```
fit:NXprocess
  parameters:NXparameters
    @model = 'Lorentzian'
    amplitude = 1301.0084322223488
      @error = 83.04682658748045
      @initial_value = 1918.64
      @max = 'inf'
      @min = '-inf'
      @vary = True
   center = 109.90841596329723
      @error = 0.151393986221809
      @initial_value = 109.5
      @max = 'inf'
      @min = '-inf'
      @vary = True
.
.
.
```
The purpose of this PR is not to confine NXparameter groups to this particular use case, but to standardize the attributes when it applies. Others are welcome to suggest other uses of this group.

Until now, the NXparameter group has been an empty container, so I don't think this PR changes the standard in any substantive way - it just expands the documentation - so I don't believe a NIAC vote is required. However, it is probably still worth discussion at a Telco to get feedback.

### DIFF
--- a/base_classes/NXparameters.nxdl.xml
+++ b/base_classes/NXparameters.nxdl.xml
@@ -3,7 +3,7 @@
 <!--
 # NeXus - Neutron and X-ray Common Data Format
 # 
-# Copyright (C) 2008-2024 NeXus International Advisory Committee (NIAC)
+# Copyright (C) 2008-2025 NeXus International Advisory Committee (NIAC)
 # 
 # This library is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
@@ -26,11 +26,74 @@
     xsi:schemaLocation="http://definition.nexusformat.org/nxdl/3.1 ../nxdl.xsd"
     name="NXparameters" 
 	type="group" extends="NXobject">
-    <doc>Container for parameters, usually used in processing or analysis.</doc>
+    <doc>
+        Container for parameters used in processing or analysing data.
+
+        This base class can be used in a number of different contexts,
+        and is allowed in any other base class. However, a common use
+        case is to store the results of optimizing parameters, for
+        example, by least-squares fitting, so a number of attributes
+        have been defined for this purpose. Normally, this group will be
+        stored in a :ref:NXprocess group, which defines the program used
+        to process these parameters. If multiple functions are used to
+        analyze the data, there can be additional NXparameter groups for
+        each one.        
+    </doc>
+    <attribute name="model">
+        <doc>
+            The name of the model used in optimizing the data. Fitting
+            packages such as LMFIT (https://lmfit.github.io/lmfit-py/)
+            provide models, which instantiate functions to be fitted to
+            the data. If this attribute is provided, it is assumed that
+            all the parameters in this group are associated with this
+            model.
+        </doc>
+    </attribue>
     <field name="TERM" minOccurs="0" maxOccurs="unbounded" nameType="any">
-        <!-- maxOccurs="unbounded" is intended but is not allowed by current syntax -->
-        <doc>A parameter (also known as a term) that is used in or results from processing.</doc>
-        <attribute name="units"/>
+        <doc>
+            A parameter (also known as a term) that is used in or
+            results from processing. 
+        </doc>
+        <attribute name="units" />
+        <attribute name="error" type=NX_NUMBER>
+            <doc>
+                The standard deviation resulting from data optimization.
+            </doc>
+        </attribute>
+        <attribute name="expression">
+            <doc>
+                A string representing an expression that can be used to
+                relate the parameter to another parameter's value. The
+                format of this string is dependent on the program used
+                to optimize the parameters and is not specified by
+                NeXus.
+            </doc>
+        </attribute>
+        <attribute name="initial_value" type=NX_NUMBER>
+            <doc>
+                The initial value of the parameter.
+            </doc>
+        </attribute>
+        <attribute name="max" type=NX_NUMBER>
+            <doc>
+                The upper bound of the parameter value.
+            </doc>
+        </attribute>
+        <attribute name="min" type=NX_NUMBER>
+            <doc>
+                The lower bound of the parameter value.
+            </doc>
+        </attribute>
+        <attribute name="vary" type=NX_BOOLEAN>
+            <doc>
+                True if the parameter was varied in the fitting
+                procedure.
+            </doc>
+            <enumeration>
+                <item value=True />
+                <item value=False />
+            </enumeration
+        </attribute>
     </field>
 </definition>
 


PeterC-DLS:
It's worth taking into consideration the MPES proposal where there's [`NXfit`](https://github.com/nexusformat/definitions/blob/cfb6b623713ae0d5219a3925373d003e6fddd9a4/base_classes/NXfit.nxdl.xml), [`NXfit_function`](https://github.com/nexusformat/definitions/blob/cfb6b623713ae0d5219a3925373d003e6fddd9a4/base_classes/NXfit_function.nxdl.xml) and other base classes

base_classes/NXparameters.nxdl.xml
```diff
@@ -26,11 +26,69 @@
...
+        and is allowed in any other base class. However, a common use
+        case is to store the results of optimizing parameters, for
+        example, by least-squares fitting, so a number of attributes
+        have been defined for this purpose. Normally, this group will be
+        stored in a :ref:NXprocess group, which defines the program used
+        to process these parameters. If multiple functions are used to
+        analyze the data, there can be additional NXparameter groups for
+        each one.        
+    </doc>
+    <attribute name="model">
```
mkuehbach, line 42: Personally would make this a field as then you can add version of the model for an attribute as attribute of an attribute is not possible in NeXus, in FAIRmat sometimes we use e.g. as a field "algorithm" same idea like model.

mkuehbach: Exactly these were also my thoughts

mkuehbach (approved):

base_classes/NXparameters.nxdl.xml
mkuehbach: Maybe then we should mention algorithm as well in the docstring and then that is fine

base_classes/NXparameters.nxdl.xml
```diff
@@ -26,11 +26,69 @@
...
     <field name="TERM" minOccurs="0" maxOccurs="unbounded" nameType="any">
-        <!-- maxOccurs="unbounded" is intended but is not allowed by current syntax -->
-        <doc>A parameter (also known as a term) that is used in or results from processing.</doc>
-        <attribute name="units"/>
+        <doc>
+            A parameter (also known as a term) that is used in or
+            results from processing. 
+        </doc>
+        <attribute name="units" />
+        <attribute name="error" type="NX_NUMBER">
```
mkuehbach, line 58: We have the reserved suffix "_errors" why not use this?
Like parameter called "alpha" and given that errors exists implicitly we also have "alpha_errors"
https://manual.nexusformat.org/datarules.html

base_classes/NXparameters.nxdl.xml
```diff
@@ -26,11 +26,69 @@
...
+                to optimize the parameters and is not specified by
+                NeXus.
+            </doc>
+        </attribute>
+        <attribute name="initial_value" type="NX_NUMBER">
+            <doc>
+                The initial value of the parameter.
+            </doc>
+        </attribute>
+        <attribute name="max" type="NX_NUMBER">
```
mkuehbach, line 77: The "TERM"'s type is currently NX_CHAR which kicks in as the default what if TERM is a number quantity, to my understanding if you set type=NX_NUMBER than TERM can only be a number but what if the default NX_CHAR kicks in? Would should a verifier do if one gives still an NX_NUMBER. Is there anywhere in the NeXus docs a statement that explains that when one does not give a type one states effectively "I do not want the type to be constrained or does one say implicitly it will be NX_CHAR". My understanding is that the second case is the situation right now.

base_classes/NXparameters.nxdl.xml
```diff
@@ -26,11 +26,69 @@
...
+            <doc>
+                The upper bound of the parameter value.
+            </doc>
+        </attribute>
+        <attribute name="min" type="NX_NUMBER">
```
mkuehbach, line 82: Rather suggestion: min and max I agree that these are  useful as summary statistics for especially quantities would it be an (additional idea) to make _min, _max reserved suffixes that would cover for this?

mkuehbach (changes requested):

base_classes/NXparameters.nxdl.xml
rayosborn: I agree this could cause confusion to validators. I have added a type of "NX_CHAR_OR_NUMBER".

base_classes/NXparameters.nxdl.xml
rayosborn: If you are suggesting that these should be fields, I would prefer they remain as attributes for the reasons I gave [above](https://github.com/nexusformat/definitions/pull/1560#discussion_r2071684581).

base_classes/NXparameters.nxdl.xml
lukaspie: See the [now approved change](https://github.com/nexusformat/definitions/pull/1424/files#diff-e0aeb0a680e5a38f3cc38a586d6abfb21428b240d5c01f99b6769aea542ac9b3) to `NXparameters/TERM` in #1424 which renames it to uppercase `PARAMETER` and makes it an `NX_CHAR_OR_NUMBER` (as suggested by @PeterC-DLS).

base_classes/NXparameters.nxdl.xml
```diff
@@ -26,11 +26,68 @@
...
-        <attribute name="units"/>
+    <doc>
+        Container for parameters used in processing or analysing data.
+
+        This base class can be used in a number of different contexts,
+        and is allowed in any other base class. However, a common use
+        case is to store the results of optimizing parameters, for
+        example, by least-squares fitting, so a number of attributes
+        have been defined for this purpose. Normally, this group will be
+        stored in a :ref:`NXprocess group`, which defines the program
```
lukaspie, line 37:
```suggestion
        stored in a :ref:`NXprocess` group`, which defines the program
```

lukaspie:
> It's worth taking into consideration the MPES proposal where there's [`NXfit`](https://github.com/nexusformat/definitions/blob/cfb6b623713ae0d5219a3925373d003e6fddd9a4/base_classes/NXfit.nxdl.xml), [`NXfit_function`](https://github.com/nexusformat/definitions/blob/cfb6b623713ae0d5219a3925373d003e6fddd9a4/base_classes/NXfit_function.nxdl.xml) and other base classes

I think the various attributes that are added here like `@error`, `@initial_value`, `@max`, `@min`, `@vary` would greatly improve `NXfit` and the associated base classes. What's a bit different to how we have implemented is the additional `@model` attribute for which we are using [`NXfit_function` within `NXfit`](https://github.com/FAIRmat-NFDI/nexus_definitions/blob/fairmat-2024-mpes/base_classes/NXfit.nxdl.xml), so that may potentially be redundant.

base_classes/NXparameters.nxdl.xml
mkuehbach: The third backtick should be removed.

mkuehbach:
Indeed there are similarities between NXfit and suggestions made here. Personally, I think though that NXparameters should intentionally not restrict itself to fitting only but rather be a base class for where people can store parameters of algorithms irrespective of which type of algorithms these are. NXfit is maybe a specialization of NXparameters as an idea?

phyy-nx:
@rayosborn to consider if the new [NXfit](https://github.com/nexusformat/definitions/pull/1424/files#diff-0bd69f0a4c1bf5a303bd1c85a4fbf265bc64416513d6cec9af6212934b1274d8) class serves the same function so it's not worth updating NXparameters.

Also update doc here so it's clear that fitting is just an example use case.

base_classes/NXparameters.nxdl.xml
```diff
@@ -26,11 +26,69 @@
...
+        <doc>
+            The name of the model used in optimizing the parameter
+            values. Fitting packages such as LMFIT
+            (https://lmfit.github.io/lmfit-py/) provide models, which
+            instantiate functions to be fitted to the data. If this
+            attribute is provided, it is assumed that all the parameters
+            in this group are associated with this model.
+        </doc>
+    </attribute>
+    <field name="PARAMETER" type="NX_CHAR_OR_NUMBER" nameType="any">
```
lukaspie, line 53: In the [current state of the base class](https://github.com/nexusformat/definitions/blob/main/base_classes/NXparameters.nxdl.xml#L30), this is already
```
<field name="PARAMETER" nameType="any" type="NX_CHAR_OR_NUMBER" units="NX_ANY" minOccurs="0" maxOccurs="unbounded">
```
I think that aligns well with your idea here?

mkuehbach (approved):
A reasonable set of changes

base_classes/NXparameters.nxdl.xml
```diff
@@ -26,11 +26,69 @@
...
+        <attribute name="expression">
+            <doc>
+                A string representing an expression that can be used to
+                relate the parameter to another parameter's value. The
+                format of this string is dependent on the program used
+                to optimize the parameters and is not specified by
+                NeXus.
+            </doc>
+        </attribute>
+        <attribute name="initial_value" type="NX_NUMBER">
```
lukaspie, line 73: These attributes are very helpful and I support adding them to the `PARAMETER` field in the base class. As discussed, I would just suggest that we align the attributes used in `NXfit_function` that basically do the same: https://github.com/nexusformat/definitions/blob/main/base_classes/NXfit_function.nxdl.xml#L58

Would be happy to adjust `NXfit_function` accordingly if we agree on the names of these attributes.
We should probably end up with something like this:
- initial value
- min_value / min (we just choose one name)
- max_value / max (we just choose one name)
- vary / fixed (one of these)

base_classes/NXparameters.nxdl.xml
```diff
@@ -26,11 +26,69 @@
...
+        resulting from the process defined by the parent group. However,
+        this base class can also be added to other groups for use in
+        other contexts.
+        
+        The base class defines a number of attributes that can be used
+        if the parameter values result from, e.g., fitting a function or
+        model to the measured data. If multiple functions are used,
+        there can be additional NXparameter groups for each one.        
+    </doc>
+    <attribute name="model">
```
lukaspie, line 43: Similar concept to [`NXfit_function/description`](https://github.com/nexusformat/definitions/blob/main/base_classes/NXfit_function.nxdl.xml#L44), maybe could be aligned. But probably fine to keep both ideas.

base_classes/NXparameters.nxdl.xml
lukaspie: Probably can be addressed through a rebase

phyy-nx: Yep, just needs a rebase, @rayosborn will do that.

base_classes/NXparameters.nxdl.xml
phyy-nx: @lukaspie will make a PR that points to this PR to fix NXfit_function to use vary instead of fixed
@rayosborn will change min/max to min_value/max_value and fix the conflicts

rayosborn:
I amended the documentation to add:

"Although this base class can be used to store any kind of parameter, one possible use case is to store parameters that are refined by a fitting function or model. A number of attributes have been defined to store metadata associated with such a refinement."

I hope this satisfies those who are concerned that viewers might think the group is exclusively for fitting parameters.

phyy-nx:
As discussed on the Telco, this PR is ready for a vote, with the note that we are waiting on a change from @lukaspie to change NXfit_function to use vary instead of fixed.

Please vote using emojis, 👍 for yes, 👎 for no, and anything else e.g. 👀 for abstain.  Voting will close in 2 weeks.

Thanks!

PeterC-DLS:
I thought during the telco Markus objected to the `*_value` names and that it was better to use `min`, `max`, and `initial`.

base_classes/NXparameters.nxdl.xml
```diff
@@ -26,11 +26,75 @@
...
+        order to contain parameters that are either inputs to or
+        resulting from the process defined by the parent group. However,
+        this base class can also be added to other groups for use in
+        other contexts.
+        
+        Although this base class can be used to store any kind of
+        parameter, one possible use case is to store parameters that are
+        refined by a fitting function or model. A number of attributes
+        have been defined to store metadata associated with such a
+        refinment.
```
mkuehbach: "refinment" > "refinement"

this change does not affect the ongoing vote

mkuehbach:
> I thought during the telco Markus objected to the `*_value` names and that it was better to use `min`, `max`, and `initial`.

Indeed, I asked for "_value" be dropped.

"min", "max" and other suggestive names for number quantities which are essentially summary statistics could be made a common for quantities. My suggestion is --- solely to support not having the release date for NXDL2025 blocked --- accept the PR as is and make edits a future issue for e.g. NXDL2026.

PeterC-DLS:
@mkuehbach the problem is that there will be files written using these names - do you deprecated the names straight away? It's worth getting this in a reasonable shape for the release.

rayosborn:
I appreciate your concern, @PeterC-DLS, but the two people, who might be affected with any backward-compatibility issues, i.e., @mkuehbach and me, both support this PR. I'm not sure either of us would be helped by delaying this any further. Some NeXus issues hang around for years and never get resolved by the NIAC; I would prefer this is not one of them.

mkuehbach:
> As discussed on the Telco, this PR is ready for a vote, with the note that we are waiting on a change from @lukaspie to change NXfit_function to use vary instead of fixed. Update: this vote includes #1594 from @lukaspie
> 
> Please vote using emojis, 👍 for yes, 👎 for no, and anything else e.g. 👀 for abstain. Voting will close in 2 weeks.
> 
> Thanks!

2w passed but less than 14 votes...

> (...) I'm not sure either of us would be helped by delaying this any further. Some NeXus issues hang around for years and never get resolved by the NIAC; I would prefer this is not one of them.

I agree with @rayosborn

Question for me is though --- given insufficient votes --- what will happen now???
1.) Will we just accept this PR and #1594 as is?
2.) Is there another round of votes for this PR and #1594 as they are right now?
3.) Should we just remove the `_value` suffix from here and in #1594 for the concepts?
      `min_value`, `max_value`, and `initial_value`, respectively and will this then need a second voting round?
4.) like 3.) but does not need a second voting round?

NIAC please decide

phyy-nx:
The PR did not get enough votes.  Interested parties should confer and get the pull requests aligned with all edits merged in before we try again.

base_classes/NXparameters.nxdl.xml
mkuehbach: Edited

rayosborn:
I have uploaded a revised version of the NXfit_function base class with attributes duplicated in the NXparameter base class removed.

phyy-nx:
After revision, this PR is ready for another vote.  Please vote using emojis, 👍 for yes, 👎 for no, and anything else e.g. 👀 for abstain.  Voting will close in 2 weeks.

Thanks!

PeterC-DLS:
This needs to be rebased off main to pass CI

PR #1561, opened 2025-05-01
Add NXgoniometer base class

rayosborn:
This PR adds a new base class that allows the goniometer angles used in a (usually single crystal) measurement to be defined according to a variety of naming conventions defined by an NXcite group. This group is used unofficially in the NXRefine package used for reducing single crystal diffuse scattering (https://nexpy.github.io/nxrefine). In recent years, the NeXus convention has been to define a set of NXtransformations that obviate the need to define the meaning of such angles. This has the advantate that there is no universally accepted standard for naming goniometer angles. However, there are two disadvantages to this approach:

1. It is less intuitive for the user to check these angles against their own records of the experiment.
2. It assumes that the transformations defined in the file are the correct ones.

As an example of the second problem, in NXRefine, we discovered that out initial assumptions about the chain of transformation matrices was incorrect because the detector was decoupled from any two-theta arm. In stable experimental configurations, the transformation approach should work, but when experimenting with different goniometer and detector configurations, as well as different data reduction procedures, there are advantages in simply storing the angles as recorded by the beamline motors.

As a new base class, this requires formal approval by the NIAC.

### DIFF
--- a/base_classes/NXgoniometer.nxdl.xml
+++ b/base_classes/NXgoniometer.nxdl.xml
@@ -1,0 +1,76 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?xml-stylesheet type="text/xsl" href="nxdlformat.xsl" ?>
+<!--
+# NeXus - Neutron and X-ray Common Data Format
+# 
+# Copyright (C) 2008-2025 NeXus International Advisory Committee (NIAC)
+# 
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 3 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+#
+# For further information, see http://www.nexusformat.org
+-->
+<definition xmlns="http://definition.nexusformat.org/nxdl/3.1" category="base"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://definition.nexusformat.org/nxdl/3.1 ../nxdl.xsd"
+	name="NXgoniometer" 
+	type="group" extends="NXobject"
+	>
+
+	<doc>
+		Goniometer angles used during a measurement. The goniometers
+		used at neutron and x-ray facilities are typically four-circle
+		or six-circle goniometers, or variants of either one. There are
+		a number of conventions defining the names of these angles, so
+		the group should contain a :ref:`NXcite` group, which provides
+		links relevant to this goniometer.
+	</doc>
+
+    <group name="convention" type="NXcite" >
+        <doc>
+            A reference to a publication or online documentation that
+            describe the convention used in this group to define the
+            names of angles.
+        </doc>
+    </group>
+    <field name="goniometer">
+        <doc>
+            The common name for the type of goniometer used in these
+            measurements.
+        </doc>
+    </field>
+	<field name="chi" type="NX_FLOAT" units="NX_ANGLE" />
+	<field name="omega" type="NX_FLOAT" units="NX_ANGLE" />
+	<field name="phi" type="NX_FLOAT" units="NX_ANGLE" />
+	<field name="theta" type="NX_FLOAT" units="NX_ANGLE" />
+	<field name="xs" type="NX_FLOAT" units="NX_LENGTH">
+        <doc>
+            The translation of the sample from the center of rotation
+            along the x-axis defined by the local coordinate system.
+        </doc>
+    </field>
+	<field name="ys" type="NX_FLOAT" units="NX_LENGTH">
+        <doc>
+            The translation of the sample from the center of rotation
+            along the y-axis defined by the local coordinate system.
+        </doc>
+    </field>
+	<field name="zs" type="NX_FLOAT" units="NX_LENGTH">
+        <doc>
+            The translation of the sample from the center of rotation
+            along the z-axis defined by the local coordinate system.
+        </doc>
+    </field>
+
+</definition>
--- a/base_classes/NXgoniometer.nxdl.xml
+++ b/base_classes/NXgoniometer.nxdl.xml
@@ -29,19 +29,27 @@
 	>
 
 	<doc>
-		Goniometer angles used during a measurement. The goniometers
-		used at neutron and x-ray facilities are typically four-circle
-		or six-circle goniometers, or variants of either one. There are
-		a number of conventions defining the names of these angles, so
-		the group should contain a :ref:`NXcite` group, which provides
-		links relevant to this goniometer.
+		Goniometer rotation angles and sample translations used during a
+		measurement. The goniometers used at neutron and x-ray
+		facilities are typically four-circle or six-circle goniometers,
+		or variants of either one. There is no universally recognized
+		naming convention for angles and translations, so the group
+		should contain a :ref:`NXcite` group, which describes how the
+		angles are defined.
 	</doc>
 
-    <group name="convention" type="NXcite" >
+    <group name="convention" type="NXcite">
         <doc>
             A reference to a publication or online documentation that
             describe the convention used in this group to define the
-            names of angles.
+            names of angles and sample translations.
+        </doc>
+    </group>
+    <group type="NXcoordinate_system">
+        <doc>
+            The coordinate system used by the translation fields. This
+            group should be present if these fields do not use the
+            default NeXus Coordinate System.
         </doc>
     </group>
     <field name="goniometer">
--- a/base_classes/NXgoniometer.nxdl.xml
+++ b/base_classes/NXgoniometer.nxdl.xml
@@ -50,27 +50,17 @@
             measurements.
         </doc>
     </field>
-	<field name="chi" type="NX_FLOAT" units="NX_ANGLE" />
-	<field name="omega" type="NX_FLOAT" units="NX_ANGLE" />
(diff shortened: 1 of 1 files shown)


base_classes/NXgoniometer.nxdl.xml
```diff
@@ -0,0 +1,74 @@
...
+            group should be present if these fields do not use the
+            default NeXus Coordinate System.
+        </doc>
+    </group>
+    <field name="goniometer">
+        <doc>
+            The common name for the type of goniometer used in these
+            measurements.
+        </doc>
+    </field>
```
woutdenolf, line 60: Does the "type" refer to the generic goniometer types

- "Eulerian cradle"
- "Kappa geometry"
- "4-circle goniometer"
- "6-axis motorized goniometer"
- ...

Or does it refer to specific manufacturer types?

- "Huber 420"
- "Bruker D8 Kappa"
- "Rigaku AFC11"
- "Philips PW1050"
- "Oxford Diffraction Gemini"
- ...

Or both? Would be good to specify in the doc string.

base_classes/NXgoniometer.nxdl.xml
```diff
@@ -0,0 +1,74 @@
...
+	<field name="ROTATION_ANGLE" type="NX_FLOAT" units="NX_ANGLE" nameType="any">
+        <doc>
+            One of the goniometer rotation angles defined by the
+            specified convention.
+        </doc>
+    </field>
```
woutdenolf, line 66:
```suggestion
    <field name="ROTATION_ANGLE" type="NX_FLOAT" units="NX_ANGLE" nameType="any">
        <doc>
            One of the goniometer rotation angles defined by the
            specified convention.
        </doc>
    </field>
```

woutdenolf:
So we have several `ROTATION_ANGLE` and `SAMPLE_TRANSLATION` fields (can have any name). When I read such a file, how do I know the meaning of these fields? I'm guessing the fields `convention` and/or `goniometer` would provide this mean? At least for humans, I'm not sure how this would work for software.

base_classes/NXgoniometer.nxdl.xml
```diff
@@ -0,0 +1,74 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?xml-stylesheet type="text/xsl" href="nxdlformat.xsl" ?>
+<!--
+# NeXus - Neutron and X-ray Common Data Format
+# 
+# Copyright (C) 2008-2025 NeXus International Advisory Committee (NIAC)
```
PeterC-DLS, line 6: A completely new class didn't time travel

base_classes/NXgoniometer.nxdl.xml
```diff
@@ -0,0 +1,74 @@
...
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+#
+# For further information, see http://www.nexusformat.org
+-->
+<definition xmlns="http://definition.nexusformat.org/nxdl/3.1" category="base"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://definition.nexusformat.org/nxdl/3.1 ../nxdl.xsd"
+	name="NXgoniometer" 
+	type="group" extends="NXobject"
```
PeterC-DLS, line 28: Should this be an `NXcomponent`?

base_classes/NXgoniometer.nxdl.xml
PeterC-DLS: If this was an `NXcomponent` then `NXfabtrication` would have the model name

base_classes/NXgoniometer.nxdl.xml
```diff
@@ -0,0 +1,74 @@
...
+	</doc>
+
+    <group name="convention" type="NXcite">
+        <doc>
+            A reference to a publication or online documentation that
+            describe the convention used in this group to define the
+            names of angles and sample translations.
+        </doc>
+    </group>
+    <group type="NXcoordinate_system">
```
PeterC-DLS, line 48: An `NXcomponent` would have a `depends_on` chain and `NXtransformations` group so this is unnecessary

base_classes/NXgoniometer.nxdl.xml
```diff
@@ -0,0 +1,74 @@
...
+            measurements.
+        </doc>
+    </field>
+	<field name="ROTATION_ANGLE" type="NX_FLOAT" units="NX_ANGLE" nameType="any">
+        <doc>
+            One of the goniometer rotation angles defined by the
+            specified convention.
+        </doc>
+    </field>
+    <field name="SAMPLE_TRANSLATION" type="NX_FLOAT" units="NX_LENGTH" nameType="any">
```
PeterC-DLS, line 67: Is this (vector?) field necessary? Can this be recorded by `NXtransformations` in `NXsample`?

base_classes/NXgoniometer.nxdl.xml
rayosborn: @PeterC-DLS, it had occurred to me that this should be a sub-class of NXcomponent, so I will make that change. @woutdenolf, I was thinking more of your first list. If you think it is possible to define a reasonably comprehensive enumeration (perhaps keeping it open), I am open to suggestions. Those of you managing beam lines will probably do a better job of selecting those names than I can.

base_classes/NXgoniometer.nxdl.xml
rayosborn: Fixed.

base_classes/NXgoniometer.nxdl.xml
rayosborn: I have never used NXtransformations, so you may well be right that this is the way you should handle it if you do. My personal preference is that we encapsulate all the parameters that are required to transform the data from instrumental to reciprocal space coordinates in one place. This group contains all the information we need to construct the transformations described, e.g., [here](https://nexpy.github.io/nxrefine/introduction.html#experimental-geometry). I know that NXtransformations handle that process in other implementations, such as NXmx, but that is not how we do it.

rayosborn:
> So we have several `ROTATION_ANGLE` and `SAMPLE_TRANSLATION` fields (can have any name). When I read such a file, how do I know the meaning of these fields? 

@woutdenolf, defining what the goniometer names mean was the purpose of the NXcite group. For example, in the case of the NXRefine package, it would refer to [this page](https://nexpy.github.io/nxrefine/introduction.html#experimental-geometry), where the rotation angle names chi, omega, phi, and theta would be explained.

rayosborn:
> I can see how this group is useful to gather together the `NXpositioner` groups (or solely, links to their value fields) from each goniometer arm. However these fields are also linked to fields in `NXtransformations` groups for the sample and any detectors - consequently, they have corresponding attributes (`@vector`, `@transformation_type`, etc). It is an external matter to NeXus if these attributes are wrong: any analysis program should be able to override, ignore, or correct them.

@PeterC-DLS, my comment [above](https://github.com/nexusformat/definitions/pull/1561#discussion_r2071640809) applies here as well. For the reasons I gave at the top, I prefer that the analysis code decides how to handle transformations, although I am aware of reasons why the NXtransformations chain was originally adopted.

base_classes/NXgoniometer.nxdl.xml
rayosborn: I think it is probably safer that I leave the NXcoordinate_system out to avoid confusion with those using a NXtransformations chain. The convention adopted by this base class is that the meaning of these parameters is defined by the NXcite reference.

phyy-nx:
When I was first reviewing NXmx I wondered why there was no formal NXgoniometer class and why we instead use a series of NXtransformations under the NXsample group.  That always seemed unintuitive.  Having a specific NXgoniometer class makes sense to me, and if approved, we might want to incorporate it into NXmx as an alternative way of representing a goniometer.  I would want to review the axes specifications with some of my crystallographer colleagues though, since in XFEL-land where I have typically lived, we don't have goniometers.  Tagging @graeme-winter for example.

graeme-winter:
Also:

> we might want to incorporate it into NXmx as an alternative way of representing a goniometer

NO NO NO no no please NO!

We already have enough corner cases to support in analysis software without actively trying to introduce _more_

phyy-nx:
Heh, roger that :)

rayosborn:
@graeme-winter, the existence of a base class does not mean people are forced to use it, just as I am not forced to use the NXmx application definition, even though there is some overlap between single crystal diffuse scattering and macromolecular crystallography. Also, there shouldn't be any argument about the names of rotation angles, because the proposed base class allows them to be defined by different conventions, preferably with references to online documentation. I realize that this means our files are incompatible with a Dials-based analysis, just as they are incompatible with many other applications, but once (or perhaps assuming) this is approved, it will form the basis of an application definition describing how the NXRefine package analyzes the data.

base_classes/NXgoniometer.nxdl.xml
woutdenolf: +1 for NXfabtrication from NXcomponent so the "goniometer" would be indeed the first list.

I have little experience with goniometers so I'll leave it up to the specialists to come up with an open enumeration but I guess "3-circle goniometer" and "4-circle goniometer" are in there. In any case I think we need the examples so we understand what this field is for.

PeterC-DLS:
I think a positioner group base class is more general and can encapsulate other equipment that position samples and/or detectors (robot arms, hexpods, etc).

By not recommending that physical meanings of the labelled positions recorded in some mathematical way (adding NXtransformations attributes to the fields) means that this class is not specified tightly enough to be machine readable as there's no standard to parse `NXcite` - it's not generally interoperable.

phyy-nx:
The PR did not get enough votes but a consensus here is highly desirable.  We will discuss the best course of action on the next Telco.

rayosborn:
@phyy-nx, I think we need to have a rule that once something has been put to a vote, there is no further discussion while the vote proceeds. Once a PR has been submitted, there should be a reasonable period for people to make comments online alongside discussions at two or more Telcos. Once you, as chairman, feel that there is a consensus, then you call a vote and all online discussions should stop until the vote is over. 

I believe we had a consensus at the last Telco, but this vote was sunk by two negative comments written after the Telco was over. I attempted to answer both of their criticisms but have yet to receive a reply to either one. That is why I believe that the time for debate has to be before you call a vote, because comments during the voting period just cause confusion and doubt, rendering all our previous discussions pointless.

PR #1562, opened 2025-05-02
Remove extraneous transformation type attributes from beam directions

PeterC-DLS:
Also add note on beam direction being referenced by a depends_on

### DIFF
--- a/base_classes/NXbeam.nxdl.xml
+++ b/base_classes/NXbeam.nxdl.xml
@@ -370,16 +370,12 @@
             Direction (and location) for the beam. The location of the beam can be given by
             any point which it passes through as its offset attribute.
         </doc>
-        <field name="DIRECTION" nameType="any" units="NX_TRANSFORMATION" type="NX_NUMBER">
+        <field name="BEAMdirection" nameType="partial" units="NX_UNITLESS" type="NX_NUMBER">
             <doc>
                 Direction of beam vector, its value is ignored. If missing, then the beam direction is defined as [0,0,1]
-                and passes through the origin
+                and passes through the origin. Note, this field should be referenced by the parent group's ``depends_on`` field; also,
+                as this field is a direction, its ``transformation_type`` attribute should be omitted.
             </doc>
-            <attribute name="transformation_type">
-                <enumeration>
-                    <item value="translation" />
-                </enumeration>
-            </attribute>
             <attribute name="vector" type="NX_NUMBER">
                 <doc>
                     Three values that define the direction of beam vector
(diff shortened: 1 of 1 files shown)


base_classes/NXbeam.nxdl.xml
```diff
@@ -373,13 +373,8 @@
         <field name="DIRECTION" nameType="any" units="NX_TRANSFORMATION" type="NX_NUMBER">
             <doc>
                 Direction of beam vector, its value is ignored. If missing, then the beam direction is defined as [0,0,1]
-                and passes through the origin
+                and passes through the origin. Note, this field should be referenced by the parent group's ``depends_on`` field
             </doc>
-            <attribute name="transformation_type">
-                <enumeration>
-                    <item value="translation" />
```
woutdenolf, line 380: The base class says it can be "translation" or "rotation". So removing this would allow "DIRECTION" to be a rotation...

PeterC-DLS: I'm using the field as a reference: [`NXtransformations` lines 130-134](https://github.com/nexusformat/definitions/blob/2aede967caebc84f3335d50f1344f9b3ed2e8603/base_classes/NXtransformations.nxdl.xml#L130-L134)

woutdenolf: Ah ok, makes sense, it is a direction (neither translation nor rotation). Maybe mention that the attribute "transformation_type" should not be provided?

base_classes/NXbeam.nxdl.xml
```diff
@@ -373,13 +373,8 @@
         <field name="DIRECTION" nameType="any" units="NX_TRANSFORMATION" type="NX_NUMBER">
```
lukaspie, line 373: I have a question regarding the use of `NXtransformations` in `NXbeam`. Since this is an all uppercase field with ´nameType="any"`, wouldn't _all_ fields within  `NXbeam/NXtransformations` comply with this? In other words, can I even have fields that have `units="NX_TRANSFORMATION" type="NX_NUMBER"` and are not the direction of the beam vector?

I am asking because of the rotations in `NXxps` (see e,g, [here](https://github.com/FAIRmat-NFDI/nexus_definitions/blob/fairmat-2024-mpes/applications/NXxps.nxdl.xml#L186)), which have `transformation_type=rotation`. These are rotations of the beam in another coordinate system, would that not clash with the field "DIRECTION" here?

base_classes/NXbeam.nxdl.xml
PeterC-DLS: Done

woutdenolf (approved):

base_classes/NXbeam.nxdl.xml
PeterC-DLS: This field is not intended to restrict other fields in this group but provide an single instance for the parent group's `depends_on` field to reference.

PeterC-DLS: I think a validator should check the parent's `depends_on` and follow it to this beam direction field and continue if down any transformations chain. If the beam direction field has a `transformation_type` attribute then log a warning if you are not being strict or throw an error otherwise.

Ah, I guess it cannot at that level. We need some `nameType` metadata/annotation to indicate we expect a single arbitrarily named field that matches to a `depends_on` value

lukaspie: If I come across a field `my_field` within an `NXbeam/NXtransformations`, wouldn't I have to assume that this is specializing `DIRECTION` field (i.e., in ontology language, `my_field` has an `is_a` relationship to `DIRECTION`)? So if it has a `transformation_type`, I would always need to throw an erorr because that is not to be given for `DIRECTION`.

I think this is a fundamental question as to how inheritance for fields work. Does inheritance in NeXus only apply between concepts with exact name match (e.g., if I have a field `NXbeam/my_field` in the base class and a field `NXmpes/NXbeam/my_field`  in the app. def., then the latter obviously inherits and specializes from the former). Or does the same name pattern matching that is used for data names also apply on the data modeling level (e.g., if I have a field `NXbeam/SOMEFIELD` with nameType="any" in the base class and a field `NXmpes/NXbeam/my_field`, does the latter inherit and specialize the former)?

Recent practice (e.g. `NXobject/FIELDNAME_xxx`, `NXobject/identifierNAME`) would suggest that it's the latter way of inheritance, by assuming that `FIELDNAME_xxx` will be inherited to any subsequently defined fields.

PeterC-DLS: In the case, we are not extending the `NXtransformations` formally - just an instance that occurs in `NXbeam`. Perhaps, as a workaround, I could use `name="BEAMdirection" nameType="partial"`.

lukaspie: That would have been my suggestion to avoid these ambiguities 👍

PeterC-DLS: Done

PR #1563, opened 2025-05-09
Clarify the conventions for rotations in left and right handed coordinate system

lukaspie:
There was a long-ish [discussion](https://github.com/nexusformat/definitions/pull/1415#discussion_r2075856637) in #1415 about which rotation convention to apply for left- and right-handed coordinate systems. Initially, it was agreed that if the `NXtransformation` depends on a coordinate system (i.e., its ``depends_on`` attribute points to an instance `NXcoordinate_system`), the rotation is **always** right handed in that coordinate system, even if the the coordinate system itself is left-handed (as defined by the determinant of its base vectors.

@PeterC-DLS suggested that this should not be the case. Rather, the rotation conventions **should** follow the handedness of the coordinate system because then the matrices don't have to be changed. So:

- Rotations in left-handed coordinate systems are left-handed (i.e., they follow the left-hand rule). In a left-handed coordinate system, positive rotation about an axis is clockwise when looking from the end of the positive axis in the direction of the origin (from infinity towards the origin).

- Rotations in right-handed coordinate systems are right-handed (i.e., the follow the right-hand rule). In a right-handed coordinate system, positive rotation about an axis is counter-clockwise when looking from the end of the positive axis in the direction of the origin (from infinity towards the origin).

Most importantly, the matrix transformation does not depend on handness and the same matrices can be used for left- and right-handed coordinate system (same Rodrigues' formula). Only the **interpretation** (“positive” direction of rotation) is different because the coordinate systems are mirror images! This will it also easier for most existing tools around ``NXtransformations``  to work with left-handed coordinate systems.

This is also the convention that most graphic packages and math/physics textbooks follow, from what I could gather.

I am not sure if a vote is needed for these documentation changes, but they should definitely be discussed.

### DIFF
--- a/base_classes/NXtransformations.nxdl.xml
+++ b/base_classes/NXtransformations.nxdl.xml
@@ -166,13 +166,13 @@
 				
 				* Rotations in left-handed coordinate systems are left-handed (i.e., they follow the 
 				  left-hand grip rule). In a left-handed coordinate system, positive rotation about an
-				  axis is counter-clockwise when looking from the origin in the direction of the 
-				  positive axis (from the origin towards infinity).
+				  axis is clockwise when looking from the from the end of the positive axis towards
+				  its origin (from infinity towards the origin).
 				
 				* Rotations in right-handed coordinate systems are right-handed (i.e., they follow the 
 				  right-hand grip rule). In a right-handed coordinate system, positive rotation about an
-				  axis is clockwise when looking from the origin in the direction of the 
-				  positive axis (from the origin towards infinity).
+				  axis is counter-clockwise when looking from the from the end of the positive axis towards
+				  its origin (from infinity towards the origin).
 
 				Note that by using this convention, the transformation matrices in both left- and
 				right-handed coordinate systems are the same.
--- a/base_classes/NXtransformations.nxdl.xml
+++ b/base_classes/NXtransformations.nxdl.xml
@@ -156,9 +156,11 @@
 				increasing displacement. For general axes, an appropriate direction
 				should be chosen.
 
-				Note that if the ``NXtransformation`` depends on a coordinate system
-				(i.e., its ``depends_on`` attribute points to an instance :ref:`NXcoordinate_system`),
-				for ``rotation`` axes, the rotation convention is the same as the handedness
+				By default, for ``rotation`` axes that do not explicitly depend on a coordinate system,
+				the direction should be chosen for a right-handed rotation with increasing angle.
+				If the ``NXtransformation`` depends on a coordinate system (i.e., its ``depends_on`` 
+				attribute (or a ``depends_on`` further up the transformation chain) points to an instance
+				:ref:`NXcoordinate_system`), the rotation convention is the same as the handedness
 				of the coordinate system (as defined by the determinant of its base vectors):
 				
 				* Rotations in left-handed coordinate systems are left-handed (i.e., they follow the 
@@ -173,9 +175,6 @@
 
 				Note that by using this convention, the transformation matrices in both left- and
 				right-handed coordinate systems are the same.
-
-				By default, ``rotation`` axes that do not explicitly depend on a coordinate system,
-				the direction should be chosen for a right-handed rotation with increasing angle.
 			</doc>
 			<dimensions rank="1">
 				<dim index="1" value="3" />
--- a/base_classes/NXtransformations.nxdl.xml
+++ b/base_classes/NXtransformations.nxdl.xml
@@ -163,12 +163,12 @@
 				
 				* Rotations in left-handed coordinate systems are left-handed (i.e., they follow the 
 				  left-hand grip rule). In a left-handed coordinate system, positive rotation about an
-				  axis is clockwise when looking from the origin in the direction of the 
+				  axis is counter-clockwise when looking from the origin in the direction of the 
 				  positive axis (from the origin towards infinity).
 				
 				* Rotations in right-handed coordinate systems are right-handed (i.e., they follow the 
 				  right-hand grip rule). In a right-handed coordinate system, positive rotation about an
-				  axis is counter-clockwise when looking from the origin in the direction of the 
+				  axis is clockwise when looking from the origin in the direction of the 
 				  positive axis (from the origin towards infinity).
 
 				Note that by using this convention, the transformation matrices in both left- and
--- a/base_classes/NXtransformations.nxdl.xml
+++ b/base_classes/NXtransformations.nxdl.xml
@@ -157,7 +157,8 @@
 				should be chosen.
 
 				By default, for ``rotation`` axes that do not explicitly depend on a coordinate system,
-				the direction should be chosen for a right-handed rotation with increasing angle.
+				the direction should be chosen for a right-handed rotation with increasing angle. Note,
+				McStas is a right handed coordinate system.
 
 				If the ``NXtransformation`` depends on a coordinate system (i.e., its ``depends_on`` 
 				attribute (or a ``depends_on`` further up the transformation chain) points to an instance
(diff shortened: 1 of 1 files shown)


PeterC-DLS:
I think your comment has the clockwise/anticlockwise mixed up for when viewing from the origin: RH CS has positive as clockwise and LH has that as anticlockwise.

base_classes/NXtransformations.nxdl.xml
```diff
@@ -151,16 +151,31 @@
...
 				Note that if the ``NXtransformation`` depends on a coordinate system
 				(i.e., its ``depends_on`` attribute points to an instance :ref:`NXcoordinate_system`),
-				the rotation is right handed in that coordinate system, even if the the coordinate system
-				itself is left-handed (as defined by the determinant of its base vectors).
+				for ``rotation`` axes, the rotation convention is the same as the handedness
+				of the coordinate system (as defined by the determinant of its base vectors):
+				
+				* Rotations in left-handed coordinate systems are left-handed (i.e., they follow the 
+				  left-hand rule). In a left-handed coordinate system, positive rotation about an
+				  axis is clockwise when looking from the origin in the direction of the
```
PeterC-DLS, line 166:
```suggestion
				  axis is anticlockwise when looking from the origin in the direction of the 
```

base_classes/NXtransformations.nxdl.xml
```diff
@@ -151,16 +151,31 @@
...
+				of the coordinate system (as defined by the determinant of its base vectors):
+				
+				* Rotations in left-handed coordinate systems are left-handed (i.e., they follow the 
+				  left-hand rule). In a left-handed coordinate system, positive rotation about an
+				  axis is clockwise when looking from the origin in the direction of the 
+				  positive axis (from the origin towards infinity).
+				
+				* Rotations in right-handed coordinate systems are right-handed (i.e., they follow the 
+				  right-hand rule). In a right-handed coordinate system, positive rotation about an
+				  axis is counter-clockwise when looking from the origin in the direction of the
```
PeterC-DLS, line 171:
```suggestion
				  axis is clockwise when looking from the origin in the direction of the 
```

base_classes/NXtransformations.nxdl.xml
```diff
@@ -151,16 +151,31 @@
...
 
 				Note that if the ``NXtransformation`` depends on a coordinate system
 				(i.e., its ``depends_on`` attribute points to an instance :ref:`NXcoordinate_system`),
-				the rotation is right handed in that coordinate system, even if the the coordinate system
-				itself is left-handed (as defined by the determinant of its base vectors).
+				for ``rotation`` axes, the rotation convention is the same as the handedness
+				of the coordinate system (as defined by the determinant of its base vectors):
+				
+				* Rotations in left-handed coordinate systems are left-handed (i.e., they follow the 
+				  left-hand rule). In a left-handed coordinate system, positive rotation about an
```
PeterC-DLS, line 165:
```suggestion
				  left-hand grip rule). In a left-handed coordinate system, positive rotation about an
```

base_classes/NXtransformations.nxdl.xml
```diff
@@ -151,16 +151,31 @@
...
+				  axis is clockwise when looking from the origin in the direction of the 
+				  positive axis (from the origin towards infinity).
+				
+				* Rotations in right-handed coordinate systems are right-handed (i.e., they follow the 
+				  right-hand rule). In a right-handed coordinate system, positive rotation about an
```
PeterC-DLS, line 170:
```suggestion
				  right-hand grip rule). In a right-handed coordinate system, positive rotation about an
```

lukaspie:
> I think your comment has the clockwise/anticlockwise mixed up for when viewing from the origin: RH CS has positive as clockwise and LH has that as anticlockwise.

Are you sure? https://en.wikipedia.org/wiki/Right-hand_rule#Coordinates says that for right-handed systems, positive == counter-clockwise when looking along an axis in the positive direction

lukaspie:
@PeterC-DLS thanks for the suggestions, I thinkt this is much cleaner now.

I also double-checked the additional documentation around transformations (specifically [here](https://manual.nexusformat.org/design.html#coordinate-transformations)), I don't think that there is anything there that is in contrast to the rotation conventions we define here.

base_classes/NXtransformations.nxdl.xml
```diff
@@ -151,16 +151,31 @@
...
 				For ``translation`` axes the direction should be chosen for
 				increasing displacement. For general axes, an appropriate direction
 				should be chosen.
 
-				Note that if the ``NXtransformation`` depends on a coordinate system
-				(i.e., its ``depends_on`` attribute points to an instance :ref:`NXcoordinate_system`),
-				the rotation is right handed in that coordinate system, even if the the coordinate system
-				itself is left-handed (as defined by the determinant of its base vectors).
+				By default, for ``rotation`` axes that do not explicitly depend on a coordinate system,
+				the direction should be chosen for a right-handed rotation with increasing angle.
```
PeterC-DLS, line 160: You could add that "(as McStas is right handed)"

base_classes/NXtransformations.nxdl.xml
```diff
@@ -151,16 +151,31 @@
...
+
+				If the ``NXtransformation`` depends on a coordinate system (i.e., its ``depends_on`` 
+				attribute (or a ``depends_on`` further up the transformation chain) points to an instance
+				:ref:`NXcoordinate_system`), the rotation convention is the same as the handedness
+				of the coordinate system (as defined by the determinant of its base vectors):
+				
+				* Rotations in left-handed coordinate systems are left-handed (i.e., they follow the 
+				  left-hand grip rule). In a left-handed coordinate system, positive rotation about an
+				  axis is clockwise when looking from the from the end of the positive axis towards
```
PeterC-DLS, line 169:
```suggestion
				  axis is clockwise when looking from a point on the positive axis towards
```
also below for the right hand case

lukaspie: Changed

base_classes/NXtransformations.nxdl.xml
lukaspie: But if the ``rotation`` axes that does not explicitly depend on a coordinate system (so also not on McStas), we want it to be right-handed anyway, right? This is why I didn't include any reference to McStas here intentionally.

PeterC-DLS: I mean it as a note that McStas is right handed. Perhaps change to "explicitly depend on a given coordinate system ... Note, McStas is a right handed coordinate system."

base_classes/NXtransformations.nxdl.xml
```diff
@@ -151,16 +151,31 @@
...
-				Note that if the ``NXtransformation`` depends on a coordinate system
-				(i.e., its ``depends_on`` attribute points to an instance :ref:`NXcoordinate_system`),
-				the rotation is right handed in that coordinate system, even if the the coordinate system
-				itself is left-handed (as defined by the determinant of its base vectors).
+				By default, for ``rotation`` axes that do not explicitly depend on a coordinate system,
+				the direction should be chosen for a right-handed rotation with increasing angle.
+
+				If the ``NXtransformation`` depends on a coordinate system (i.e., its ``depends_on`` 
+				attribute (or a ``depends_on`` further up the transformation chain) points to an instance
+				:ref:`NXcoordinate_system`), the rotation convention is the same as the handedness
```
PeterC-DLS, line 164:
```suggestion
				of :ref:`NXcoordinate_system`), the rotation convention is the same as the handedness
```

base_classes/NXtransformations.nxdl.xml
lukaspie: Ok, sounds good, I added that.

PR #1566, opened 2025-06-04
[MNT] Squashed commit of `fairmat-2024-mpes` branch

lukaspie:
!!! This PR changes the git history !!!

With this PR, we clean up the unfortunate situation that after merging #1424, a large number of commits has been added to the repository history, whereas a squash commit would have been the better solution. This new branch removes these large number of commits and replaces it with a clean, squashed commit. It also gives proper credit to all contributors by having them automatically added to the commit message.

This branch was created using a script, see first comment. What it does:
- Resets local main branch to exactly match the old main branch (we have this backupp-ed in https://github.com/FAIRmat-NFDI/nexus_definitions).
- Creates a new branch where we prepare the fix.
- Finds the point where the old main and the current state diverged — this helps identify the commits that were merged incorrectly. This is this commit: https://github.com/FAIRmat-NFDI/nexus_definitions/commit/fe1709e4a194316b325545f535d012270cfce2aa (the merge of #1415).
- Collects all the people who contributed to those commits, so they get credited properly later.
- Combines all those many commits into a single commit ("squash") on the new branch. This creates one clean commit with all the changes.
- Adds a list of all contributors to the commit message so GitHub shows them as co-authors.

Note that you can check for yourself that no files in `main` are changed by checking out this branch locally and running `git diff main` (or on a fork `git diff upstream/main`).


PeterC-DLS (approved):
LGTM

PR #1567, opened 2025-06-05
[Bug fix] Units in `NXxps/NXbeam/beam_direction`

lukaspie:
This fixes the units in `NXxps/NXbeam/beam_direction` such that they match with the [recently changed](https://github.com/nexusformat/definitions/pull/1562) units `NXbeam/BEAMdirection`.

@PeterC-DLS I hadn't realized you had changed the units in `NXbeam` as well. This should be an easy bug fix not requiring a vote.


PeterC-DLS (approved):
LGTM

PR #1568, opened 2025-06-11
use consistent XML header

lukaspie:
As discussed in #1424, our automatic conversion tool [`nyaml`](https://github.com/FAIRmat-NFDI/nyaml) used to write XML headers with single quote (`<?xml version='1.0' encoding='UTF-8'?>`), but the standard in this repo is double quotes (`<?xml version="1.0" encoding="UTF-8"?>`). 

We have fixed this in the tool since, but there is as set of already standardized base classes from the FAIRmat proposal, where it was missed to fix the header before merging. This PR fixes these headers. 

Can be checked by running `git grep -l "xml version='1.0'" base_classes/* applications/* contributed_definitions/*` (returns nothing after this fix).


PeterC-DLS (approved):
LGTM - great to be consistent!

Issue #1569, opened 2025-06-26
Self-description of non-NeXus terms (SECoP)


g-guenther:
As a result of the [SECoP](https://sampleenvironment.org/committees/secop/) mapping to NeXus the need for a standardized annotation for the meaning of a (data) field arises, e. g. using a term of an ontology. This belongs to fields that are part of the base class definition (e. g. [NXsensor/measurement](https://manual.nexusformat.org/classes/base_classes/NXsensor.html#nxsensor-measurement-field)) as well as to those not defined by base classes. It could be useful in general to make use of the sophisticated knowledge representations that exist. To improve machine-readability of a non-NeXus field, the following information could be provided :

- data_type attribute: defines the NXDL data type of the field
- identifier attribute: A PID of a resource defining the meaning of the field
- units_identifier attribute: A PID of a resource defining the units of the field (could be provided in addition to the units attribute)

As it is somehow related, the example of #1335 is used:
```
entry:NXentry
    ...
    instrument:NXinstrument
        ...
        neutron_guide:NXguide
            ...
            pressure_valve_1:NX_FLOAT = 0.1
                @data_type="NX_FLOAT"
                @identifier="https://qudt.org/vocab/quantitykind/Pressure"
                @units="mbar"
                @units_identifier="http://qudt.org/vocab/unit/MilliBAR"
```

I noticed similar demands in other projects and I think it could be of general interest. Would that make sense?

g-guenther:
For my understanding and to elaborate this topic, I present some examples which could help to find a proper implementation.

## Draft 1: Current NXobject definition
The current definition of NXobject which allows to annotate NeXus base classes (HDF groups).

```
entry:NXentry
    ...
    sample:NXsample
        ...
        identifier_sample="http://nmrML.org/nmrCV#NMR:1000457"
        humidity=0.3
                @units="%RH"
        humidity_env:NXenvironment
            ...
            identifier_humidity_env="http://purl.obolibrary.org/obo/ENVO_01000254"
                @type="ARK"
            measurement="humidity"
                @custom=True
            humidity_sensor_1:NXsensor
                ...
                identifier_humidity_sensor_1="http://snomed.info/id/468916005"
                    @type="ARK"
                value=0.3
                    @units="%RH"
```

## Draft 2: Current NXobject definition extended
Same example as Draft 1 but the current definition of NXobject is extended to fields and attributes.
This can't be realized in a consistent way since there are no attributes for attributes. In this example,
[...]

Issue #1398, opened 2024-07-10
Define a strategy for storing the experimental technique as a PaNET term


paulmillar:
One of the outcomes of the ExPaNDS project was developing a catalogue of experimental techniques being used at photon and neutron facilities.  Each technique is given a unique and persistent identified (e.g., anomalous small angle x-ray scattering has the ID `http://purl.org/pan-science/PaNET/PaNET01274`)

In addition, the techniques are structured to indicate when one technique is a more specific version of the other technique (e.g., microtomography is more specific variant of tomography).

You can find the complete list of techniques [here](https://expands-eu.github.io/ExPaNDS-experimental-techniques-ontology/index-en.html).

It would be helpful if NeXus allowed a file to specify with which technique the data was obtained, by supporting the inclusion of a PaNET term.

There is, perhaps, some overlap between the PaNET experimental technique and the NeXus application definitions.  The exact nation of this overlap is currently unclear and would (likely) not be as specific as we would want.

Therefore, one possible solution would be to enhance one of the base classes to include support for PaNET.

Issue #1335, opened 2023-12-06
Clarify use of non-NeXus terms


g-guenther:
What about terms that are not covered by NeXus, e.g. some motor parameter connected to a detector or a pressure value related to a neutron guide. I guess, I may add as many terms as I like anywhere in the NeXus file (if they do not use predefined NeXus terms), even outside dedicated groups such as NXcollection or NXparameters. However, I can't find a corresponding comment in the documentation. Sentences like "The base classes represent a set of components that define the dictionary of all possible terms to be used with that component" ([ref](https://manual.nexusformat.org/introduction.html)) suggest the opposite.

To be clear: Would NeXus allow to add not predefined terms to existing base classes? For example: 
```
entry:NXentry
    ...
    instrument:NXinstrument
        ...
        neutron_guide:NXguide
            ...
            pressure_valve_1:NX_FLOAT = 0.1
                @units="mbar"
```
If so, I would expect a promient comment e.g. in the [NeXus Design](https://manual.nexusformat.org/design.html) section. Have I overlooked a corresponding comment?

prjemian:
Yes, you can add fields of your choosing. Do not use existing fields (as
defined in that group) for different purposes. I agree this should be
stated clearly.  This could be a FAQ. There are some naming rules to
follow. To ensure longevity, make the names such that they are not likely
to be added to the standard in the future, unless that is your intention.
Some instruments apply a prefix (such as "usaxs_") to avoid any future
changes in the standard.

prjemian:
Another place to mention about adding non-defined fields is on the [NeXus Class Definitions](https://manual.nexusformat.org/classes/index.html) page.

PR #1570, opened 2025-06-26
Adding FIELDNAME field to NXobject

g-guenther:
Adds the field 'FIELDNAME' with attributes for data_type, identifier, and units_identifier to improve machine-readability of NeXus terms and non-NeXus terms, solving issue #1569. Might also be related to #1335 and #1398.

### DIFF
--- a/base_classes/NXobject.nxdl.xml
+++ b/base_classes/NXobject.nxdl.xml
@@ -31,7 +31,7 @@
         within this base class are allowed to be present in any derived base class.
 
         The FIELDNAME refers to a defined field of the derived base class or any
-        element that is not defined by the base class.
+        field that is not defined by the base class.
 
         If nameType="partial", the placeholders (e.g., FIELDNAME or GROUPNAME)
         can be replaced by the name of any object (field or group, 
--- a/base_classes/NXobject.nxdl.xml
+++ b/base_classes/NXobject.nxdl.xml
@@ -27,23 +27,60 @@
 	name="NXobject" 
 	type="group">
 	<doc>
-        This is the base object of NeXus. The groups and fields contained 
-        within this file are allowed to be present in any derived base class.
+        This is the base object of NeXus. The groups and fields contained
+        within this base class are allowed to be present in any derived base class.
+
+        The FIELDNAME refers to a defined field of the derived base class or any
+        element that is not defined by the base class.
 
         If nameType="partial", the placeholders (e.g., FIELDNAME or GROUPNAME)
         can be replaced by the name of any object (field or group, 
-        respectively) that exists within the same group.
-	</doc>
+        respectively).
+    </doc>
     <group type="NXcollection" minOccurs="0" />
     <group type="NXdata" minOccurs="0" />
     <group type="NXlog" minOccurs="0" />
     <group type="NXnote" minOccurs="0" />
     <group type="NXparameters" minOccurs="0" />
+
     <group name="GROUPNAME_log" type="NXlog" nameType="partial" minOccurs="0">
         <doc>
             NXlog group containing logged values of GROUPNAME.
         </doc>
     </group>
+    <field name="FIELDNAME" minOccurs="0">
+        <doc>
+            A field that is defined by the derived base class or added without
+            existing definition. Its type can be defined by the data_type attribute
+            and a meaning could be provided by the identifier attribute.
+        </doc>
+        <attribute name="data_type" type="NX_CHAR">
+            <doc>
+            Declares the NXDL data type of the field if the field (and its
+            type) is not defined by the derived class.
+           </doc>
+        </attribute>
+        <attribute name="identifier" type="NX_CHAR">
+            <doc>
+            An identifier for a (persistent) resource.
+
+            An identifier, provided by some authority, that has been assigned to an
+            object describing this field.
+
+            See the type attribute of the identifierNAME field for examples of identifiers.
+        </doc>
+        </attribute>
+        <attribute name="units_identifier" type="NX_CHAR">
+            <doc>
+            An identifier for a (persistent) resource.
+
+            An identifier, provided by some authority, that has been assigned to an
+            object describing the units of the field.
+
+            See the type attribute of the identifierNAME field for examples of identifiers.
+        </doc>
+        </attribute>
+    </field>
     <field name="FIELDNAME_set" type="NX_NUMBER" nameType="partial" minOccurs="0">
         <doc>
             Target values of FIELDNAME. 


mkuehbach (changes requested):
This PR, if merged, will have far reaching consequences for the NeXus standard. A few initial thoughts:

The reason why currently ``NXobject`` does neither contain a nx:fieldType ``FIELDNAME`` nor a nx:groupType ``GROUPNAME`` is because it so far assumed that as ``NXobject`` is formulated in NXDL i.e. uses the grammar defined in nxdl.xsd you can add always nx:fieldType, nx:groupType, nx:attributeType instances to any base class. If that were not possible how at all would all field and groups and attributes in NeXus connect, see also that before the Autumn Code Camp 2024 edits, ``NXobject`` had any content.

The reason why instances using the semantically clarified nameType="partial" were added, like concepts following the pattern ``UPPERCASE_lowercase`` (e.g. ``FIELDNAME_set``) was to reserve these concept names.

If ``FIELDNAME`` were to be added like here proposed every instance in every base class will in almost all cases be a specialization because since the Autumn Code Camp 2024 there is inheritance for base classes. All base classes inherit from ``NXobject`` means all concepts that are nx:fieldType instances will thus be connected to the here proposed ``FIELDNAME``, so this PR is at least incomplete in the sense that one needs to carefully think about nx:groupType ``GROUPNAME`` as well as nx:attributeType ``ATTRIBUTENAME``. 

The idea of adding nx:fieldType ``FIELDNAME`` like here proposed is sound.

``FIELDNAME`` has no "type" XML parameter. This is substantiated but again the documentation needs to be carefully investigated and discussed surplus clarification is required if "type" and other XML parameter like "dimensions" serve to impose a restriction on the allowed type and dimensions, like nx:fieldType ``value`` with type="NX_NUMBER" can be a scalar or tensor but if "dimensions" is added the shape is restricted and eventually a scalar no longer allowed. Are we 100% sure that afore-mentioned cases of an absence of the XML parameter "type" there is no default type = NX_CHAR kicking in, which would cause that each field in every base class that is not of type NX_CHAR will also automatically specialize.

When almost everything is specialized I think this reads weird, as special cases should not be the majority IMHO.

The intention of adding the identifier in this PR here are noble though demand edits in IMHO though.

Indeed, ``identifierNAME`` exists but strictly offers an identifier only for the parent group; not for fields, groups under ``NXobject`` will automatically inherit again via a chain from NXobject so also for these there is again an identifierNAME, nx:attributeType ``ATTRIBUTENAME`` though will not.

What indeed will have beneficial and far reaching consequences also is - when this PR is edited such that an nx:attributeType e.g. ``identifierNAME`` is added to each instance of ``FIELDNAME``. This could then enable that each and every concept in NeXus gets equipped with the here envisaged possibility to assign a GUPRI (globally unique persistent resource identifier).

There is a catch though: At least in the HDF5 data model there are no attributes of an attribute. NeXus is not HDF5 alone though, so for other serialization formats such sub-graph of attributes could again be possible but not for the currently most frequently used serialization container format HDF5. Anyway, exactly because of that attribute of an attribute not existent, identifierNAME was designed as a field with attributes that detail the types of attributes. Also the concept was name ``identifierNAME`` to enable dropping ``NAME`` (see that nameType=partial) so each group can have again an identifier.
This was a good achievement useful for research data management, driven forward by FAIRmat, cuz now e.g. an instance of NXsample will get thereby an ``identifier`` for the same "for free".

What one could indeed think about is if one adds ``FIELDNAME``, ``GROUPNAME_thinkaboutseveralones``, e.g. ``FIELDNAME_gupri`` this would achieve what is desired - with far reaching consequences.
Each field can then be connected to a persistent identifier or one could similarly point to other semantic concepts.

Thinking, though, then about the qualifiers in SSSOM that tell how precise each equivalent, narrow, or broad the precise the concept resolved by gupri matches to ``FIELDNAME`` in the NeXus realm. Now all these qualifiers would blow up NXobject but again with for the benefit exactly like it was already exercised for ``FIELDNAME_set``, ``FIELDNAME_errors`` would allow to inject further connections with semantic artifacts that proliferate through as said above.

FYI: @lukaspie (here has completely by chance and magic timing already one such PR been proposed that raises exactly the technical questions that can be beneficial for NeXus users that seek to connect to concepts from "other ontologies"), @sanbrock

g-guenther:
@mkuehbach Many thanks for the feedback! My intention was indeed to establish a connection to other knowledge representations mainly for terms that are not defined in NeXus, as explained in #1569. Since I was reading the newly developed approach of identifiers in NXobject, I thought this would be an appropriate way.

Even if this PR is the wrong approach, I think, it would be very useful for the NeXus standard to address the questions that you raised: how to deal with identifiers with respect to all NeXus objects (groups, fields, attributes) in a consistent way. Of course, there is a limit of HDF5 due to the fact that there are no attributes of attributes. This I tried to solve with the __identifier_ suffix in @units_identifier, and, if required, could be a general attribute @ATTRIBUTENAME_identifier.

The FIELDNAME field could also help to define the units attribute with NXDL. So far, the units attribute is only defined in a [paragraph of the manual](https://manual.nexusformat.org/datarules.html#nexus-data-units).

> Are we 100% sure that afore-mentioned cases of an absence of the XML parameter "type" there is no default type = NX_CHAR kicking in, which would cause that each field in every base class that is not of type NX_CHAR will also automatically specialize.

Maybe you can leave the type explicitly undefined, e.g. an NX_ANYTYPE that is something similar to None in Python. However, predefined NeXus fields would still be specializations as their type will override the NX_ANYTYPE.

> What one could indeed think about is if one adds `{FIELDNAME || GROUPNAME}_thinkaboutseveralones`, e.g. `FIELDNAME_gupri` this would achieve what is desired - with far reaching consequences. Each field can then be connected to a persistent identifier or one could similarly point to other semantic concepts.

I have chosen the attribute to make the connection of identifier to the field it defines as close (and unambigiuous) as possible. However, a separate field would also be possible.

mkuehbach:
A few more thoughts here also relevant for this issue: https://github.com/nexusformat/definitions/issues/1573:

- We should distinguish identifiers for semantic concepts (T-box) from their instances (A-box) both profit from a clean handling of identifiers. An example --- NXsample is essentially a class used to model a concept for which there be instance data. Having an identifier for NXsample can be useful for stating that this concept has some relation to another semantic concept (aka the concept for a sample in another ontology). Applying for all instances implicitly. The value for `identifier` could then be the resolvable PID e.g. purl of that other ontology's concept. In contrast, the real physical object call it e.g., *MySampleA* can use `identifier` for resolving an id in a given namespace be this a sample registered in some global sample registry or some lab-local database.
- We observe identifier may specify a name and resolver service, typically via  the first part of an URI whereby the identifier is resolvable and the actual identifier unique ID part.
- The current choice with `identifierNAME` exemplifies there are cases where the first part of an URI is provided, like purl but there are also cases it isn't like ISSN-L or ISBN. Exactly this second case is why an attribute `type` is currently required for `identifierNAME` because if just the ID part is given it is tricky to know the namespace. (Is it some local thingi, or an ISSN-L.?)
- However, that demand for having two parts --- namespace (explicitly) and ID part --- make it tricky to extend identifiers to attributes. Specifically for the example of the HDF5 data model there attributes do not have childs.
- So compression of both parts into one value to avoid the need for a `type` attribute is problematic.

- One could use reserved suffixes to resolve the situation. An example for a fieldType `temperature_value`, `temperature_unit`, and then have attributeTypes for each, most importantly for `_unit` to state that this unit is e.g., equivalent to a specific unit mentioned in e.g., qudt ontology.
The reason why we this _value, _unit approach is poor though is that in HDF5 the best practice recommendation is that units be attributes of fields and we would typically double the number of fields if making current attributes to units. All fine but the downside is now that we cannot make `type` an attribute of `identifierNAME`.
Exactly this nesting would be good though to have so that also NeXus attributeTypes can have an identifierNAME.

- Now fieldTypes that are enums: `emitter_type` = "thermionic", I see value in that instead of a string one could also give a PID to a concept that is defined in another ontology specifying sth equivalent to a thermionic emitter.

- Let's thing about moving all essence of `identifierNAME` to attributes, spelling it out:
- For groupType:
```
sample/@NX_class = "NXsample"
sample/@PREFIXidentifier_value = "purl ..."  (maybe drop the _value?)
sample/@PREFIXidentifier_type = "purl ..."
```

- For fieldType:
```
sample/volume = "1."
sample/volume/@PREFIXidentifier_value = "https://w3id.org/emmo#EMMO_f1a51559_aa3d_43a0_9327_918039f0dfed"
```

Possibly, instead of the literal *identifier_value* one might even wish to qualify the extend of semantic mapping similarity,
e.g. `sample/volume/@is_equivalent_to = "https://w3id.org/emmo#EMMO_f1a51559_aa3d_43a0_9327_918039f0dfed"

- For attributeType:
```
sample/volume/@units = "m^3"
sample/volume/@units_identifier_value = "[qudt ...](https://qudt.org/vocab/unit/M)" 
# sample/@units_identifier_type = "qudt"  # optional because identifier_value is descriptive enough to be auto-resolvable given there is an URI with namespace part
```

Issue #1575, opened 2025-07-04
broken links to https://www.nexusformat.org/ pages in base class definitions


zdemat:
There is a link to [https://www.nexusformat.org/2014_How_to_find_default_data.html ](https://www.nexusformat.org/2014_How_to_find_default_data.html) in the [NXdata@signal](https://github.com/nexusformat/definitions/blob/c0942c95a2cb77452c0c0ce666d4100fe34dfd03/base_classes/NXdata.nxdl.xml#L227) description. But that link is 404 Not found. 

It seems the page is in "content" now: [https://www.nexusformat.org/content/2014_How_to_find_default_data.html ](https://www.nexusformat.org/content/2014_How_to_find_default_data.html)

There is obviously another broken link in the "Description", in the "Note" box.

zdemat:
```
base_classes/NXentry.nxdl.xml:88:			      https://www.nexusformat.org/NIAC2016.html,
base_classes/NXdata.nxdl.xml:61:			https://www.nexusformat.org/NIAC2018Minutes.html#nxdata-plottype--attribute
base_classes/NXdata.nxdl.xml:227:			See https://www.nexusformat.org/2014_How_to_find_default_data.html
base_classes/NXdata.nxdl.xml:242:			    https://www.nexusformat.org/NIAC2018Minutes.html
base_classes/NXroot.nxdl.xml:125:            See https://www.nexusformat.org/2014_How_to_find_default_data.html
base_classes/NXobject.nxdl.xml:243:			See https://www.nexusformat.org/2014_How_to_find_default_data.html
base_classes/NXresolution.nxdl.xml:106:             See https://www.nexusformat.org/2014_How_to_find_default_data.html
base_classes/NXcalibration.nxdl.xml:205:            See https://www.nexusformat.org/2014_How_to_find_default_data.html
base_classes/NXsensor.nxdl.xml:168:            See https://www.nexusformat.org/2014_How_to_find_default_data.html
```

PeterC-DLS:
Ah, good catch. There's more files in additional to the ones you list. @RussBerg and I may have to tweak the wiki generation to fix these dead links.

zdemat:
@PeterC-DLS, on the wiki it was this issue [wiki/issues/23](https://github.com/nexusformat/wiki/issues/23). But @prjemian suggested to report rather here.

I somehow do not like that we are showing the full link and with #1576 the link is longer.

PR #1576, opened 2025-07-07
fixing links to https://www.nexusformat.org/

zdemat:
fixes issue #1575


PeterC-DLS:
Might be better to fix the wiki...

zdemat:
I will not be offended if this MR is rejected. As I noted on the issue, this makes the links longer and that is not nice.

PR #1577, opened 2025-07-07
NXazint contribution

woutdenolf:
Two application definitions for storing the result of azimuthal unwrapping of diffraction patterns: NXazint1d and NXazint2d.

An example application that creates these files at MAX IV is documented here: https://maxiv-science.github.io/azint_writer/.

An example file can be inspected [here](https://myhdf5.hdfgroup.org/view?url=https%3A%2F%2Fzenodo.org%2Frecords%2F15744977%2Ffiles%2Fnx_azint1d_azint2d.h5%3Fdownload%3D1).

Organize this contribution: https://github.com/nexuscontributions/nxazint/issues/3

HTML rendering of [NXazint1d](https://hdf5.gitlab-pages.esrf.fr/nexus/nxazint/classes/contributed_definitions/NXazint1d.html) and [NXazint2d](https://hdf5.gitlab-pages.esrf.fr/nexus/nxazint/classes/contributed_definitions/NXazint2d.html)

### DIFF
--- a/applications/NXazint1d.nxdl.xml
+++ b/applications/NXazint1d.nxdl.xml
@@ -97,16 +97,8 @@
         <doc>
           is a normalization correction applied or not.
 
-          The normalization correction accounts for the effective number or weighted contribution of detector pixels to each integration bin.
-          
-          Note: An important aspect of the normalization strategy is how polarization and solid angle corrections are incorporated, which can vary
-          depending on the specific application, software, and its configuration options (see, for example, PyFAI documentation).
-          Additionally, the normalization strategy may include a relative or absolute calibration factor. Two common normalization approaches are:
-          "Relative normalization" to the PONI (Point Of Normal Incidence) pixel, and "Absolute calibration", which yields the number of photons
-          scattered by the sample in a given direction per unit solid angle. The type of the normalization strategy is not indicated on this level.
-          It must be concluded from the software used or its parameters.
-
-          The monitor correction is not included in the normalization correction and it is specified separately.
+          It indicates that integrated intensities and their errors were already divided by the appropriate normalization factors accounting
+          for the effective number or weighted contribution of detector pixels to each integration bin.
         </doc>
       </field>
 
@@ -279,6 +271,20 @@
       </field>
 
       <field name="norm" type="NX_NUMBER" optional="true">
+        <doc>
+            Values of the normalization correction.
+            
+            The normalization correction accounts for the effective number or weighted contribution of detector pixels to each integration bin.
+          
+            Note: An important aspect of the normalization strategy is how polarization and solid angle corrections are incorporated, which can vary
+            depending on the specific application, software, and its configuration options (see, for example, PyFAI documentation).
+            Additionally, the normalization strategy may include a relative or absolute calibration factor. Two common normalization approaches are:
+            "Relative normalization" to the PONI (Point Of Normal Incidence) pixel, and "Absolute calibration", which yields the number of photons
+            scattered by the sample in a given direction per unit solid angle. The type of the normalization strategy is not indicated on this level.
+            It must be concluded from the software used or its parameters.
+
+            The monitor correction is not included in the normalization correction and it is specified separately.
+        </doc>
         <dimensions rank="1">
           <dim index="1" value="nRad" />
         </dimensions>
--- a/applications/NXazint2d.nxdl.xml
+++ b/applications/NXazint2d.nxdl.xml
@@ -103,16 +103,8 @@
         <doc>
           is a normalization correction applied or not.
 
-          The normalization correction accounts for the effective number or weighted contribution of detector pixels to each integration bin.
-          
-          Note: An important aspect of the normalization strategy is how polarization and solid angle corrections are incorporated, which can vary
-          depending on the specific application, software, and its configuration options (see, for example, PyFAI documentation).
-          Additionally, the normalization strategy may include a relative or absolute calibration factor. Two common normalization approaches are:
-          "Relative normalization" to the PONI (Point Of Normal Incidence) pixel, and "Absolute calibration", which yields the number of photons
-          scattered by the sample in a given direction per unit solid angle. The type of the normalization strategy is not indicated on this level.
-          It must be concluded from the software used or its parameters.
-
-          The monitor correction is not included in the normalization correction and it is specified separately.
+          It indicates that integrated intensities and their errors were already divided by the appropriate normalization factors accounting
+          for the effective number or weighted contribution of detector pixels to each integration bin.
         </doc>
       </field>
 
@@ -285,6 +277,20 @@
         </field>
 
         <field name="norm" type="NX_NUMBER" optional="true">
+          <doc>
+            Values of the normalization correction.
+            
+            The normalization correction accounts for the effective number or weighted contribution of detector pixels to each integration bin.
+          
+            Note: An important aspect of the normalization strategy is how polarization and solid angle corrections are incorporated, which can vary
+            depending on the specific application, software, and its configuration options (see, for example, PyFAI documentation).
+            Additionally, the normalization strategy may include a relative or absolute calibration factor. Two common normalization approaches are:
+            "Relative normalization" to the PONI (Point Of Normal Incidence) pixel, and "Absolute calibration", which yields the number of photons
+            scattered by the sample in a given direction per unit solid angle. The type of the normalization strategy is not indicated on this level.
+            It must be concluded from the software used or its parameters.
+
+            The monitor correction is not included in the normalization correction and it is specified separately.
+          </doc>
           <dimensions rank="2">
             <dim index="1" value="nEta" />
             <dim index="2" value="nRad" />


contributed_definitions/NXazint1d.nxdl.xml
```diff
@@ -0,0 +1,275 @@
...
+      </doc>
+
+      <field name="definition">
+        <doc>Official NeXus NXDL schema to which this file conforms.</doc>
+        <enumeration>
+          <item value="NXazint1d"></item>
+        </enumeration>
+      </field>
+            
+      <field name="solid_angle_applied" type="NX_BOOLEAN">
```
PeterC-DLS, line 87: I think these boolean fields are better off in `NXparameters`

contributed_definitions/NXazint1d.nxdl.xml
```diff
@@ -0,0 +1,275 @@
...
+      <group type="NXparameters" name="input">
+        <doc>
+          Parameters should exactly match those required by the algorithm used in the processing.
+          For example, `azint` requires `error_model`, `mask`, `n_splitting`, `poni`, etc.
+        </doc>
+      </group>
+
+    </group>
+    
+    <group type="NXMonitor" name="MONITOR" optional="true">
```
PeterC-DLS, line 155:
```suggestion
    <group type="NXmonitor" name="MONITOR" optional="true">
```

contributed_definitions/NXazint2d.nxdl.xml
```diff
@@ -0,0 +1,317 @@
...
+        <doc>Number of azimuthal bins</doc>
+      </symbol>
+      <symbol name="nEtaEdge">
+        <doc>Number of azimuthal bin edges (nEta+1)</doc>
+      </symbol>
+    </symbols>
+
+    <doc>
+      This application definition describes data from two-dimensional area detectors that has been
+      integrated azimuthally, with a certain radial binning in units of q or 2theta and with
```
PeterC-DLS, line 56: The data is mapped, projected, binned or resampled to a polar coordinate system - not sure it's strictly an integration!

contributed_definitions/NXazint1d.nxdl.xml
```diff
@@ -0,0 +1,275 @@
...
+      <field name="solid_angle_applied" type="NX_BOOLEAN">
+        <doc>is solid angle correction applied or not.</doc>
+      </field>
+
+      <field name="polarization_applied" type="NX_BOOLEAN">
+        <doc>is polarization correction applied or not.</doc>
+      </field>
+
+      <field name="normalization_applied" type="NX_BOOLEAN">
+        <doc>is normalization correction applied or not.</doc>
```
PeterC-DLS, line 96: This should state what type of normalization is done

zdemat:
Hi @PeterC-DLS 

thanks for reviewing. I missed the worked done. I will go over your comments step by step.

contributed_definitions/NXazint1d.nxdl.xml
```diff
@@ -0,0 +1,275 @@
...
+
+      <group type="NXinstrument" name="INSTRUMENT">
+        
+        <field name="name" type="NX_CHAR">
+          <doc>Name of instrument (beamline) where data was collected.</doc>
+        </field>
+  
+        <group type="NXmonochromator" name="MONOCHROMATOR">
+          <field name="wavelength" type="NX_FLOAT">
+            <doc>Wavelength in angstrom.</doc>
```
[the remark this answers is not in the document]
zdemat: @fgjorup is there any strong opinion about this or should we just copy what is in NXmonochromator but making the _wavelength_ effectively obligatory.

contributed_definitions/NXazint1d.nxdl.xml
```diff
@@ -0,0 +1,275 @@
...
+          </field>
+          <field name="energy" type="NX_FLOAT">
+            <doc>Energy in keV.</doc>
+          </field>
+        </group>
+
+        <group type="NXsource" name="SOURCE">
+          <field name="name" type="NX_CHAR">
+            <doc>Name of laboratory where data was collected.</doc>
```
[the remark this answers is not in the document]
zdemat: @fgjorup it is a similar case. There is a fixed unit.

zdemat:
@sanbrock 

> An `NXsample` group would be handy too

It is `NXentry` so `NXsample` can always be there. However we do not want to make it obligatory. The idea here is that e.g. detector recording and Azimuthal Integration procedures are distributed tasks and they are writing different files that are later merged in a big application definition. We are not always capable to propagate reliable sample information to the distributed system. We intent to keep `NXazint` simple that it can be written with software missing the sample information. Similar reason is for the `monitor` being optional.

contributed_definitions/NXazint1d.nxdl.xml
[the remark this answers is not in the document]
fgjorup: I think it makes sense to specify the units attribute for both wavelength and energy, angstrom and keV, respectively.

phyy-nx:
Current thoughts from the Telco:

- One way this could be done is to instead make `NXazint1d` and `NXazint2d` base classes that extend from `NXprocess`.  That would be a different solution than an application definition, but then these common use cases could be reused elsewhere for other diffraction related use cases (eg SAXS/WAXS).  Thoughts?
- This seems to need more review and discussion.  Therefore we should move it to the next milestone.

phyy-nx:
Alternatively since this is a contributed definition, it could go into this release to serve as a kickoff for more conversation down the line. In that case, the above comments would all need to be resolved.

Happy either way :)

zdemat:
@rayosborn thanks for running the validator.

The file at Zendo is wrong. The required fields should not be in the parent /entry and the intention is not they are shared between /subentries. To my opinion the applications class as /subentry cannot dictate what should be in the parent /entry.

This was corrected during summer but the file at Zenodo remained wrong and not following the definition. @meghdadyazdi can you please fix the reference file at Zenodo so that it follows the definition?

zdemat:
(just polishing)

@rayosborn in relation to the NeXpy validator output. It says that `monitor_applied` is a required field. It should be optional in the definition. Do we have anything wrong there that the validator does not recognize that?

```
<field name="monitor_applied" type="NX_BOOLEAN" optional="true">
. . .
```

zdemat:
(formal)

- moved from contributed to applications as requested above ([358139e](https://github.com/nexusformat/definitions/pull/1577/commits/358139e6b652cff474382209316cd3493f5b7233))
- cosmetic changes in the description, mainly in order to fit a reasonably long sentence in the applications definition list. ([019cdf5](https://github.com/nexusformat/definitions/pull/1577/commits/019cdf5aab463ef3207749e543b3ded5a1d51e4e))

applications/NXazint1d.nxdl.xml
```diff
@@ -93,7 +94,31 @@
...
-        <doc>is normalization correction applied or not.</doc>
+        <doc>
+          is normalization correction applied or not.
+
+          The normalization correction accounts for the effective number or weighted contribution of detector pixels to each integration bin.
+          An important aspect of the normalization strategy is how polarization and solid angle corrections are incorporated, which can vary
+          depending on the specific application, software, and its configuration options (see, for example, PyFAI documentation).
+          Additionally, the normalization strategy may include a relative or absolute calibration factor. Two common normalization approaches are:
+          "Relative normalization" to the PONI (Point Of Normal Incidence) pixel, and "Absolute calibration", which yields the number of photons
+          scattered by the sample in a given direction per unit solid angle.
```
PeterC-DLS, line 105: This is confusing for the casual reader as this flag does not express which normalization has been applied.

applications/NXazint1d.nxdl.xml
```diff
@@ -93,7 +94,31 @@
...
+
+          Contrary the "monitor" correction is not expected to be included in the normalization correction.
```
PeterC-DLS, line 107: This statement is a note and should be expressed in a more definite manner

applications/NXazint1d.nxdl.xml
```diff
@@ -93,7 +94,31 @@
...
+
+          Practically, this indicates that "DATA:norm" has been already applied to "DATA:I".
```
PeterC-DLS, line 109: Don't introduce new undefined notation

applications/NXazint2d.nxdl.xml
```diff
@@ -52,10 +52,10 @@
     </symbols>
 
     <doc>
-      This application definition describes data from two-dimensional area detectors that has been
-      integrated azimuthally, with a certain radial binning in units of q or 2theta and with
-      a binning around the azimuthal angle eta.  An example application that creates these files at MAX IV 
-      is documented here: https://maxiv-science.github.io/azint_writer/   
+      Application definition for data from two-dimensional area detectors that has been integrated azimuthally,
```
PeterC-DLS, line 55: This copy and paste is wrong. There has not been any integration performed

zdemat:
(polishing)

Hi @rayosborn and @woutdenolf 

the file at Zenodo was updated but that means it got a new doi:[10.5281/zenodo.17286999](https://doi.org/10.5281/zenodo.17286999). The old doi redirects to the new doi at Zenodo but the link for data file access needs to be updated. So the link to [myhdf5.hdfgroup.org](https://myhdf5.hdfgroup.org) on top is pointing to the old version of the example file. Here is the [updated link to hdf5 preview](https://myhdf5.hdfgroup.org/view?url=https%3A%2F%2Fzenodo.org%2Frecords%2F17286999%2Ffiles%2Fnx_azint1d_azint2d.h5) however there are more files in Zenodo post now.

zdemat:
Hi @PeterC-DLS 

thanks for the second round of comments.

Concerning the comments about the documentation of the  `normalization_applied` and `monitor_applied` fields.
- I removed the notes referring to undefined notation.
- I made some statements more explicit.
- However I left the long note about normalization correction strategy. First it must be clear that it is mostly geometric but it may depend on the polarization too. Second that different software has different normalization and we cannot resolve it here.

Concerning the note that 2D-azint is not integration. I was commenting on this in the earlier conversations here, partially sharing your opinion but also giving counterarguments.
- scientists are calling that "integration"
- the procedure is very specific in the sense that the 1d case is clearly some averaging process over the pixels on the scattering ring. The averaging process also compensates for the varying polarization and solid angle. That is very similar for the 2D case. Especially for the higher angles, the value in each pixel of the output 2D-"integrated" data is corrected for these effects and it may be an average (integral) over several pixels of the original flat 2D detector image.

applications/NXazint2d.nxdl.xml
```diff
@@ -0,0 +1,339 @@
...
+        <doc>is polarization correction applied or not.</doc>
+      </field>
+
+      <field name="normalization_applied" type="NX_BOOLEAN">
+        <doc>
+          is a normalization correction applied or not.
+
+          The normalization correction accounts for the effective number or weighted contribution of detector pixels to each integration bin.
+          
+          Note: An important aspect of the normalization strategy is how polarization and solid angle corrections are incorporated, which can vary
```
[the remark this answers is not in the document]
PeterC-DLS: Same change for the other app def too

applications/NXazint1d.nxdl.xml
```diff
@@ -0,0 +1,298 @@
...
+      <field name="monitor_applied" type="NX_BOOLEAN" optional="true">
+        <doc>
+          is a monitor correction applied or not.
+
+          The monitor correction accounts for external factors that are independent of the azimuthal integration process.
+          Most commonly, this involves normalizing for fluctuations in the incident beam intensity or, where applicable, variations in exposure time.
+        </doc>
+      </field>
+
+      <group type="NXinstrument" name="INSTRUMENT">
```
PeterC-DLS:
```suggestion
      <group type="NXinstrument">
```

applications/NXazint1d.nxdl.xml
```diff
@@ -0,0 +1,298 @@
...
+        
+        <field name="name" type="NX_CHAR">
+          <doc>Name of instrument (beamline) where data was collected.</doc>
+        </field>
+  
+        <group type="NXmonochromator" name="MONOCHROMATOR">
```
PeterC-DLS:
```suggestion
        <group type="NXmonochromator">
```

applications/NXazint1d.nxdl.xml
```diff
@@ -0,0 +1,298 @@
...
+        </attribute>
+      </field>
+
+      <field name="norm" type="NX_NUMBER" optional="true">
+        <dimensions rank="1">
+          <dim index="1" value="nRad" />
+        </dimensions>
+        <attribute name="long_name" type="NX_CHAR">
+          <enumeration>
+            <item value="number of pixels contributing to the corresponding bin"></item>
```
PeterC-DLS: When correcting for solid angle, is this the "number" of pixels?

zdemat: Unfortunately this is not so simple. The descriptions is simplified. Note that it must be short as it will appear mainly as x-axis label in plots. A more precise definition is with the `normalization_applied` field: _the effective number or weighted contribution of detector pixels to each integration bin_. But that is too long.

zdemat: One can suggest different short description. However we should keep in mind people can use plots with such description for non-expert audience and we cannot explain all details how a particular sw is implementing all geometrical, mathematical and physical corrections that may be part of the normalization.

applications/NXazint2d.nxdl.xml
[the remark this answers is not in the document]
zdemat: I see where is the confusion:
- "polarization_applied" means that polarization correction was part of the calculation and is included (in all reduced/calculated data, in particular the integrated intensity "I", "I_errors" and "norm").
- "solid_angle_applied" means that solid angle correction was part of the calculation and is included.
- "normalization_applied" means that intensities ("I") and their errors ("I_errors") are actually already divided by the normalization data ("norm"). The normalization data includes "somehow" or not the polarization or solid angle corrections according to fields indicated above. They can also contain additional normalization (mostly geometric or physical) factors as e.g. area of the (virtual) normal incidence pixel.

I can try to fix this today by moving this description text rather to the "norm" field and making a statement I with previous commit, as there was an objection we are introducing new symbols. I can describe that by words instead: "normalization_applied" - *Indicates that integrated intensities and their errors were already divided by the appropriate normalization factors accounting for the effective number or weighted contribution of detector pixels to each integration bin.*

PeterC-DLS (approved):
LGTM

Issue #1432, opened 2024-09-26
can NXlog/value be used to record strings or boolean readouts ?


jkotan:
It is convenient to record string-like status of the hardware  or bool-like hardware flags. Ideal for those purpose would be NXlog/value. 
However, in the documentation NXlog/value is of NX_NUMBER type. It would be good to extend it by NX_BOOLEAN and NX_CHAR. Especially that in the documention it is written  that and NeXus field which changes during the experiment  can be replaced as an  NXlog group.

prjemian:
The intent of NXlog (as it is described) is to log numerical data.  Expanding to non-numerical types could have consequences for software which expects only numbers.

rerpha:
@prjemian is there an alternative method of storing time-series strings, ie. a different class?

prjemian:
NXnote is the class to use when nothing else fits.

Issue #687, opened 2019-09-16
Allow strings in NXlog?


matthew-d-jones:
The `value` dataset in `NXlog` is documented as type `NX_NUMBER`, but it would be useful to us to use it for strings too.
Any reason we should not change it to `NX_ANY`?
Interestingly the documentation for `NX_ANY` explicitly mentions logs as a use case...

zjttoefs:
NXnote mentions logs. I like the fact that quantifies are numbers + units that are universally understood and people cannot start "logging" temperatures of "300 K".

What is the real-world use case? If it is system logs, then we should make a separate class.

SkyToGround:
The NeXus standard is not necessarily very clear on how its various "constants" can be applied/used but it is my understanding that `NX_ANY` is meant to indicate possible units and not possible data types. Hence the solution for allowing strings in NXlog would then be something along the lines of replacing `value: (optional) NX_NUMBER {units=NX_ANY}`with `value: (optional) NX_NUMBER {units=NX_ANY}, NX_CHAR` on the [NXlog documentation page](http://download.nexusformat.org/doc/html/classes/base_classes/NXlog.html#nxlog).

zjttoefs:
I am against turning NXlog into something that cannot be interpreted (=arbitrary strings). That is why we have a standard. But maybe there is a use case I am not seeing. Hence the "awaiting input".

SkyToGround:
I have two use cases for allowing strings in NXlog:

1. Log messages
We have long discussed the fact that we want to store some clear text log messages to the NeXus files. One (possible) example would be to store some information from the accelerator control room, e.g. `Accelerator power will decrease from 2 MW to 1 MW between 18:00 and 19:00.`
The (weak) arguments against using `NXnote` are that the `date` field does not allow timestamps as anything other than strings and that the documentation does not necessarily (explicitly) allow for the fields (in this case `date`, `type`, `data`) to be arrays.

2. EPICS updates
The channel access and pvAccess EPICS protocols allows for the `VAL` field to be in the form of a string. Thus the `f142` flat buffer schema is also able to store its value as a string and the `f142` writer module can write those strings to the `value` field of a `NXlog`. By not allowing "value"-strings in NXlog, the string functionality will have to be removed from the `f142` writer (and possibly schema).
[...]

PR #1590, opened 2025-09-19
Add NXtextlog base class

FreddieAkeroyd:
Proposal for an `NXtextlog` base class, this is similar to `NXlog` but specifically for text based data as `NXlog` only supports numeric data. It could be used for example to record status/error/log messages from hardware or software and meant to handle cases where a large number of `NXnote` entries is undesirable. See #1432 and #1591

The implementation below currently has the option of a `separator` attribute to tell the reader how to split an indiidual value, but i'm not sure how useful that may or may not be. It may just be better to write multiple values in one log with the same timestamp, or as separate `NXtextlog` entries, for these cases.

Other questions might be:
- should there be a text `encoding` attribute or just mandate `UTF-8`
- is it useful to allow an optional end-of-line character to be specified, or just leave that interpretation for the reader


rayosborn:
Is it really necessary to create a new base class? Can't we just add a text array as an option to the NXlog class? We have just added dozens of new base classes, so perhaps I shouldn't complain about an extra one, but I think we need to be clear if it's really necessary.

FreddieAkeroyd:
Hi Ray, this PR is in relation to #1432 which I forgot to reference in the PR description (have added now). In a previous discussion on this in #687 the consensus seemed to favor not changing `NXlog` to allow non-numeric data, but as that was a few years ago it may be worth revisiting this discussion.

phyy-nx:
> > Datasets with variable length datatypes cannot be compressed.

Hi, thanks for the comment, discussed in Telco May 27 and it seems too technical to include here when it's an implementation detail

phyy-nx (approved):
Moved to contributed so doesn't need a vote.

PR #1593, opened 2025-09-24
datetime.UTC introduced in python 3.11

woutdenolf:
`datetime.UTC` has only been introduced in python 3.11.

It is an alias for `datetime.timezone.utc`.

https://docs.python.org/3/library/datetime.html#datetime.UTC

`datetime.timezone.utc` works for all modern python versions (3.2 -> 3.13+).


woutdenolf:
Ping @lukaspie

PR #1594, opened 2025-10-23
use `vary` and `initial_value` from `NXparameters` in `NXfit_function`

lukaspie:
This align `NXparameters` used in `NXfit_function` with the changes to the base class `NXparameters` applied in https://github.com/nexusformat/definitions/pull/1560.

I couldn’t make my PR point to @rayosborn's branch because it is on his own private repository, so I just pointed it to the main branch. Not sure if the whole thing needs a vote, as it is basically just clean-up.

### DIFF
--- a/base_classes/NXfit_function.nxdl.xml
+++ b/base_classes/NXfit_function.nxdl.xml
@@ -67,26 +67,6 @@ and a BNF of valid grammar.-->
                     A description of what this parameter represents.
                 </doc>
             </attribute>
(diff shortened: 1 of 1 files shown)


base_classes/NXfit_function.nxdl.xml
mkuehbach: If #1560 will be accepted, the attributes `initial`, `min`, `max`, `vary` do not need to be defined here in NXfit_function, they are inherited from NXparameters

PR #1599, opened 2025-11-18
Remove specification of one year curation requirement for contributions

phyy-nx:
As discussed in the Nov 2025 Telco, nowadays the one-year requirement for contributed definitions to incubate seems too long given the pace of development.  This PR drops the requirement, but notably, I couldn't find the explicit requirement anywhere, including our [constitution](https://www.nexusformat.org/NIAC.html#constitution), except [this `applying-nexus.rst` page](https://manual.nexusformat.org/applying-nexus.html#step-4-standardize-with-the-niac)!

### DIFF
--- a/manual/source/applying-nexus.rst
+++ b/manual/source/applying-nexus.rst
@@ -474,14 +474,16 @@ type, some more steps are required:
 
 - Correct your definition per the comments of the NIAC
 
-- Cure and use the definition for a year
+- Cure and use the definition to verify it is ready
 
 - After a final review, it becomes the standard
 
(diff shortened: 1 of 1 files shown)


benajamin:
I don't think you'll find it written anywhere, except maybe in the minutes of a very early NIAC meeting. I remember discussion of the policy's aim was for people to "eat their own dog food" for a while to make sure that the proposed definition did fulfill the requirements of the proposers. I think that now we have a much better understanding of the general requirements and a broad range of examples to compare to, such that the 1-year policy can be relaxed, or perhaps removed altogether.

benajamin (approved):

Issue #1602, opened 2025-12-09
Some application definition categorization needs amending


PeterC-DLS:
In [Application Definitions](https://manual.nexusformat.org/classes/applications/index.html), there are several research fields including [Diffraction Techniques](https://manual.nexusformat.org/classes/applications/diff-structure.html) and [(Small-Angle) Scattering Techniques](https://manual.nexusformat.org/classes/applications/sas-structure.html#). However, the former contains definitions for tomography.

Perhaps better categories would be "Diffraction and scattering" and "Imaging" with the latter containing the tomography definitions and allowing coherent imaging, ptychography, etc.

PeterC-DLS:
Closed by #1603

PR #1603, opened 2025-12-09
Recategorise some application definitions for #1602

PeterC-DLS:
Move tomography definitions to a new imaging category and move scattering definitions to diffraction


mkuehbach (changes requested):
Implemented changes I agree with,
linting issue in the CI,
run make style once again and should be fine for merging then

mkuehbach (approved):
lgtm

PR #1608, opened 2026-01-07
Ensure any reference to inherited nodes match tag type

PeterC-DLS:
This fixes the CI issue for #1560 where a `model` field collides with `model` attribute

### DIFF
--- a/dev_tools/docs/nxdl.py
+++ b/dev_tools/docs/nxdl.py
@@ -722,6 +722,16 @@ class NXClassDocGenerator:
         except FileNotFoundError:
             return ""
         if len(parents) > 1:
+            for parent in parents:
+                # iterate back and check tag matches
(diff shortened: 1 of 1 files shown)


dev_tools/docs/nxdl.py
```diff
@@ -722,6 +722,16 @@ def get_first_parent_ref(self, path, tag):
         except FileNotFoundError:
             return ""
         if len(parents) > 1:
+            for parent in parents:
+                # iterate back and check tag matches
+                if not parent.tag.endswith(tag) and not parent.tag.endswith(
```
woutdenolf: FYI I think we should handle this tag business properly at some point: https://github.com/nexusformat/definitions/pull/1613#discussion_r2793596691.

woutdenolf (approved):

PR #1610, opened 2026-01-08
Rhs chopper

ggoneiESS:
The handedness of the NeXus coordinate system implies that chopper rotation should be positive when rotating clockwise facing away from the source, but the documentation states it is the exact opposite.

This cannot be corrected easily historically, as many systems should have this in place, but it is not correct to describe NeXus as using a right-handed coordinate system with this discrepancy.  It also leads to an unintuitive and ugly exception when describing an instrument.

This PR proposes a fix by including a new field, `handedness`, that defaults to `Left` unless otherwise specified, thus not breaking any existing setups or data files, but allows for another value, `Right`, which agrees with the standard NeXus coordinate basis and which arises naturally when defining a RHS with a z-axis that runs in a positive direction from source to sample.

### DIFF
--- a/base_classes/NXdisk_chopper.nxdl.xml
+++ b/base_classes/NXdisk_chopper.nxdl.xml
@@ -45,9 +45,23 @@
 		We refer to this below as the "top-dead-center signal".
 
 		Angles and positive rotation speeds are measured in an anticlockwise
-		direction when facing away from the source.
+		direction when facing away from the source unless otherwise specified in the `handedness` field.
 	</doc>
 
+	<field name="handedness">
+		<doc>
+      Direction of measurement for rotation and positions in the disk-chopper.
+      Left is the default for historical reasons and implies positive rotation and angles in an anticlockwise
+      direciton when facing away from the source.
+      Right matches the default NeXus coordinate system and implies positive rotation and angles in an anticlockwise
+      direction when looking towards the source - positive rotation matches the positive roll of the system.
+      Only one from the enumerated list (match text exactly).
+    </doc>
+		<enumeration>
+			<item value="Left" />
+			<item value="Right" />
+		</enumeration>
+	</field>
 	<field name="type">
 		<doc>Type of the disk-chopper: only one from the enumerated list (match text exactly)</doc>
 		<enumeration>
@@ -58,8 +72,7 @@
 	</field>
 	<field name="rotation_speed" type="NX_FLOAT" units="NX_FREQUENCY">
 		<doc>
-			Chopper rotation speed. Positive for anticlockwise rotation when
-			facing away from the source, negative otherwise.
+			Chopper rotation speed.
 		</doc>
 	</field>
 	<field name="slits" type="NX_INT">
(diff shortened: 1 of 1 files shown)


phyy-nx:
From Telco Apr 29, this looks fine.  Is there a way in NeXus to specify a default for an enumeration?  Right now, the description says the default is 'Left' but this isn't in the XML itself.

base_classes/NXdisk_chopper.nxdl.xml
```diff
@@ -45,9 +45,23 @@
...
 
 		Angles and positive rotation speeds are measured in an anticlockwise
-		direction when facing away from the source.
+		direction when facing away from the source unless otherwise specified in the `handedness` field.
 	</doc>
 
+	<field name="handedness">
+		<doc>
+      Direction of measurement for rotation and positions in the disk-chopper.
+      Left is the default for historical reasons and implies positive rotation and angles in an anticlockwise
```
phyy-nx, line 54:
```suggestion
      Left was assumed in the past and implies positive rotation and angles in an anticlockwise
```

base_classes/NXdisk_chopper.nxdl.xml
```diff
@@ -45,9 +45,23 @@
...
+      direciton when facing away from the source.
+      Right matches the default NeXus coordinate system and implies positive rotation and angles in an anticlockwise
+      direction when looking towards the source - positive rotation matches the positive roll of the system.
+      Only one from the enumerated list (match text exactly).
```
phyy-nx, line 58:
```suggestion
      Only one from the enumerated list (match text exactly).  If `handness` is not specified, it's likely that downstream consuming applications will assume left-handed.
```

base_classes/NXdisk_chopper.nxdl.xml
```diff
@@ -45,9 +45,23 @@
...
 		Angles and positive rotation speeds are measured in an anticlockwise
-		direction when facing away from the source.
+		direction when facing away from the source unless otherwise specified in the `handedness` field.
 	</doc>
 
+	<field name="handedness">
+		<doc>
+      Direction of measurement for rotation and positions in the disk-chopper.
+      Left is the default for historical reasons and implies positive rotation and angles in an anticlockwise
+      direciton when facing away from the source.
```
phyy-nx, line 55:
```suggestion
      direction when facing away from the source.
```

phyy-nx (changes requested):

phyy-nx:
@ggoneiESS after the above comments are addressed, from Telco May 27 we can move this to a 2-week online vote.  Thanks!

base_classes/NXdisk_chopper.nxdl.xml
```diff
@@ -58,8 +72,7 @@
 	</field>
 	<field name="rotation_speed" type="NX_FLOAT" units="NX_FREQUENCY">
 		<doc>
-			Chopper rotation speed. Positive for anticlockwise rotation when
-			facing away from the source, negative otherwise.
+			Chopper rotation speed.
```
phyy-nx, line 75:
```suggestion
			Chopper rotation speed, in the convention of the handedness of the chopper.
```

phyy-nx: Note, it's a bit odd that the user can specify a negative rotation speed to bypass the handedness.  Is that the convention in neutrons?

tacaswell:
"power cycled" to try and get the CI to run correctly.

phyy-nx (changes requested):

PR #1614, opened 2026-02-07
[WIP] XAS detection modes as base classes

mretegan:
This PR introduces distinct base classes for XAS detection modes and updates the NXxas application definition to utilize them via a choice element as suggested here: https://github.com/nexusformat/definitions/pull/1352#issuecomment-3422349124

### DIFF
--- a/dev_tools/docs/nxdl.py
+++ b/dev_tools/docs/nxdl.py
@@ -364,7 +364,7 @@ class NXClassDocGenerator:
         :returns: formatted text
         """
         tag = node.tag.split("}")[-1]
-        if tag in ("field", "group"):
+        if tag in ("field", "group", "choice"):
             optional_default = not self._use_application_defaults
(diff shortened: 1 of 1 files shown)


mretegan:
I keep going back and forth between these three options. 

The last approach is the best one in terms of validation at a cost of having ~15 new base classes, which might be a bit too much considering the existing NeXus classes. If this could, however, be reused for other techniques, then it might make sense.

Grouping modes avoid part of the issue, but aside from the fact that now some validation will need to be done at the application, the groups are at least partially arbitrary. Another person will come up with a different way to do it.

The first approach delegates all validation at the application level. FAIRmat seems to use a simple string to indicate the type of experiment: 
https://fairmat-nfdi.github.io/nexus_definitions/classes/applications/NXxps.html
https://fairmat-nfdi.github.io/nexus_definitions/classes/applications/NXoptical_spectroscopy.html

Even though this PR goes towards approach 2 or 3, I am more favorable to using 1.

PR #1621, opened 2026-03-02
NXobject: add W3ID to identifier type

### LINKED ISSUE #1620: Add W3ID to the open enumeration /NXobject/identifierNAME@type-attribute
https://manual.nexusformat.org/classes/base_classes/NXobject.html#nxobject-identifiername-type-attribute

woutdenolf:
FYI @paulmillar @hgoerzig 

Examples:

- https://w3id.org/PaN/ESRFET#XAS
- https://w3id.org/PaN/ESRFET/v0.1.0#XAS

### DIFF
--- a/base_classes/NXobject.nxdl.xml
+++ b/base_classes/NXobject.nxdl.xml
@@ -192,6 +192,18 @@
 						Example: https://purl.org/dc/elements/1.1/title
 					</doc>
 				</item>
+				<item value="W3ID">
+					<doc>
+						A w3id is a persistent HTTP URI provided via the w3id.org service,
+						operated by the W3C Permanent Identifier Community Group.
+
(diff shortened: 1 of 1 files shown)


phyy-nx:
From Telco Apr 29th.  Instead of W3ID, we should add an IRI field:
https://en.wikipedia.org/wiki/Internationalized_Resource_Identifier

The reason is that IRI is a more general concept that can be used for any ontology concept.  In fact, we could deprecate PURL for the same reason.

@paulmillar will put in an alternate PR.  Thanks for the contributions everyone!

paulmillar:
Please see pull request #1631.

phyy-nx:
@woutdenolf given #1631, ok to close?

PR #1631, opened 2026-04-29
NXobject: add IRI as valid persistent identifier type

paulmillar:
Allow `IRI` as a persistent identifier type.

Deprecate PURL in favour of IRI.


phyy-nx:
After the above comment is resolved by @paulmillar, we will start a 2 week online vote.  Thanks!

PR #1640, opened 2026-07-08
`NXxas` and `NXxas_trans` application definitions and required base classes

mretegan:
The MR introduces a subset of the X-ray absorption application definitions that use the inheritance-based mechanism discussed in #1352 and https://github.com/XraySpectroscopy/nexus_definitions/pull/7. Once validated, new MRs will be opened for the remaining classes, which are somewhat more complex.

The new/updated application definitions are:

- `NXxas` is a generic application definition for X-ray absorption spectroscopy (XAS).
- `NXxas_trans` is an application definition that specializes `NXxas` for data acquired in transmission mode. 

The new classes will replace the current `NXxas` and `NXxasproc`, as each of them is designed to hold both raw and processed data, i.e., no need to have separate definitions.

One particularly is the use of `dataRank` to allow for either 1D or 2D data. Currently this is only expressed in the `doc` of different fields, but @woutdenolf had a few suggestions on how this could also be validated (see https://github.com/XraySpectroscopy/nexus_definitions/pull/7#discussion_r3522835404).


phyy-nx:
From Telco: This removes NXxas from application definitions which requires a vote and adds new content to contributed definitions, which doesn't require a vote.  To me that means this should be split into two PRs, one removing old content and one adding new content to contributed definitions.  Also your comment mentions removing NXxasproc, but I don't see that in this pull request?

Out of curiosity, how close is this to being fully ready for adoption in the community?

Finally, just verifying that the old NXxas is fully unused, so greatly changing its meaning won't affect a huge of amount of users/data, correct?

mretegan:
@phyy-nx Thank you for the feedback.

I separated the removal of the existing `NXxas` and `NXxasproc` to another MR #1651. As I commented there, these classes are not used in the community. @newville can further comment on this if needed.

The current MR builds on this and on a second commit #1619 that introduces a few base classes, like `NXabsorption_edge` and `NXelement`, which are required for the new classes. 

All the new classes land in the contributed definitions.

Regarding adoption by the community, this will certainly take some time. However, during our meeting in July, we received full support from the chairs of the International X-ray Absorption Society (IXAS) and the IUCr commission on XAS. Additionally, following feedback from scientists doing quick XAS acquisitions at the same meeting, the new definitions were updated to allow multiple spectra to be stored.

newville (commented):
Can 'is_experimental' be true by default?

newville (commented):
Does this mean that only NXxas_trans can have a reference channel?  
Why is there that limitation?  How would one communicate fluorescence data with a reference channel?

mretegan:
> Can 'is_experimental' be true by default?

No, it is not possible to have defaults.

> Does this mean that only NXxas_trans can have a reference channel?
> Why is there that limitation? How would one communicate fluorescence data with a reference channel?

It can be added to any of the subclasses that need it.

newville:
@mretegan

>> Can 'is_experimental' be true by default?
>
> No, it is not possible to have defaults.
 
But lots of fields are optional, no? 
So if that became 'is_simulation', then that would only be needed for calculations.

It's really a question of what the minimum is that people/beamlines adding spectra in this format need to give....  Having to say that every spectra measured as a beamline seems sort of odd.

Maybe it would be possible to have NXxas_calc subclass of NXxas?

>> Does this mean that only NXxas_trans can have a reference channel?
>> Why is there that limitation? How would one communicate fluorescence data with a reference channel?
>
> It can be added to any of the subclasses that need it.

If it was in the base class, then it would be available to all sub-classes.

PR #1645, opened 2026-08-20
add stagnant_contributions dir

### LINKED ISSUE #1644: create a stagnant_contributions directory to hold definitions that appear to be unused instead of deleting them
As part of the effort to clean up the _contributed_definitions_ directory of the repo it was determined that it is not easy to find out if a contributed definition is actually being referenced in the wild by a user, as a result to just remove the definition poses 2 problems, 

1. The abrupt un-announced removal may surprise facilities that are currently using it and break their workflow, 
2. It may be the case that although the definition may not be in use now due to loss of a champion for it, to remove it completely would be to remove any work that would have initially been put into it, and would therefore be lost if a champion for the definition was later found.

As an option to mitigate these 2 issues I propose that much like _contributed_definitions_ is a staging area for promotion, there also be a staging area for demotion, a new directory called _stagnant_contributions_ be created for this purpose where definitions deemed as most "likely unused" can be put indefinitely. If we hear back that the definitions was actually in use the file can be put back or promoted, also if someone comes along and wants to propose a definition that is already in or similar to one in the _stagnant_contributions_ directory they would have the option of using it as a start and also view any GitHub issues/comments/discussion that would have been assigned to that definition so that the same dead ends are not traveled during its development, ultimately we would want to retain the github history.

RussBerg:
stagnant_contributions is a directory intended to hold definitions that are not used but may have been used or have had development done to them.

closes #1644

### DIFF
--- a/stagnant_contributions/readme.txt
+++ b/stagnant_contributions/readme.txt
@@ -1,0 +1,3 @@
+# stagnant_contributions is a directory intended to hold definitions that are not used
+# but may have been used or have had development done to them.
+


phyy-nx:
Recommendations from telco:
- Move sub folder of contributed 
- Rename to dormant_contributions
- Check if the documentation will be built if these are in a sub folder
- When definitions are moved to dormant, add to them a note as to why they are there.  "This application definition is dormant, we believe it is not used, and it is not planned to move into the main application definitions folder.  Please contact XXX for more details."  Or something like this.

PR #1653, opened 2026-09-09
Move NXregion to base classes

PeterC-DLS:


phyy-nx:
I found a place NXregion is referred to in a base class, but only as documentation:

https://github.com/nexusformat/definitions/blob/main/base_classes/NXroi_process.nxdl.xml#L42

Further, it isn't in any application definitions yet.  Should it be promoted?

For `scale`, has the scale been applied in the downsampled data or is the user/consumer supposed to apply scaling and the downsampled data is unscaled?

